PR #12, opened 2020-08-11
feat(ui): add component Form

VienNguyen24hdev:


libs/ui/src/components/form/Form.data.ts
```diff
@@ -0,0 +1,201 @@
+import { ReactNodeI } from '../../../../graph/src/node/codec/Node-react'
```
webberwang, line 1: Can do scoped import `from '@codelab/graph'`

PR #21, opened 2020-08-17
feat(ui): add component Dropdown

VienNguyen24hdev:


libs/ui/src/renderer/TreeDom.tsx
```diff
@@ -24,6 +30,7 @@ export class TreeDom {
      */
     const buildComponent = (node: Node<P>) => {
       const [Type, props] = factory(node)
+      const reactNodeProps = produceReactNodeProps(props)
```
webberwang, line 33: - Extract into Node.tsx, we can add a HOC for Node.Component
- In the same HOC, we can extract https://github.com/codelab-ai/codelab.ai/blob/67e0d7e32b01acf17341e9a366ad44c3debd607d/libs/graph/src/node/Node.tsx#L85
- If you think about it, both `produceReactNodeProps` & `evalPropsWithContext` are transforming props for the component, would make sense to extract & combine both into the same HOC function.
- This could also eliminate the `import/no-cycle`. Usually circular dependency is caused putting the function in the wrong module. TreeDom is more of a top-level API, the internals should be hidden
- By extracting logic into HOC, this makes testing easier as we can unit test this function

webberwang (commented):
- [ ] Add unit test

We want to add a unit test for the HOC. But before we test the HOC, we should add tests for `produceReactNodeProps` & `evalPropsWithContext`

Inside the HOC, we will basically do the same testing logic in the previous 2 tests, but just combined together.

We should have 3 total test files added.

- 1) test that eval props are evaluated
- 2) test that react input are transformed into components (do integration test test that TreeDom.render is called with parameter, use spy?)
- 3) 1 + 2 combined

PR #39, opened 2020-08-19
feat(ui): add prop types to old component

VienNguyen24hdev:
- Checkbox
- Dropdown
- Form
- Icon
- Select
- Typography


libs/props/src/libs/Props.interface.ts
```diff
@@ -14,5 +17,5 @@ export type Props = {
 }
 
 export type PropsFromKeys<Keys extends string> = {
-  [K in Keys]?: string | number
-}
+  [K in Keys]?: string | number | boolean | PropValue | ReactNodeI | any
```
webberwang, line 20: I removed `any` here, having that makes typing not as useful

PR #59, opened 2020-08-27
feat(libs-model): add schema api

VienNguyen24hdev:


webberwang:
This looks great and everything works perfectly, but I'd like to use the approach we are currently using with our react component.

`Tree-factory` `makeTree` takes in the json input, and uses a treeWalker function to traverse the nodes (starting at the root node then down each children). Then it reduces it down to a single object.

The iteratee is where you would build the model/schema.

Instead of building a Tree, we're building a Model/Schema

webberwang:
idea is to co-locate the logic for building each node type (mode & schema) if possible

not sure if this is possible, but could be worth looking into





the traversal order will ensure that the schema is built first, then passed to the loop that then builds the model

very happy with the work so far, these refactors will help make our code more generic & help refactor down the line

PR #74, opened 2020-09-08
Conditional form field

VienNguyen24hdev:


libs/ui/src/components/form/Form.spec.tsx
```diff
@@ -34,4 +34,32 @@ describe('Form', () => {
     expect(getByDisplayValue('Webber').id).toBe('company_devs_0_name')
     expect(getByDisplayValue('Vien').id).toBe('company_devs_1_name')
   })
+
+  it('should render conditional form field', async () => {
+    const { getByText, getByLabelText, getByRole } = render(<ConditionalForm />)
+
+    expect(getByLabelText('Select')).toBeTruthy()
+
+    const fieldA = getByText('Field A').parentElement.parentElement
```
webberwang, line 43: when possible use `@testing-library/react`, we can use getByLabelText & test for truthy. this way we don't need to depend on className

PR #99, opened 2020-09-13
chore: add lsp logs to .gitignore

aburtsev:
## Current Behavior

If you run the LSP server separately then it generates logs in the root folder. It has any sense only for local copy, so PR about to add folder '.log' to .gitignore. It'll make the project more IDE agnostic.

## Expected Behavior

git shouldn't track the ".log" folder in the root folder

## Related Issue(s)

Fixes #

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -26,6 +26,9 @@ node_modules/
 !.vscode/launch.json
 !.vscode/extensions.json
 
+# LSP
+.log/*
+
 # misc
 .sass-cache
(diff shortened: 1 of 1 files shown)


webberwang:
Let me add your github/slack username in the circleci config first

aburtsev:
I'll check circleCI.

I've been using emacs.

aburtsev:
Duplication #116

PR #107, opened 2020-09-16
Refactor props based on RTM

VienNguyen24hdev:


libs/shared/interface/props/src/props.ts
```diff
@@ -18,3 +18,17 @@ export interface Props {
 export type PropsFromKeys<Keys extends string, P extends object = {}> = {
   [K in Keys]?: string | number | boolean | PropValue | P
 } & { ctx?: PropItem; style?: CSSProperties }
+
+export type PropsFactory = (
```
webberwang, line 22: Use `interface` where possible

libs/core/props/src/libs/Props-eval.tsx
```diff
@@ -27,52 +38,64 @@ export const evalPropsFactory = (propValue: PropItem, ctx?: any) => {
...
 
-  return propValue
+  return { ...acc, [propKey]: propValue }
 }
 
 /**
  * Goes through each object key/value pair, and apply the iteratee
  */
-export const evalPropsIterator = <P extends Props = Props>(
+export const evalPropsIterator: PropsIterator = <P extends Props = Props>(
```
webberwang, line 57: rename to `propsIterator`, we named with eval because it used to deal with evaluating string functions. now responsibility is moved out

libs/core/props/src/libs/Props-eval.tsx
```diff
@@ -27,52 +38,64 @@ export const evalPropsFactory = (propValue: PropItem, ctx?: any) => {
...
   props: P,
+  iteratee: PropsFactory,
   ctx?: any,
```
webberwang, line 60: Notice `PropIterator` doesn't have `ctx`. We used to use `ctx` so we can add to `this` on same level, now we just defer to `Provider`

Remove `ctx`, instead of passing TreeDom to `ctx` (attempt to remove circular dep), we'll just add TreeDom to wherever we setup `ctx` (it's okay if we have circular dep, just eslint ignore for now)

webberwang (approved):
`Node.tsx` should have no knowledge of how to construct props, it will access a single method from props module. Refer to `PROP-F-1` notes section

webberwang:
`Node.tsx` should have no knowledge of how to construct props, it will access a single method from props module. Refer to `PROP-F-1` notes section

libs/components/ui/src/renderer/TreeDom.tsx
```diff
@@ -15,9 +15,9 @@ import { elementParameterFactory } from './ElementFactory'
 /**
  * We need this function in ui package because TreeDom is required, can't put in node or props package
  */
-const evalPropsWithTreeContext = (props: Props): Props => {
+const evalPropsWithTreeContext = (props: Props, parentProps: Props): Props => {
```
webberwang, line 18: Remove this method, we'll just add TreeDom to `evalPropsWithContext`, which should be renamed to `buildProps()`

libs/core/props/src/libs/Props-eval.tsx
```diff
@@ -27,52 +38,64 @@ export const evalPropsFactory = (propValue: PropItem, ctx?: any) => {
...
+    ctx ? { ctx } : {},
   )
 }
 
 /**
  * Allows us to build ctx & pass into props without needing a parent & child component
  */
-export const evalPropsWithContext = (props: Props): Props => {
-  const { ctx = {}, ...restProps } = props
+export const evalPropsWithContext: PropsBuilder = (
```
webberwang, line 74: rename to `buildProps`, this is the only method `Node.tsx` should call

libs/shared/interface/props/src/props.ts
```diff
@@ -18,3 +18,17 @@ export interface Props {
 export type PropsFromKeys<Keys extends string, P extends object = {}> = {
```
webberwang, line 18: Refactor according to `PROP-F-8`

webberwang (changes requested):
More changes (still learning how best to use review feature in Github)

libs/shared/interface/props/src/props.ts
webberwang: PropsFromKeys can be simplified via PROP-F-8

Issue #108, opened 2020-09-16
`eval` props behavior


aburtsev:
@webberwang 
## Current Behavior

If wrap any component with Modal then basic prop passing becomes broke. By basic I mean known props (for ex. onClick for a button). 

## Expected Behavior

The modal component should work consistently with others.

## Steps to Reproduce

Please find the example story called modal -> "With Form" in [branch in fork](https://github.com/aburtsev/codelab.ai/tree/modal-form-bug).

webberwang:
Fixed by #109, issue is that we need to specify `renderProps: ['leaf', 'eval']'` for a function to pass more than 1 level

aburtsev:
@webberwang  But in the current setup, all basic props passed through multiple levels, but if you wrap something with Modal. That this behavior is breaking. Please have a look at this [example](https://github.com/aburtsev/codelab.ai/commit/a243517b7fb47ca98f03e93761b2bbff31bbb56a) . If you remove 'Modal' wrapper than everything will work as expected. So to be consistent it should or not work in every case (=without renderProps should be passed only through 1 level) or work in all cases (=every basic props should be passed through all levels)

webberwang:
Props should only pass 1 level down without specifying renderProps, could you show me where it is passing down multiple levels?

aburtsev:
> @aburtsev Props shouldn't pass to children if we don't specify renderProps, could you show me where it is passed down?

@webberwang please have a look an example "BasicPropsPassedInDeep" in #114

webberwang:
@aburtsev I see now, thanks for catching this. I have added `PROP-TC-13` & `PROP-TC-14` to check for this

aburtsev:
> Fixed by #109, issue is that we need to specify `renderProps: ['leaf', 'eval']'` for a function to pass more than 1 level

It will be fixed by #109? I've tried to use it and such chaining doesn't work

webberwang:
> > Fixed by #109, issue is that we need to specify `renderProps: ['leaf', 'eval']'` for a function to pass more than 1 level
> 
> It will be fixed by #109? I've tried to use it and such chaining doesn't work

I meant will be fixed, is this a blocker for you?

webberwang:
Closed by #140

VienNguyen24hdev:
via Hubstaff
User: Nguyen Luong Vien

Project: Codelab - https://app.hubstaff.com/projects/1093588
Date Range: 09/22/20 - 09/22/20
[...]

PR #114, opened 2020-09-17
Modal form bug

aburtsev:
## Current Behavior

This is PR for demonstrate issue described in the #108   

## Related Issue(s)

#108 

Fixes #

### DIFF
--- a/libs/components/ui/src/components/modal/Modal.data.ts
+++ b/libs/components/ui/src/components/modal/Modal.data.ts
@@ -101,7 +101,7 @@ export const modalDataWithForm: ReactNodeI<Modal.Props | Button.Props> = {
       nodeType: 'React',
       props: {
         title: 'Basic Modal',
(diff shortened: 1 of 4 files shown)


webberwang:
I won't merge these stories because we'll have unit tests to test for those edge conditions.

aburtsev:
> I won't merge these stories because we'll have unit tests to test for those edge conditions. Will close after the feature is fixed & will retest the stories.

Great! Sorry for the unclearness. Tests definitely fit better. I just fixed styles after discussion of eslint and precommit-hooks and didn't have intention to merge this branch. 
Maybe it'll be convenient if we mark such PRs somehow (for ex. with prefix in the title), so it will not looks as real PR. And then we can quickly understand do we need to treat this branch as PR or not.

webberwang:
would be great if GitHub let us add tags to mr as well

Issue #71, opened 2020-09-08
Node/Tree builder


webberwang:
Before getting started...

Run `nx storybook ui` & check out all the storybooks, you'll notice each component receives data that follows the ReactNodeI interface

currently we create this data using code, but we want to create a UI that supports this

------------

Create `core/pages/node.tsx` that can build any Node/tree structure (which includes props & children). Avoid writing any CSS code.

![Tree UI](https://user-images.githubusercontent.com/26264925/92437795-87737f00-f15c-11ea-98b6-020b46b3cb36.png)

* Forgot to add submit for popover form

To visualize data we will use https://ant.design/components/tree/#API. 

We will use `libs/node/src/base/Node.tsx` for building the tree.

- Use `Node.tsx` 'addChildren()`
- Create `removeChildren()` for removing children
- ~~Create `moveChildren()` for moving nodes, which should work with dragging~~ skip for now

For nodeType we can use enum/select input with predefined values. For props, user should be able to add any key/value pair.

For state management use hook for now, for DOM handlers create hooks as well.

------------

Write unit tests for the form
- assert that the nodeType select box contains all options 
- assert that the nodeType select box contains all type
- assert that clicking on add for prop will bring up a form with
  - name field
  - type containing the 3 select options
  - value field
- assert that submit has the correct data format

webberwang:
do this first actually instead, this is a general node tree builder

right now we are constructing our Node tree using code

we are creating a way to do the same thing with UI

it might be worthwhile to explore building our own tree/form component using d3

since we'll be manipulating trees quite a bit

having the UI match the data structure would be great for interoperability

I think in the long term it might save us some headache trying to make antd form work

let me know your thoughts, as creating our own d3 form component is extra work we'll have to do. but we already have d3 tree working so may not take that much extra work? (edited)

webberwang:
I would start with `Node.tsx` class's addChild method, this allows us to append children

webberwang:
<img width="475" alt="Codelab_bmpr_—_New_Project_1" src="https://user-images.githubusercontent.com/26264925/92839110-71500380-f394-11ea-9a04-cb7196b72b85.png">

webberwang:
Display all added data to a table, add delete/edit buttons on the table rows. Edit should bring up modal with pre-filled data.

PR #115, opened 2020-09-17
WIP 71 node tree builder

aburtsev:
## Related Issue(s)

Fixes #
#71


apps/web/pages/node/Node-react--type.i.ts
```diff
@@ -0,0 +1,100 @@
+// temporary file. Until it'll be not merged in upstream master
```
webberwang, line 1: Already exists here https://github.com/codelab-ai/codelab.ai/blob/master/libs/shared/interface/node/src/subtypes/react/node-react--enum.ts

apps/web/pages/node/node.helpers.ts
```diff
@@ -0,0 +1,26 @@
+import { ReactNodeI } from '@codelab/shared/interface/node';
+
+interface ISelectOptionProps<T> {
+  value: T
+}
+
+export function generateSelectOption<T>(
```
webberwang, line 7: Exists here https://github.com/codelab-ai/codelab.ai/blob/master/libs/components/ui/src/components/form/Form.types.ts

webberwang (changes requested):

apps/web/pages/node/Node-react--type.i.ts
aburtsev: I'll fix it

apps/web/pages/node/node.helpers.ts
aburtsev: I'll remove it and use the implementation from From.types.ts
Waiting for your decision about #119

webberwang: Updated

webberwang:
@aburtsev I refactored the organization a bit. A couple things I noticed.

- Breaking down a component into too many pieces is very difficult to follow. Utilize code folding & we can easily keep all the data required on the same file
- Use lower case naming for function props, as React throws an error
- For mapping data check out `tree-map.spec.tsx`

ButtonGroup_tsx_—_codelab__Workspace_

Hopefully this initial task gives you a better idea of how the project works. I've made some changes how types work will post update in #team on slack.

Issue #118, opened 2020-09-18
Lib Form(Form.types.ts) wrong type definition


aburtsev:
## Current Behavior

[Link to Form.types.ts](https://github.com/codelab-ai/codelab.ai/blob/master/libs/components/ui/src/components/form/Form.types.ts)

Current logic expects only `typeof NodeTypeEnum` as type of "options" field in interface "CreateSelect".

## Expected Behavior
"createSelect" and others methods should work with any pair of [key,value].
Additional: In method "createOptions" I propose to use 'unknown' instead of 'any'. It will force to use additional checks for the value when somebody will use it

aburtsev:
PR #119 is merged

PR #119, opened 2020-09-18
fix(libs): #118 fix Lib Form(Form.types.ts) wrong type definition

aburtsev:
## Current Behavior

Described in #118 

## Expected Behavior

Described in #118 

## Related Issue(s)

Fixes #

#118

### DIFF
--- a/libs/components/ui/src/components/form/Form.types.ts
+++ b/libs/components/ui/src/components/form/Form.types.ts
@@ -1,4 +1,4 @@
-import { PropsFromKeys } from '@codelab/shared/interface/props'
+import { PropsFromKeys, PropJsonValue } from '@codelab/shared/interface/props'
 import { ReactNodeI } from '@codelab/shared/interface/node'
 import { Select } from '../select'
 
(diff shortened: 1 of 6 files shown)


webberwang:
I do like how we accept an object instead of enum to make it more useable. Unknown could be changed to accepted value which is `string | number | boolean` for json I believe

aburtsev:
> I do like how we accept an object instead of enum to make it more useable. Maybe no reason to use generic if the values will always be this type `string | number | boolean`.
> 
> Let me know your thoughts on new commit

I more prefer generic, because it gives you the ability to be more specific if you want. But your solution is also OK, thanks for the quick review.

webberwang:
> > I do like how we accept an object instead of enum to make it more useable. Maybe no reason to use generic if the values will always be this type `string | number | boolean`.
> > Let me know your thoughts on new commit
> 
> I more prefer generic, because it gives you the ability to be more specific if you want. But your solution is also OK, thanks for the quick review.

Okay sure I'll add the generic back in, since if we have a case where only strings are allowed we can specify that. We'll default to `string | number | boolean` instead of unknown so we aren't forced to specify a type.

PR #122, opened 2020-09-19
refactor(libs-core-renderer): add the ability to specify the returned type for TreeDom renderer

aburtsev:
## Current Behavior
Even if you specify the type in generic in TreeDom.render, it will not validate the resulted React component props for validness 

## Expected Behavior
- Static analysis can validate React component props
- Autocompletion work 

## Related Issue(s)

Fixes #

### DIFF
--- a/apps/web/src/node/ButtonGroup.tsx
+++ b/apps/web/src/node/ButtonGroup.tsx
@@ -82,11 +82,9 @@ const nodeButtonGroupData: NodeDtoReactI = {
   ],
 }
 
-export interface NodeButtonGroupProps {
-  toggleform: () => void
-  handledelete: () => void
+export interface ButtonGroupProps {
+  setvisibility: Function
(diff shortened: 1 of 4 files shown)


webberwang:
in `ElementFactory` we have to cast to `any`, is there a way around that with this approach?

aburtsev:
> Do you mean the rendered component has no prop checking?
> 
> ```
> const Button = TreeDom.render(button)
> 
> return <Button someProp={} />
> ```
> 
> Do you mean `someProp` doesn't trigger eslint error? If not could you provide a screenshot

I've installed and set up VS Code to be sure that the issue not related to my setup. Please have a look screenshots below
image
image

webberwang:
I see will take a look

webberwang:
thanks for suggesting this, now we have better type safety when using components

PR #123, opened 2020-09-19
refactor(libs-components-ui): extend the Form.createSelect params with "style"

aburtsev:
## Expected Behavior

Form.createSelect should give the ability to apply styles 

## Related Issue(s)

Fixes #


webberwang:
So this helper method is only temporary & used for rapid prototyping. The production site will actually pull data from the database.

Since you've already completed this I'll merge it

aburtsev:
The current default width of Select is 120px. We have long names and I just want to add the ability to quickly apply some style. It only for prototyping. And as we build UI for configuration based on our UI, I thought it'll be convenient to have a chance to add styles quickly through helper. Otherwise, if I need to apply some style, I need to refuse from the use of our helper for create Select

webberwang:
You can increase it directly there for now, it doesn’t hurt to have all the select a bit wider

PR #156, opened 2020-10-06
Refactor node page to use Service, Repository

VienNguyen24hdev:


apps/web/pages/node.tsx
```diff
@@ -27,19 +26,28 @@ const NodePage = () => {
...
-      .get('/api/v1/Node')
-      .then((res) => {
-        setNodes(res.data)
-      })
-      .catch((err) => {
-        console.log(err)
-      })
+    service
+      .getNodes()
+      .then((res: any) => setNodes(res.data))
```
webberwang, line 31: move the success then into `node-repository`. catch the error in `node-service`

apps/web/pages/node.tsx
```diff
@@ -58,8 +66,8 @@ const NodePage = () => {
   const handleCreateNode = (formData: any) => {
     console.log(formData)
 
-    axios
-      .post('/api/v1/Node', formData)
+    service
+      .createNode(formData)
```
webberwang, line 70: you can change to `createNode(formData, cb)`, then cb can call setNodes, setVisibility. if you need access to nodes, you can even do `createNode(formData, cb(nodes))`

where `cb = (nodes) => () => {}`

apps/web/pages/node.tsx
```diff
@@ -27,19 +26,28 @@ const NodePage = () => {
...
-      .catch((err) => {
-        console.log(err)
-      })
+    service
+      .getNodes()
+      .then((res: any) => setNodes(res.data))
+      .catch((err: any) => console.log(err))
+  }
+
+  const findChildren = (inputNodes: Array<any>) => {
```
webberwang, line 35: we have a `findNode` method for this, use vscode symbol search to find this method

apps/web/pages/node.tsx
```diff
@@ -70,28 +78,26 @@ const NodePage = () => {
...
     console.log(formData)
-    if (typeof editedNode?._id) {
-      return
-    }
 
-    axios
-      .patch(`/api/v1/Node/${editedNode._id}`, formData)
-      .then((res) => {
+    service
+      .updateNode(editedNode._id, formData)
```
webberwang, line 88: we actually want `id` instead of `_id`, we should 

node-repository_ts_—_codelab-reclone

I think we can have a single response mapper like this.

webberwang (changes requested):

PR #158, opened 2020-10-07
Cz commit

### LINKED ISSUE #157: yarn commit. Numbers are not corresponding to hotkeys
## Current Behavior
If you type for example "3" it will select the third line from the top 

## Expected Behavior

1. All options have numbers
2. If you type number it will select the option with the corresponding number

webberwang:
Fixes #157


webberwang:
@aburtsev thanks for catching this!

PR #164, opened 2020-10-08
test(apps-web-e2e): create node form

aburtsev:
## Related Issue(s)

Fixes #


webberwang (changes requested):
- best not to have class names in the actual test, we can move those to the helper commands
- no need to test for exist if we're already clicking on it
- use testing library when possible, like `ByRole`
- emulate user interaction when possible, such as clicking on dropdown instead of setting attributes manually

Issue #142, opened 2020-09-29
Enable form hooks


webberwang:
To add custom component follow these steps

1.  <img width="824" alt="element-factory_tsx_—_codelab__Workspace_" src="https://user-images.githubusercontent.com/26264925/94500789-07699380-01b5-11eb-93f9-8306ca631189.png">

2. <img width="1011" alt="Form_tsx_—_codelab__Workspace_" src="https://user-images.githubusercontent.com/26264925/94500837-2536f880-01b5-11eb-9509-1ffa5032272c.png">

3. Inside `Form`, we should using binding form hook ` const [form] = Form.useForm();`, then pass `form` as a leaf renderProp so all children component can access. Then you can access `this.              form.resetFields();` on the form success callback

webberwang:
@aburtsev I pushed the new refactor for the props, not sure if it allows us to create the custom form component.

If not could you look into modifying props flow so we are able to add the form hook as a custom component?

webberwang:
@aburtsev hey did you have any questions regarding this task?

aburtsev:
@webberwang I've seen the result of refactoring and that currently, I'm able to use form hooks. I'll try to implement a custom form that meets the requirements of the description.

PR #168, opened 2020-10-12
feat(libs-core-renderer): #142 enable form hooks

aburtsev:
ISSUES CLOSED: #142

## Related Issue(s)

Fixes #


webberwang:
Storybook_and_Franz

The `form` isn't getting passed, can you add a spec for reset?

webberwang (changes requested):
Storybook_and_Franz

The `form` isn't getting passed, can you add a spec for reset?

The docs show `form` should be added like this `<Form form={form} />`, but `form` isn't available

libs/core/renderer/src/libs/renderer-components.tsx
```diff
@@ -32,6 +33,19 @@ export const buildComponents = <P extends Props = {}>(
     }: PropsWithChildren<P>) => {
       // console.log(node.type, internalProps)
 
+      // eslint-disable-next-line react-hooks/rules-of-hooks
+      const [form] = AntForm.useForm()
+
+      if (node.type === 'React.Form') {
```
webberwang, line 39: Storybook_and_Franz

The `form` isn't getting passed, can you add a spec for reset?

The docs show `form` should be added like this `<Form form={form} />`, but `form` isn't available

webberwang (changes requested):

libs/core/renderer/src/libs/renderer-components.tsx
aburtsev: It is a bug. To avoid an error message I moved the initiating of form inside Component function and maybe some cache hasn't updated, so when I checked it worked, but I checked today and it isn't worked. I moved it back and everything is fine, but it appeared that this solution works fine in the storybook example, but produces errors on the node-page. I try to find the root of the issue

apps/web/src/node/ModalForm.tsx
```diff
@@ -187,50 +186,53 @@ export const modalFormData: NodeReactI = {
...
+    {
+      type: 'React.FormHooksWrapper',
+      children: [
+        {
+          type: 'React.Form',
+          props: {
+            name: 'create-node-form',
+            onFinish: {
+              __type: ['Eval'],
+              value: 'return this.handlesubmit',
```
webberwang, line 200: If we're using this approach could we add the hook here without having an extra wrapper component?

webberwang (changes requested):

apps/web/src/node/ModalForm.tsx
aburtsev: When we 'eval' our props we use as context the ctx with libs(React, etc) and renderedProps passed from the parent, hence if we will not use wrapper we can't use formHooks on the level of React.Form (we could use only in descendants), but it's might be very useful (even more we already use it like this on the node-page)

webberwang: is useForm required on level of Form? could be use pass product of useForm from ctx with libs along with antd Form

libs/core/renderer/src/libs/element-factory.tsx
```diff
@@ -142,6 +142,8 @@ export const elementParameterFactory: NodeFactory<
       return [Dropdown as any, props]
     case 'React.Form':
       return [Form, props]
+    case 'React.FormHooksWrapper':
```
webberwang, line 145: We don't have to add this, we can just use a `React.Fragment` to pass down `form`

libs/core/renderer/src/libs/renderer-components.tsx
```diff
@@ -19,6 +20,29 @@ export const buildComponents = <P extends {} = {}>(
   const componentBuilderIteratee: NodeFactory<void> = (node: any) => {
     const [Component, props] = elementParameterFactory(node)
 
+    if (node.type === 'React.FormHooksWrapper') {
```
webberwang, line 23: We can just use a `React.Fragment` with an eval prop to pass form down

webberwang (changes requested):

apps/web/src/node/ModalForm.tsx
webberwang: We can create a regular React Fragment and pass down with eval props

aburtsev: Sorry. Maybe I don't get the goal right.
I thought that we want to be able to have form hooks on the level of JSON. I mean we have JSON-config and have our engine which processes this JSON to create the app.
So we have the following options right now:
1. The same as we have right now. But using such an approach we are polluting global ctx
image
2. Create a wrapper (my current solution)

I still can't get your idea about React.Fragment. To pass down eval props I need to process them on the level of processing JSON. So I need somehow identify that this current context should be extended with form. I can't eval props with form without polluting the global context.

Or you mean that we can pass form as root props?
image
But in this case, our JSON would be relying on where we use it and will not be independent and declarative

webberwang:
closed since we don't need a custom component for hook

PR #176, opened 2020-10-14
feat(libs-core-node): add dot object mapper

VienNguyen24hdev:


webberwang:
@VienNguyen24hdev this should be built into the form component such that `<Form onFinish={(values) => {}}/>` values is mapped like so

libs/core/renderer/src/libs/element-factory.tsx
```diff
@@ -142,6 +142,8 @@ export const elementParameterFactory: NodeFactory<
       return [Dropdown as any, props]
     case 'React.Form':
       return [Form, props]
+    case 'React.CustomForm':
```
webberwang, line 145: We are overriding default behavior of `React.Form`, so there is no need for `React.CustomForm`

libs/ui/antd/src/components/form/Form.stories.tsx
```diff
@@ -40,3 +41,9 @@ export const NodeForm = () => {
 
   return <Form />
 }
+
```
webberwang, line 44: no need for this

libs/ui/antd/src/components/form/data/Form-5--custom.data.ts
```diff
@@ -0,0 +1,86 @@
+import { Button } from '../../button'
+import { Text } from '../../text'
+import { Form } from '../Form.types'
+import { NodeReactI } from '@codelab/shared/interface/node'
+
+export const formCustomData: NodeReactI<
```
webberwang, line 6: we can just use `Form-2--object.data.ts`, we are changing default behavior of objects

webberwang (changes requested):

libs/ui/antd/src/components/form/Form.tsx
```diff
@@ -31,12 +33,46 @@ export type FormListProps = Omit<AntFormListProps, 'children'> & {
   children: Array<React.ReactElement>
 }
 
+export const mapperDotProps = (propsArr: Array<object>) => {
```
webberwang, line 36: For the form mapper, we should be using our declarative api to map. We have created the utils class for this purpose

webberwang (changes requested):

libs/ui/antd/src/components/form/data/Form-5--custom.data.ts
webberwang: @VienNguyen24hdev

webberwang:
we'll create new PR for this

PR #187, opened 2020-10-17
178 integrate layout to pages

aburtsev:
@webberwang 

## Related Issue(s)

Fixes #


aburtsev:
@webberwang  After discussion, I think we can close this PR and also #142 issue cause we've decided to use the workaround with global form

libs/ui/component/src/layout/Layout.tsx
```diff
@@ -1,36 +1,42 @@
 import React from 'react'
+import { layoutData } from './Layout.data'
 import { Renderer } from '@codelab/core/renderer'
-import { propLayoutData } from '@codelab/ui/antd'
+import { ComponentProps } from '@codelab/shared/interface/component'
+import { ContextModal, EventModal } from '@codelab/state/modal'
+import { withActor } from '@codelab/ui/hoc'
 
-export const Layout = () => {
+type LayoutProps = {
```
webberwang, line 8: The idea is to make each component independent, so we will create separate components for Header, Footer, Sidebar, Content then pass them in here

webberwang (changes requested):

libs/ui/component/src/machine/MachineProvider.tsx
```diff
@@ -21,6 +21,7 @@ interface MachineContextProps<
   }
   actors: {
     modal: any
```
webberwang, line 23: let's refactor to a single ui actor, with context like `context.modal.visibility` etc

webberwang (changes requested):

PR #189, opened 2020-10-19
feat(libs-ui-antd): add React.RenderComponent

VienNguyen24hdev:


libs/ui/antd/src/components/render-component/RenderComponent.stories.tsx
```diff
@@ -0,0 +1,17 @@
...
+}
+
+const Parent = (props: any) => {
+  return <>{props.children({ data: 'Data' })}</>
+}
+
+export const Default = () => {
+  const Children = Renderer.components(renderComponentData)
+
+  return <Parent>{Children}</Parent>
```
webberwang, line 16: Let's create a separate `React.RenderContainer` for the parent component

webberwang (changes requested):

webberwang:
@VienNguyen24hdev a note on commit, don't squash them when adding revisions, otherwise it makes it harder for me to see what has changed

webberwang:
nice work has been merged!

PR #200, opened 2020-10-21
chore(apps): docker-compose with mongo and neo4j for local dev

aburtsev:
@webber I've quickly set up docker-compose for databases for local development. Maybe it will be useful for you.

## Related Issue(s)

Fixes #


webberwang:
awesome! we will definitely need this. I stopped user Docker locally when they had some memory issues with Mac, they might have fixed it by now

PR #211, opened 2020-10-23
fix(libs-api): fix warning export in yarn build

VienNguyen24hdev:


webberwang:
Is there a solution without having to use `type`?

VienNguyen24hdev:
> Is there a solution without having to use `type`? maybe some tsconfig settings

I'm still looking for. This is just a way to fix temporarily

webberwang:
in this case extends does work better, but for importing `type`, will that warning always show?

VienNguyen24hdev:
> in this case extends does work better, but for importing `type`, will that warning always show?
> 
> how come it doesn't show for react only nestjs?

I think it shows warning here because we export a type from a class WinstonLogger. It doesn't show if we export a regular type

webberwang:
Oh I see now thanks

PR #217, opened 2020-10-26
feat(libs-shared-utils): add mapper for array data

VienNguyen24hdev:
This is just a dump imperative way to satisfy our goal.


libs/shared/utils/src/mapper/mapper-object.spec.ts
```diff
@@ -17,13 +18,109 @@ describe('Mapper', () => {
         city: 'New York',
       },
     }
+
+    const map: Array<MapData> = [
```
webberwang, line 22: if we're using `object-mapper` package, we can just follow their api

webberwang (changes requested):

webberwang:
Fixed by #223

PR #221, opened 2020-10-27
Refactor NodeType enum

VienNguyen24hdev:


libs/shared/interface/node/src/enums/node-enum--react.ts
```diff
@@ -5,105 +5,105 @@ export enum ReactTypeEnum {
   /**
```
webberwang, line 5: We should only keep `node-enum`, no need for `node-enum--react`. If you look at `NodeReactI` & `ReactType`, they all derive from `NodeType`

Same for other subtypes

webberwang (changes requested):

PR #238, opened 2020-11-02
change cmd overidding

nutshell113:
## Related Issue(s)

Fixes #

### DIFF
--- a/.docker/run.sh
+++ b/.docker/run.sh
@@ -10,4 +10,4 @@ docker run \
   --env NEXT_PUBLIC_API_ORIGIN=http://localhost:4000 \
   --env NEXT_PUBLIC_API_PATHNAME=api/v1 \
   codelabai/codelab:0.01 \
-  /usr/local/bin/yarn start:prod:web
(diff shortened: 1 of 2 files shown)


.docker/run.sh
```diff
@@ -6,9 +6,8 @@ docker run \
...
   --expose 3000 \
   --publish 3000:3000 \
-  --entrypoint="['yarn', 'start:prod:web']" \
   --env WEB_PORT=3000 \
   --env NEXT_PUBLIC_API_ORIGIN=http://localhost:4000 \
   --env NEXT_PUBLIC_API_PATHNAME=api/v1 \
-  codelabai/codelab:0.01
-
+  codelabai/codelab:0.01 \
+  /usr/local/bin/yarn start:prod:web
```
webberwang, line 13: thanks! `yarn start:prod:web` works as well, didn't realize we could run command this way

webberwang: @nutshell113 does this seem to be a bug in the `--entrypoint` flag?

PR #241, opened 2020-11-02
init to validate staging workflow

nutshell113:


.circleci/config.yml
```diff
@@ -66,6 +64,18 @@ executors:
     resource_class: large
 
 commands:
+  setup-node-version:
```
webberwang, line 67: I would keep this inside setup-env, same command but should work for both executor

.circleci/config.yml
```diff
@@ -158,6 +159,24 @@ commands:
           mentions: $SLACK_USER
 
 jobs:
+
+  install-and-build-workspace-staging:
```
webberwang, line 163: I'd like to keep only `install-and-build-workspace` for both

webberwang (changes requested):
Also see if you can rebase and figure out the naming convention

.circleci/config.yml
nutshell113: They're different ENVs(diff docker images), we must separate them to diffrent job to let it work .

nutshell113:
This PR is too mess, let replace it by https://github.com/codelab-ai/codelab.ai/pull/245

PR #253, opened 2020-11-04
add k8s prod cluster

nutshell113:


.terraform/backend.tf
```diff
@@ -0,0 +1,12 @@
+terraform {
+  backend "s3" {
+    skip_credentials_validation = true
+    skip_metadata_api_check     = true
+    endpoint                    = "https://sfo2.digitaloceanspaces.com"
+    region                      = "us-east-1"
```
webberwang, line 6: does this get override by the variable "region"?

.terraform/README.md
```diff
@@ -0,0 +1,36 @@
...
+export AWS_SECRET_ACCESS_KEY=yyy
+export DIGITALOCEAN_TOKEN=zzz
+```
+### Step 3: Bootstrap the system
+
+```
+# init plugins
+terraform init
+# generate plan
+terraform plan
```
webberwang, line 32: terraform plan --destroy to remove resources?

nutshell113: no, it's `terraform destroy`, or just simply comment/remove source code and then : `terraform plan` and `terraform apply`

.terraform/backend.tf
nutshell113: `region` is just a required var in this case, since backend is designed for AWS S3, not DO Spaces. So, it's just a tricky to by pass this requirement.

webberwang: are we able to set region as "", since it's not being used it is confusing

.terraform/README.md
nutshell113: We can't, it would be a correct region:
image

.terraform/backend.tf
nutshell113: we can't, it would be a correct region: 
image

PR #264, opened 2020-11-08
202 user module

webberwang:
## Related Issue(s)

Fixes #


libs/api/services/graph/src/edge/edge.entity.ts
```diff
@@ -1,14 +1,15 @@
...
+import { Column, Entity, OneToMany, PrimaryColumn } from 'typeorm'
+import { GraphEntity } from '../graph'
 import { IEdge } from './IEdge'
 
 @Entity('edge')
 @ObjectType({
   implements: [IEdge],
 })
 export class EdgeEntity {
-  @PrimaryGeneratedColumn()
```
webberwang, line 10: why not keep it auto-generated?

libs/api/services/graph/src/edge/edge.entity.ts
```diff
@@ -24,4 +25,7 @@ export class EdgeEntity {
     type: 'jsonb',
   })
   declare props?: any
+
+  @OneToMany((type) => GraphEntity, (graph) => graph.edge)
```
webberwang, line 29: I think it's okay if Vertex doesn't know about which Graph it's in, since anytime we use a Vertex in application code, it is retrieved from the graph.

Or is the inverse required?

libs/api/services/graph/src/graph/graph.entity.ts
```diff
@@ -0,0 +1,42 @@
...
+import { EdgeEntity } from '../edge'
+import { VertexEntity } from '../vertex'
+import { IGraph } from './IGraph'
+import { UserEntity } from '@codelab/api/services/user'
+
+@Entity('graph')
+@ObjectType({
+  implements: [IGraph],
+})
+export class GraphEntity {
```
webberwang, line 18: the graph simply holds an array of vertices & edges, they have no knowledge of anything relations of each other

https://github.com/codelab-ai/codelab.ai/issues/262

```typescript
interface Graph {
  edges: Array<Edge>
  vertices: Array<Vertex>
}
```

libs/api/services/graph/src/graph/graph.service.ts
```diff
@@ -0,0 +1,41 @@
...
+  constructor(
+    @InjectRepository(GraphEntity)
+    private readonly graphEntityRepository: Repository<GraphEntity>,
+  ) {}
+
+  async findAll(): Promise<Array<GraphEntity>> {
+    return this.graphEntityRepository.find()
+  }
+
+  async getGraphsForUser(userId: number) {
```
webberwang, line 18: we actually wouldn't need this

webberwang (commented):
@artonio feel free to make the PR next time, the `Request changes` option is greyed out for me if I create the PR

webberwang:
don't worry much about the merge conflict (no need to rebase)

libs/api/services/graph/src/edge/edge.entity.ts
artonio: in the example you provided, id of an edge could be a string i.e 'e-a', only id of type number can be auto generated

libs/api/services/graph/src/graph/graph.entity.ts
artonio: I thought you wanted an edge and a vertex table. GraphEntity is what is know as a Join table, it joins vertices and edges to the user. We could go even simpler and have graph table like so:

 id | vertices(jsonb)                                    | edges(jsonb)                                                                                                 | userId(fkey) |   |
|----|----------------------------------------------------|--------------------------------------------------------------------------------------------------------------|--------------|---|
| 1  | [{id: 'v-A'}, {id: 'v-B'}, {id: 'v-C'}, props: {}] | [{id: 'e-A', source: 'v-A', target: 'v-B', props: {}}, {id: 'e-B', source: 'v-A', target: 'v-C', props: {}}] | 1            |   |
|    |                                                    |                                                                                                              |              |   |

Would you like to do that instead?

libs/api/services/graph/src/graph/graph.service.ts
artonio: sure, that would work as well

libs/api/services/graph/src/edge/edge.entity.ts
artonio: It is required to make the JoinTable work, but see my comment bellow about a simpler DB structure with just single graph table connected to a user

libs/api/services/graph/src/edge/edge.entity.ts
webberwang: I was actually using that for testing purposes, since since uuid there was a bit more work

.docker/docker-compose.yaml
```diff
@@ -21,7 +21,7 @@ services:
     depends_on:
       - codelab
     env_file:
-      - ../.env
+      - .env
```
webberwang, line 24: actually we run `yarn docker:up []` from root

webberwang:
related to https://github.com/nrwl/nx/issues/2975

PR #274, opened 2020-11-09
feat(apps-web): create graphql data machine

aburtsev:
## Related Issue(s)

Fixes #


webberwang:
@aburtsev the linter broke, circleci must have changed the env variables for the git commits, it should be working now

webberwang (changes requested):
for now we should combine the rxjs into a single stream/function so we can see what parts need to be refactored

webberwang:
nice work closing now, we'll come back it

PR #276, opened 2020-11-10
refactor: before move vertex interface to shared

webberwang:
fix(apps-api-graph): add react to compilerOptions

fix(libs-state-modal): add OK event

## Related Issue(s)

Fixes #


webberwang:
@aburtsev this modified some xstate, rebase early to avoid too many merge errors

Issue #202, opened 2020-10-22
User Module


webberwang:
1. Follow the `Dev Tools` section to generate `libs/api/services/user` using `@codelab/schematics:nest-lib`. This is our Nest.js UserModule--

2. we'll store User model in Neo4j

3. Enable federation at `apps/api/federation/user` and add to gateway

4. add authentication https://docs.nestjs.com/techniques/authentication

<img width="485" alt="node_input_ts_—_codelab__Workspace_" src="https://user-images.githubusercontent.com/26264925/96837125-c43cc200-13fa-11eb-9583-fddd0d740370.png">

5. Create 2 files `queries.graphql`, `mutation.graphql` in `apps/api/federation/user/src/graphql`

Put the queries there you use for manually testing out CRUD operations

6. Also add some tests https://www.apollographql.com/docs/apollo-server/testing/testing/

artonio:
Hasura looks cool. But it only works with Postgres atm. You ok with storing user data in Postgres and not Mongo?

webberwang:
Yeah Postgress actually has object store as well

artonio:
DB Schema
```typescript
interface Edge {
  id: string
  source: string
  target: string
  props?: object
}
interface Vertex {
  id: string
  props?: object
}
interface Graph {
  edges: Array<Edge>
  vertices: Array<Vertex>
}
interface User {
  graphs: Array<Graph>
  email: string
  password: string
}
```

PR #284, opened 2020-11-11
202 user module

artonio:
## Expected Behavior
Should be able to create a JWT access token by executing a GQL query in the playground
```GraphQL
query AccessToken {
  createAccessToken {
    accessToken
  }
}
```
After getting the token create a json object and paste into Http Headers in GraphQL Playground

```json
{
  "Authorization" : "Bearer <token>"
}
```

Now you should be able to execute the following Hasura generated query in GraphQL playground
```GraphQL
query GetUserHasuraQuery {
  user(where: {id: {_eq: 1}}) {
    username
  }
}
```

## Related Issue(s)

Fixes #


webberwang:
Closes #202

Issue #282, opened 2020-11-11
Re-implement renderer 


webberwang:
We want to use `cytoscape's` built in graph DFS traversal to re-implement the DOM building function

`libs/core/renderer/src/libs/renderer.tsx`

We currently use a recursive function to implement traversal, but it is not very flexible, we want to use graph traversal.

The DOM tree needs to be built from the leaf nodes upwards, so we use DFS and create a queue, then pop the queue in reverse order, then we call the iteratee to build the tree

`shared/factory/src/cytoscape.js` is the idea for constructing cytoscape from our data.

We also need to refactor how we store data, currently we use a tree approach to store our node data. If you look at the `*.data.ts` files inside our `ui-antd` lib, you notice we use children for relations between nodes.

we are currently adding a new structure to store data, `apps/api/graph/src/models`, you'll see we store a graph as vertices & nodes. We'll use cytoscape to construct the graph, using our own factory function `shared/factory/src/cytoscape.js`, we then use traversal to build our react dom tree

PR #294, opened 2020-11-13
feat(Re-implement renderer) (#282)

valtkachenko:
## Related Issue(s)

Fixes #

### DIFF
--- a/libs/core/renderer/src/libs/render-components-graph.tsx
+++ b/libs/core/renderer/src/libs/render-components-graph.tsx
@@ -1,0 +1,64 @@
+import React, { FunctionComponent } from 'react'
+import { makeCytoscape } from '../../../../shared/factory/src/cytoscape.js/factory'
+import { elementParameterFactory } from '@codelab/core/renderer'
+import { NodeType } from '@codelab/shared/interface/node'
+import { Props } from '@codelab/shared/interface/props'
+
+export interface IGraphVertex<
+  T extends NodeType = NodeType,
+  P extends Props = Props
+> {
+  id: string
+  type: T
+  props?: P
+  parent: string | null
+}
+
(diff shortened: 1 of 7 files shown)


libs/ui/antd/src/components/render-graph/RenderGraph.data.ts
```diff
@@ -0,0 +1,86 @@
...
+
+export const renderGraphData: IComponentGraph = {
+  vertices: [
+    {
+      id: '1',
+      type: NodeType.React_Html_Div,
+      props: {
+        style: { background: 'antiquewhite', padding: '12px' },
+      },
+      parent: null,
```
webberwang, line 12: the data structure shouldn't need a `parent` field, since `edges` contain `source` & `target`. we let cytoscape build the graph, then we should be able to access `node.parent()`

webberwang (changes requested):
- [ ] Rebase master, I would rebase daily since changes occur frequently (this should solve the lint error)
- [ ] Once you solve that, you'll see commit message lint error

libs/core/renderer/src/libs/render-components-graph.tsx
```diff
@@ -0,0 +1,64 @@
...
+  cy.elements().breadthFirstSearch({
+    roots: `#${root.id()}`,
+    visit: (node) => {
+      const [Component, props] = elementParameterFactory(node.data())
+      const element = React.createElement(Component, props)
+
+      // Add react element to node
+      node.attr('Component', element)
+
+      if (node.parent().nonempty()) {
```
webberwang, line 50: does `.parent()` require us to set `parent` field?

webberwang (changes requested):

PR #308, opened 2020-11-15
Kustomize api-services-graph

nutshell113:
## Current Behavior

`api-services-graph` deployment is not yet been deploy on cluster(s)

## Expected Behavior

Set `api-serivces-graph` up an running on both staging and prod clusters. Integrated well argocd must be required

## Related Issue(s)

https://github.com/codelab-ai/codelab.ai/issues/259


webberwang:
The web port is 3000, api port is 4004 in our `.env.example` (we can use 4000 too). let me know if new image is working

PR #327, opened 2020-11-19
ci: docker-compose fix missing env

nutshell113:


webberwang:
I ended up splitting the docker-compose files into 1 for building & 1 for running

PR #328, opened 2020-11-19
321 add unit test

### LINKED ISSUE #321: Add Unit Test
General idea https://github.com/codelab-ai/codelab.ai/wiki/RTM

You should have received an invitation to Airtable, which is where we keep our RTM

Feel free to make suggestions to the tests

artonio:
## Expected Behavior

Resolves #321

## Related Issue(s)

Fixes #


apps/api/graph/src/models/user/user.entity.spec.ts
```diff
@@ -0,0 +1,51 @@
+import * as bcrypt from 'bcrypt'
+import { CodelabValidationError } from '../../app/filters/CodelabValidationError'
+import { UserEntity } from './user.entity'
+
+describe('UserEntity class', () => {
+  it('Should hash password', async () => {
+    const u = new UserEntity()
```
webberwang, line 7: what do you think about switching to repository.create({}) for all entity, in let userRepository then in beforeAll assign

apps/api/graph/src/models/user/user.service.spec.ts
```diff
@@ -0,0 +1,114 @@
...
+
+const email = 'codelab@gmail.com'
+const password = 'password'
+
+describe('UserService', () => {
+  let userService: UserService
+  let repository: Repository<UserEntity>
+
+  beforeAll(async () => {
+    const module = await Test.createTestingModule({
```
webberwang, line 26: could we extract to `libs/api/providers/config/src/test/config-typeorm.service.ts`, and load using config service

webberwang (changes requested):

apps/api/graph/src/models/user/user.entity.spec.ts
artonio: i see what you mean, makes sense, will do

webberwang (approved):

PR #346, opened 2020-11-25
feat(apps-api-graph): revised schema

### LINKED ISSUE #337: Create typeorm models
User - App (1-M)
App - Page (1-M)
App - Graph (1-M)
Page - Graph (1-M) These must be layout graphs, usually have only 1

The Page's Graph contains only edges that serve to combine graphs

New_Project_1_—_Edited

artonio:
- rewrote moveVertex and unit tests

## Expected Behavior

Resolves #337 
## Related Issue(s)

Fixes #


apps/api/graph/src/models/graph/graph.service.ts
```diff
@@ -1,16 +1,69 @@
...
     return this.graphEntityRepository.find()
   }
+
+  async findOne(graphId: string): Promise<GraphEntity> {
+    return this.graphEntityRepository.findOneOrFail(graphId, {
+      relations: ['vertices', 'edges'],
+    })
+  }
+
+  async moveVertex(src: string, target: string): Promise<GraphEntity> {
```
webberwang, line 32: instead of `string` use `VertexID`

on second thought, we should perform the move using cytograph https://js.cytoscape.org/#collection/graph-manipulation/eles.move

we should write a graph sync function that compares 2 graphs and update the db accordingly https://js.cytoscape.org/#collection/comparison

If cytoscape doesn't support a top level graph comparison, we can use "dynamic programming" to write a recursive comparison function using "memoization".

Tree's are especially good for dynamic programming because we can divide the problem into subsets (sub trees in our case). The memoization will cache the comparison results if the subtree hasn't changed.

If we have this sync function, then all CRUD's on Graph will be trivial since it would be automatically synced. We just cytoscape graph methods to modify our graph

apps/api/graph/src/models/graph/graph.service.unit.spec.ts
```diff
@@ -0,0 +1,162 @@
...
+      ],
+    }).compile()
+
+    graphService = m.get<GraphService>(GraphService)
+    graphService.onModuleInit()
+    edgeService = m.get<EdgeService>(EdgeService)
+    graphRepositoryMock = m.get(getRepositoryToken(GraphEntity))
+  })
+
+  it('Should throw error if graphs are different', async () => {
```
webberwang, line 51: for future let's hold off on writing the tests, if we use cytoscape move method we wouldn't really need those specs

also there may not be a need for testing on different graph since the UI wouldn't allow that to be possible

webberwang (changes requested):

apps/api/graph/src/models/graph/graph.entity.ts
```diff
@@ -160,29 +152,48 @@ export class GraphEntity {
...
+  moveUsingCytoscape(
+    cy: cytoscape.Core,
+    source: VertexID,
+    target: VertexID,
+  ): cytoscape.Core {
+    const cyJson = cy.json() as any
+    const cyEdges = cyJson.elements.edges.map((e: any) => e.data)
+
+    const sourceEdge = cyEdges.find((edge: ICyEdge) => {
+      return edge.target === source
```
webberwang, line 169: there could be multiple target with same source id

Issue #354, opened 2020-12-01
Throw error if .env exists


webberwang:
Throw inside `main.ts` of graph nestjs app

PR #355, opened 2020-12-01
build(libs-api-providers): #354 throw error if .env exists

aburtsev:
CLOSES: #354

## Related Issue(s)

Fixes #

### DIFF
--- a/libs/api/providers/config/src/checkDotEnvExistence.ts
+++ b/libs/api/providers/config/src/checkDotEnvExistence.ts
@@ -1,8 +1,9 @@
 import findConfig from 'findup-sync'
 
+const DEFAULT_DOTENV_FILE = '.env'
+
 export const checkDotEnvExistence = () => {
-  const defaultDotevFile = '.env'
-  const path = findConfig(defaultDotevFile)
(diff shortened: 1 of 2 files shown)


libs/api/providers/config/src/checkDotEnvExistence.ts
```diff
@@ -0,0 +1,12 @@
+import findConfig from 'findup-sync'
+
+export const checkDotEnvExistence = () => {
+  const defaultDotevFile = '.env'
```
webberwang, line 4: maybe rename to `DEFAULT_DOTENV_FILE` & move outside of function

libs/api/providers/config/src/config.module.ts
```diff
@@ -9,6 +10,7 @@ import {
 import { ConfigJwtService } from './jwt'
 import { envPath, envs, isDev } from '@codelab/shared/utils'
 
+checkDotEnvExistence()
```
webberwang, line 13: I think inside constructor could be better

webberwang (changes requested):

PR #360, opened 2020-12-04
348 cytoscape move method

### LINKED ISSUE #348: Refactor to use Cytoscape move method
We'll have a single move function with 2 inner functions. Move is only called if different parent.

Also instead of passing `cy` as a parameter, we should add `cy` as a getter property

Makes sense as well, since `this` is our `Graph` class, we can easily map to cy representation at anytime during runtime

We'll also have a getter for `d3`.

1) Cytoscape move

2) Edge order update

Untitled_Artwork 2

-------------------------------

The single source of truth is our Graph vertices/edges. We have `get cy()` to convert to cy at runtime. Anytime cy makes a change, like `move` it updates vertices/edges on Graph class. 

Which means we only need a mapper from Graph to Cy, as Cy operates on Graph data already. We don't need to convert from Cy to Graph back.

We'll do the same for `get d3()`, which we can get data to plug into D3 ui.

Graph is the source of truth that we can easily convert to any other representation.

artonio:
## Related Issue(s)

Fixes #348


apps/api/graph/src/models/graph/graph.entity.ts
```diff
@@ -134,66 +134,52 @@ export class GraphEntity {
     return index !== -1
   }
 
-  public moveVertex(sourceId: VertexID, targetId: VertexID) {
-    const sourceIndexE = this.edges.findIndex((e: EdgeEntity) => {
-      return e.target === sourceId
-    })
+  moveVertex(source: VertexID, target: VertexID) {
```
webberwang, line 137: it would be nice if we could access repository here, but that would break the dependency rule, since entity is more inner than repository.

so previously you had `source: Vertex` here & I think that is better. we could create a separate `GraphService` class that calls this `moveVertex` method, and our service class would have access to repository. This service class would be considered a domain service, which helps to retrieve the vertex object.

our entity should be domain rich, and the concern of searching for the element doesn't really belong here. also any error handling should be handled 1 layer outwards in the domain service level. this way we keep any non-essential code out of here

webberwang (changes requested):

apps/api/graph/src/models/graph/sample.data.ts
```diff
@@ -0,0 +1,155 @@
...
+    a,
+    b,
+    c,
+    d,
+    e,
+  }
+}
+
+export const createAtoIVertices = () => {
+  const root = new VertexEntity()
```
webberwang, line 93: If we create our own domain model, we can have built in invariance into static factory methods. We could have a testing factory method that doesn't check for `type`, while a normal create does

libs/ui/d3/src/cytoscape/Cy.data.ts
```diff
@@ -0,0 +1,34 @@
+import { EdgeDefinition, NodeDefinition } from 'cytoscape'
```
webberwang, line 1: we should create using our entities & map to cy representation

libs/ui/d3/src/cytoscape/Cytoscape.stories.tsx
```diff
@@ -0,0 +1,131 @@
+import React, { useEffect, useState } from 'react'
+import { CyGraph } from './CyGraph'
+import { CyGraphService } from './CyGraph.service'
+import { ICyGraphProps } from './ICyGraphProps'
+
+export default {
```
webberwang, line 6: I'm looking into storybook fixtures for better re-use of component & sharing of data

webberwang (commented):
very nice work on getting all the tests to pass, and creating the custom storybook component!

I have some ideas I'd like to introduce in terms of domain architecture, but I'll need to write up some docs to outline my motivation

I've added these comments here so we can refer back later on, I'll merge these for now

the nest task should get you to see what i had in mind for the new project structure

PR #378, opened 2020-12-10
test(apps-api-codelab): create user e2e test

### LINKED ISSUE #369: Add GraphQL e2e testing
This is actually a rewrite of the project in a separate folder, we've added 1 feature currently `createUser`. Update with new `.env.example`. 

We break down features into use cases in domain driven design, I'll add the docs soon for the concepts behind this restructure.

Add an e2e test for the resolver https://github.com/codelab-ai/codelab.ai/blob/d933bbf94d989651ed391e26f1d3a9080e4484be/libs/ddd/modules/users/presentation/controllers/UserResolver.ts#L20-L30

The inner use case is already tested using https://github.com/codelab-ai/codelab.ai/blob/master/libs/ddd/modules/users/core/application/useCases/createUser/CreateUserUseCase.spec.ts. 

The outer involves more adapters such as GraphQL, write a test for `createUser` & it should return a user. It currently doesn't actually create a user in the database, but does return data.

artonio:
## Related Issue(s)

Fixes #369


libs/ddd/shared/infrastructure/config/TestDatabaseConfig.ts
```diff
@@ -20,4 +20,16 @@ export class TestDatabaseConfig {
   public static readonly DB_NAME: string = get('TEST_DB_NAME')
     .required()
     .asString()
+
+  public static readonly TYPEORM_SYNCHRONIZE: boolean = get(
```
webberwang, line 24: let's hard code these

apps/api/codelab/src/app/user/user.e2e.spec.ts
```diff
@@ -0,0 +1,39 @@
+import { INestApplication } from '@nestjs/common'
+import { Test } from '@nestjs/testing'
+import request from 'supertest'
+import { AppModule } from '../app.module'
+
+describe('User E2E test', () => {
```
webberwang, line 6: we'll test UserModule, move test to `ddd/modules/users/application/useCases/createUser/CreateUserUseCase.e2e.spec.ts`

webberwang (changes requested):

PR #398, opened 2020-12-15
396 cqrs task

### LINKED ISSUE #396: CQRS Task
- [x] Refactor update / delete (https://codelabstudios.slack.com/archives/D01CW67QMRC/p1607967231005600)
- [x] Add 2 e2e test for delete & update for user
- [x] Add crud for vertex
- [x] Add crud for edge
- [ ] Add crud for graph

- `this.addChildNode(parent: VertexID, vertex: Vertex, order?: number)` order defaults to last, for now default to last only
- graphql resolver class will be called `GraphCommandQueryAdapter.ts`, while the method name is `addChildNode`
- graphql input will contain graphId

----------

## Note the types in each layer

The repository shouldn't do validation, leave that to the service class. For example, we'll check in service class whether user exists, then once repository is called, we assume user exists.

TypeOrmUserRepositoryAdapter_ts_—_codelab

Repository takes in a `User` object, we call `toPlain()` to cast to serialized object, then hydrate back to `User` object and return.  ie. `createUser`

The input and output for repo are both domain model types

CreateUserResponse_ts_—_codelab

artonio:
## Related Issue(s)

Fixes #396

### DIFF
--- a/libs/modules/users/src/core/adapters/UserRepositoryPort.ts
+++ b/libs/modules/users/src/core/adapters/UserRepositoryPort.ts
@@ -1,13 +1,18 @@
-import { DeleteResult } from 'typeorm'
-import { FindUserBy } from '../../common/CommonTypes'
+import {
+  FindUserBy,
+  FindUserByEmail,
+  FindUserByID,
+} from '../../common/CommonTypes'
 import { User } from '../domain/user'
-import { UserEmail } from '../domain/user-email'
 import { TypeOrmUser } from '@codelab/backend'
 
 export interface UserRepositoryPort {
-  findUser(by: FindUserBy): Promise<TypeOrmUser>
+  // findUser(by: FindUserBy): Promise<TypeOrmUser | undefined>
   createUser(user: User): Promise<User>
-  deleteUser(email: UserEmail): Promise<DeleteResult>
-  updateUser(user: User): Promise<User>
+  deleteUser(user: TypeOrmUser): Promise<Array<TypeOrmUser>>
+  updateUser(existingUser: TypeOrmUser, updatedUser: User): Promise<TypeOrmUser>
   exists(by: FindUserBy): Promise<boolean>
+
+  findUserById(by: FindUserByID): Promise<TypeOrmUser | undefined>
+  findUserByEmail(by: FindUserByEmail): Promise<TypeOrmUser | undefined>
 }
--- a/libs/modules/users/src/core/adapters/UserRepositoryPort.ts
+++ b/libs/modules/users/src/core/adapters/UserRepositoryPort.ts
@@ -8,4 +8,5 @@ export interface UserRepositoryPort {
   updateUser(existingUser: User, updatedUser: User): Promise<User>
   exists(by: FindUserBy): Promise<boolean>
   findUser(by: FindUserBy): Promise<Option<User>>
+  findAll(): Promise<Array<User>>
 }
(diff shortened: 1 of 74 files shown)


libs/modules/users/src/core/adapters/UserRepositoryPort.ts
```diff
@@ -1,13 +1,18 @@
...
-import { TypeOrmUser } from '@codelab/backend'
 
 export interface UserRepositoryPort {
-  findUser(by: FindUserBy): Promise<TypeOrmUser>
+  // findUser(by: FindUserBy): Promise<TypeOrmUser | undefined>
   createUser(user: User): Promise<User>
-  deleteUser(email: UserEmail): Promise<DeleteResult>
-  updateUser(user: User): Promise<User>
+  deleteUser(user: User): Promise<Option<User>>
+  updateUser(existingUser: User, updatedUser: User): Promise<User>
```
webberwang, line 13: We'll do the validation in the service, so all user passed here exists already

libs/modules/users/src/core/adapters/UserRepositoryPort.ts
```diff
@@ -1,13 +1,18 @@
...
   exists(by: FindUserBy): Promise<boolean>
+
+  findUserById(by: FindUserByID): Promise<Option<User>>
```
webberwang, line 16: `findUserBy(by: FindUserBy)`

libs/modules/users/src/core/application/handlers/DeleteUserCommandHandler.ts
```diff
@@ -15,14 +16,14 @@ export class DeleteUserCommandHandler
...
-  async execute({ request }: DeleteUserCommand): Promise<DeleteResult> {
+  async execute({ request }: DeleteUserCommand): Promise<UserUseCaseDto> {
     const deleteUserResults = await this.service.execute(request)
 
     return fold(
       (errors) => {
         throw errors
       },
-      (results: Result<DeleteResult>) => results.value,
+      (results: Result<User>) => (results.value as unknown) as UserUseCaseDto,
```
webberwang: cast to unknown is required? wouldn't be call `results.value.toPlain()`

libs/modules/users/src/core/application/services/DeleteUserService.ts
```diff
@@ -11,13 +13,26 @@ export class DeleteUserService implements DeleteUserUseCase {
   constructor(private readonly userRepository: UserRepositoryPort) {}
 
   async execute(request: DeleteUserRequest): Promise<DeleteUserResponse> {
-    const userEmail = new UserEmail({ value: request.email })
-    const result = await this.userRepository.deleteUser(userEmail)
+    const existingUser: Option<User> = await this.userRepository.findUserByEmail(
```
webberwang, line 16: would change according to interface

libs/modules/users/src/core/application/useCases/UserUseCase.i.spec.ts
```diff
@@ -32,7 +34,7 @@ describe('CreateUserUseCase', () => {
 					mutation {
             createUser(user:
               {
-                email: "admin@codelab.ai",
+                email: "${email}",
```
webberwang, line 37: for tests i generally hard code the values

webberwang (changes requested):

libs/modules/users/src/core/application/handlers/DeleteUserCommandHandler.ts
artonio: no need to cast, was a mistake, fixed it in latest push

PR #410, opened 2020-12-25
406 stroybook loader

aburtsev:
## Related Issue(s)

Fixes #


libs/alpha/ui/antd/src/apolloLoader.ts
```diff
@@ -0,0 +1,38 @@
...
+        id
+        source
+        target
+        props
+      }
+    }
+  }
+`
+
+export const queryGraph = async (graphLabel: string) => {
```
webberwang, line 24: should use our own query helper

webberwang (changes requested):

libs/alpha/ui/antd/src/apolloLoader.ts
```diff
@@ -0,0 +1,38 @@
+import { gql } from '@apollo/client'
+import { getApolloClient } from '@codelab/alpha/ui/hoc'
+
+const GraphByLabelDocument = gql`
+  query graphByLabel($label: String) {
```
webberwang, line 5: move this to same file as the component

webberwang (changes requested):

Issue #401, opened 2020-12-17
Create Signup + Login


webberwang:
- [x] rename `CreateUserUseCase` to `SignupUserUseCase`
- [x] create `LoginUserUseCase`
- [ ] connect to backend
- [ ] add cypress e2e tests

work on generator at the same time, use case generator should insert di token using ast util to insert method to class.

generator should actually replace files that don't require manual editing, core logic files shouldn't be replaced.

for example `CreateUserRequest` should be replaced because it could be generated in its final form

add a remove use case generator as well, which deletes files and remove di token

add section in the use case generator for `InputType` `ObjectType` request & dto respectively, user should be able to specify key value pair, whichever ui the vscode plugin gives us (will need to check that out on vscode)

PR #413, opened 2020-12-27
411 use case generator

### LINKED ISSUE #411: UseCase generator
Checkout nrwl plugin for vscode, it gives us a ui form to configure generator.

- [x] Add select field for `Command` or `Query`
- [x] Should insert a method on graphql resolver class, generator either Mutation or Query, should 
- [x] Should Insert UseCase to DITokens
- [x] Should Modify Module useCaseProviders and handlerProviders array

- [ ] delete use case generator
- [ ] should undo all inserts & and remove files

artonio:
## Related Issue(s)

Fixes #411 and #401 and #420 and #422


libs/tools/plugins/codelab/src/schematics/library/use-case/helpers/modify-files.ts
```diff
@@ -0,0 +1,148 @@
...
+          resultChanges.push(insertChange)
+        }
+      }
+    })
+  })
+
+  return resultChanges
+}
+
+const addToClass = (
```
webberwang, line 39: would this method help? https://github.com/nrwl/nx/blob/9923ce93d57c25624002d428348561e2568ad8ae/packages/workspace/src/utils/ast-utils.ts#L250-L267

libs/tools/plugins/codelab/src/schematics/library/use-case/helpers/modify-files.ts
```diff
@@ -0,0 +1,148 @@
...
+
+        resultChanges.push(insertChange)
+      }
+    })
+  })
+
+  return resultChanges
+}
+
+const insertIntoModule = (host: Tree, options: NormalizedSchema): void => {
```
webberwang, line 65: the nestjs schematics has to insert import to module as well, we could take some inspiration from that?

https://github.com/nestjs/schematics/blob/6e6855386742bf7c886cd129c4886f720bc63258/src/lib/provider/provider.factory.ts#L75-L95

Or is our use case more complicated because it's separated out to an array variable?

webberwang (changes requested):

libs/tools/plugins/codelab/src/schematics/library/use-case/helpers/modify-files.ts
artonio: exactly, we have a custom array that needs to be modified. actually angular schematics library is using the same method to insert code into arrays inside the module. I do the same thing, I lookup the position to insert code in by using TypeScript Compiler API, (the same API that is used by linters) and InsertChange which return Array<Change> just like addMethod function

libs/tools/plugins/codelab/src/schematics/library/use-case/helpers/modify-files.ts
artonio: no, because we are not inserting a method here

PR #444, opened 2021-01-07
414 front end auth second

### LINKED ISSUE #414: Front end auth
- [ ] Use json schema react form, import generated schema
- [ ] Connect to backend with graphql, use generated gql & apollo helper methods

The register use case should return jwt token to be stored in local storage, then sent as header on each api call. the backend will then take this jwt token and authorize user

image

antonmihaylov:
## Related Issue

Fixes #414


webberwang:
- [ ] Refactor to https://github.com/codelab-ai/codelab.ai/blob/master/documentation/getting-started/3-project-architecture.md#input-vs-request

`nx build tools-generators-graphql` will generate the types

- [ ] Add a 1 second delay to each request, this way we can test out our UI's
- [ ] After submit, the the button should be disabled loading state
- [ ] On success the button state will revert as the modal closes
- [ ] Then either a success or error notification will appear

libs/modules/user-stories/src/useCases/registerUser/RegisterUserForm.tsx
```diff
@@ -1,26 +1,36 @@
...
+import { RegisterUserRequest } from '../../../../user/src/core/application/useCases/registerUser/RegisterUserRequest'
 import { useUserMachine } from '../../store'
 import { requestJsonSchema } from '@codelab/tools/generators/json-schema'
 
+const Form = withTheme(AntDTheme)
+
 export const RegisterUserForm = ({
   formId,
-  hasSubmitButton = true,
+  hasSubmitButton,
```
webberwang, line 13: we always show submitButton, but change to disabled loading state

libs/modules/user-stories/src/useCases/registerUser/RegisterUserForm.tsx
```diff
@@ -1,26 +1,36 @@
...
 }: {
   formId?: string
   hasSubmitButton?: boolean
+  submitBtnRef?: React.RefObject<HTMLButtonElement>
 }) => {
   const user = useUserMachine()
 
-  const onSubmit = ({ formData }: any) => {
+  // TODO: register fields are not required, perhaps find a way to make them required in the schema through the generator
+  const [formData, setFormData] = useState<RegisterUserRequest>({
```
webberwang, line 23: is there a reason we need this useState?

libs/modules/user-stories/src/useCases/registerUser/RegisterUserModal.tsx
```diff
@@ -1,32 +1,42 @@
...
   const root = useRootMachine()
   const user = useUserMachine()
-  const sharedModalProps: ModalProps = {
-    visible: user.state.value.guest.signingUp === 'idle',
-    onCancel: () => root.send('ON_MODAL_CANCEL'),
-    onOk: () => root.send('ON_MODAL_OK'),
-  }
+  const submitBtnRef = createRef<HTMLButtonElement>()
+
+  const error =
```
webberwang, line 12: image

webberwang (changes requested):

libs/modules/user-stories/src/useCases/registerUser/RegisterUserForm.tsx
antonmihaylov: Yes, without the state the form input values get reset on each render. https://github.com/rjsf-team/react-jsonschema-form/issues/391

libs/modules/user-stories/src/useCases/registerUser/RegisterUserModal.tsx
antonmihaylov: That would mean we will handle errors globally, not in the form components, right?

libs/modules/user-stories/src/useCases/registerUser/RegisterUserForm.tsx
antonmihaylov: We don't actually, I'm hiding it here, so that the modal "OK" button acts as a submit button

libs/modules/user-stories/src/useCases/registerUser/RegisterUserModal.tsx
webberwang: yeah that's right

PR #461, opened 2021-01-11
448 create and delete app

### LINKED ISSUE #448: Create & Delete App
/apps should have a 3 column list of cards, each card being a clickable app that takes us to /apps/[id] with main content containing title for now

top right of each card should have a ellipsis icon that is clickable, which should bring up a popover menu with delete button

antonmihaylov:
Fixes #448

### DIFF
--- a/libs/modules/app-stories/src/useCases/deleteApp/useDeleteApp.tsx
+++ b/libs/modules/app-stories/src/useCases/deleteApp/useDeleteApp.tsx
@@ -1,0 +1,56 @@
+import { Modal } from 'antd'
+import { useEffect, useRef } from 'react'
+import { useAppMachine } from '@codelab/modules/app-stories'
+
+export const useDeleteApp = () => {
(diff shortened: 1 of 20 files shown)


libs/modules/app-stories/src/useCases/deleteApp/useDeleteApp.tsx
```diff
@@ -0,0 +1,56 @@
...
+  const destroyModal = () => {
+    modalRef.current?.destroy()
+    modalRef.current = undefined
+  }
+
+  const onDeleteConfirmed = (id: string) =>
+    appMachine.send({ type: 'ON_DELETE_APP', data: { id } })
+
+  const handleDeleteClick = (app: { id: string; title: string }) => {
+    modalRef.current = Modal.confirm({
```
webberwang, line 19: why not move this to useEffect in the popover component? this way as we open close popover react can manage the ref lifecycle for us

PR #499, opened 2021-01-21
476 nextjs page auth guard

### LINKED ISSUE #476: Nextjs Page Auth Guard
Goal is the create an auth guard for pages, so we show the users the home page if they're accessing `/apps` without logging in.

I first found https://github.com/nextauthjs/next-auth. Seems like a cool concept but we are using xstate for session data & have our own api's for authentication. Next-auth would be good to prototype a brand new project with any existing infrastructure.

I think we should fetch user data server side so we don't get a flicker for page (Next.js SSR feature). 

This is an auto-generated server side query to be used with SSR (https://graphql-code-generator.com/docs/plugins/typescript-apollo-next)

image

The issue is that we don't have access to local storage at server side.

May need to use cookies instead per this discussion

https://github.com/vercel/next.js/discussions/17247

This method would also mean we run the auth check each time we load a page, incase jwt expires. Currently we only update user data on login/register, which may not be as ideal

Let me know your thoughts on how to implement auth guard for nextjs pages

antonmihaylov:
Fixes #476


libs/frontend/src/model/store/apollo/apolloClient.tsx
```diff
@@ -9,24 +9,32 @@ import {
...
 import { NextPage } from 'next'
 import React from 'react'
-import { combinedLink } from './links/combinedLink'
+import { isServer } from '../../../config'
+import { getCombinedLink } from './links/combinedLink'
 
 export const getApolloClient = (
   ctx?: any,
   initialState?: NormalizedCacheObject,
+  token?: string | undefined,
```
webberwang, line 18: we can attach header to ctx, I think ctx.req.headers? not sure exactly the shape, but we should remove `any` and type it.

libs/frontend/src/model/store/apollo/apolloClient.tsx
```diff
@@ -9,24 +9,32 @@ import {
...
 ) => {
   const cache = new InMemoryCache().restore(initialState || {})
 
   return new ApolloClient({
-    link: combinedLink,
+    link: getCombinedLink(token),
```
webberwang, line 23: forward the header to apollo client instead

image

libs/frontend/src/infrastructure/auth/withAuthServerSideProps.ts
```diff
@@ -0,0 +1,71 @@
...
+import { ssrGetMe } from '../../../../../apps/web/src/apollo/apollo-page.generated'
+import { AUTH_TOKEN_COOKIE } from '@codelab/frontend'
+
+const getAuthResult = async (cookie: string | undefined) => {
+  if (cookie) {
+    const cookies = ParseCookies.parse(cookie)
+    const token = cookies ? cookies[AUTH_TOKEN_COOKIE] : undefined
+
+    try {
+      return await ssrGetMe.getServerPage({}, undefined, token)
```
webberwang, line 14: we shouldn't need to pass token here if headers is forwarded to apolloClient

webberwang (changes requested):

libs/frontend/src/infrastructure/auth/withAuthServerSideProps.ts
```diff
@@ -0,0 +1,71 @@
...
+  if (getServerSidePropsFunc) {
+    const originalData = await getServerSidePropsFunc(context, result)
+
+    return mergeDeep(result, originalData)
+  }
+
+  return result
+}
+
+export const withAuthGuardServerSideProps = (redirectTo = '/') => {
```
webberwang, line 58: I think passing callback is better

webberwang (changes requested):

apps/web/src/apollo/apollo-page.generated.tsx
```diff
@@ -10,8 +10,9 @@ import { getApolloClient } from '@codelab/frontend'
 export async function getServerPageGetMe(
   options: Omit<Apollo.QueryOptions<Types.GetMeQueryVariables>, 'query'>,
   ctx?: any,
+  token?: string,
```
webberwang, line 13: this function can't be edited as its generated

webberwang (changes requested):

libs/frontend/src/infrastructure/auth/withAuthServerSideProps.ts
```diff
@@ -0,0 +1,71 @@
...
+  ReturnType<typeof getAuthResult>
+>
+
+export const withAuthServerSideProps = (
+  getServerSidePropsFunc?: (
+    context: GetServerSidePropsContext,
+    intermediateResult: AuthServerSideResult,
+  ) => Promise<GetServerSidePropsResult<any>>,
+) => async (context: GetServerSidePropsContext<any>) => {
+  const cookie = context.req?.headers?.cookie
```
webberwang, line 45: better to pass context around, defer de-structure to the function that needs the cookie

libs/frontend/src/infrastructure/auth/withAuthServerSideProps.ts
```diff
@@ -0,0 +1,71 @@
...
+
+  const result = await getAuthResult(cookie)
+
+  if (getServerSidePropsFunc) {
+    const originalData = await getServerSidePropsFunc(context, result)
+
+    return mergeDeep(result, originalData)
```
webberwang, line 52: can't use `...` merge?

webberwang (changes requested):

libs/frontend/src/infrastructure/auth/authTokenStorage.ts
```diff
@@ -0,0 +1,24 @@
+import Cookie from 'js-cookie'
+import { isServer } from '@codelab/frontend'
+
+export const AUTH_TOKEN_COOKIE = 'AUTH_JWT'
+
+export const getAuthToken = () => {
+  if (isServer) return null
```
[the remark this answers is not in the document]
antonmihaylov: We do, js-cookie is just a wrapper around client-side cookies. It uses ```document.cookie``` under the hood. If we're on the server, there is no document

libs/frontend/src/infrastructure/auth/withAuthServerSideProps.ts
antonmihaylov: If the request fails, e.g. if the user is not authenticated it throws an error and doesn't display the page at all.

libs/frontend/src/infrastructure/auth/withAuthServerSideProps.ts
antonmihaylov: Nope, it's nested. The ```data``` object won't get merged if we use ```...```

libs/frontend/src/infrastructure/auth/withAuthServerSideProps.ts
antonmihaylov: I made this more so that we have an example. We can be flexible with ```withAuthServerSideProps```, where you get the user as a callback param

libs/frontend/src/infrastructure/auth/authTokenStorage.ts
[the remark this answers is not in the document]
webberwang: would this package help? https://github.com/reactivestack/cookies/tree/master/packages/react-cookie/#readme

libs/frontend/src/model/store/apollo/apolloClient.tsx
```diff
@@ -9,24 +9,35 @@ import {
 } from '@apollo/client'
 import { NextPage } from 'next'
 import React from 'react'
+import { isServer } from '../../../config'
 import { combinedLink } from './links/combinedLink'
 
+export interface ApolloContext {
```
webberwang, line 15: is there a context type from apollo?

libs/frontend/src/model/store/apollo/apolloClient.tsx
```diff
@@ -9,24 +9,35 @@ import {
...
-  ctx?: any,
+  ctx?: ApolloContext | null,
   initialState?: NormalizedCacheObject,
 ) => {
   const cache = new InMemoryCache().restore(initialState || {})
 
   return new ApolloClient({
     link: combinedLink,
     cache,
+    ssrMode: isServer, // Disables forceFetch on the server (so queries are only run once)
```
webberwang, line 28: wouldn't we want to fetch on server so we can redirect serverside?

libs/frontend/src/infrastructure/auth/authTokenStorage.ts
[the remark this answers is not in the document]
antonmihaylov: Not much. It's for using cookies in a component as far as i see. And it probably is still a wrapper around document.cookie

libs/frontend/src/model/store/apollo/apolloClient.tsx
antonmihaylov: Not sure. But I don't know think it would make sense. We can add anything to the context with a setContext link, like in authLink

libs/frontend/src/model/store/apollo/apolloClient.tsx
antonmihaylov: Got thay from Apollo docs https://www.apollographql.com/docs/react/performance/server-side-rendering/

There are some benefits to ssrMode which we can use later on perhaps

PR #501, opened 2021-01-22
497 spec tests

### LINKED ISSUE #497: Fix all specs for integration tests
Make all integration tests pass. We'll remove skip in a separate task. Goal is to pass ci tests so we can merge.

You can skip the `MoveNode` for now as we have a separate issue for that.

artonio:
## Related Issue(s)

Fixes #497

### DIFF
--- a/apps/api/codelab/tsconfig.app.json
+++ b/apps/api/codelab/tsconfig.app.json
@@ -2,7 +2,7 @@
   "extends": "./tsconfig.json",
   "compilerOptions": {
     "outDir": "../../../dist/out-tsc",
-    "types": ["node", "@xstate/react", "object-mapper"],
+    "types": ["node", "@xstate/react", "object-mapper", "@rjsf/antd"],
     "jsx": "preserve",
     "esModuleInterop": true,
     "module": "CommonJS",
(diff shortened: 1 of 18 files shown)


libs/backend/src/infrastructure/graphql/test/TestClientApollo.ts
```diff
@@ -0,0 +1,106 @@
...
+    })
+    this.setClient()
+  }
+
+  getClient() {
+    return this.client
+  }
+
+  setClient() {
+    this.client = new ApolloClient({
```
webberwang, line 77: is there a need for a class? couldn't we build from ApolloClient

libs/backend/src/infrastructure/graphql/test/app/AppTestClient.ts
```diff
@@ -0,0 +1,22 @@
+import { FetchResult } from '@apollo/client'
+import gql from 'graphql-tag'
+import { TestClientBase } from '../TestClientBase'
+
+export interface ICreateApp {
+  createApp: { id: string; title: string }
+}
+
+export class AppTestClient extends TestClientBase {
```
webberwang, line 9: seems like an extra step to use the client, why not use our mutate method in utils `mutate(client, options)`

webberwang (changes requested):

PR #516, opened 2021-01-25
ci: join env settings to 1 step

nutshell113:
## Current Behavior

After vault ENVs are generated, all steps which needed to use vault env must source ~/.env on its commands 

## Expected Behavior

We should inject vault envs to $BASH_ENV to save time to run CI

## Related Issue(s)

https://github.com/codelab-ai/codelab.ai/issues/515

Fixes #


nutshell113:
@webberwang : I joined various of ENVs set to 1 steps / Also join vault-env to BASH_ENV already. CI worked well already, please take a look, thanks!

Issue #522, opened 2021-01-26
Integration tests for use cases


webberwang:
- [x] MoveVertex spec (don't test for order for now)
- [x] Remove all skips
- [x] use setup/teardown, use generated gql, keep supertest for now
- [x] When asserting try to assert object equals using `toStrictEqual` or `matchObject` etc

- [x] DeletePage should throw error if last page, should throw friendly error like can't delete last page.

PR #560, opened 2021-02-02
feat(apps-api-codelab): global error handling wip

### LINKED ISSUE #521: Refactor to global error handling for prisma
Let's do a prototype of what global handler with prisma code messages would look like.

This helps debugging as the error we throw in service may not be indicative of the actual error, since we're catching the generic error, which could be anything.

artonio:
## Related Issue(s)

Fixes #521 #522 #567

### DIFF
--- a/libs/backend/src/infrastructure/persistence/prisma/PrismaService.ts
+++ b/libs/backend/src/infrastructure/persistence/prisma/PrismaService.ts
@@ -1,18 +1,46 @@
-import { OnModuleDestroy, OnModuleInit } from '@nestjs/common'
+import { Logger, OnModuleDestroy, OnModuleInit } from '@nestjs/common'
+import { PrismaDelete } from '@paljs/plugins'
 import { PrismaClient } from '@prisma/client'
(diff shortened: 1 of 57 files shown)


libs/backend/src/infrastructure/persistence/prisma/PrismaService.ts
```diff
@@ -7,8 +13,30 @@ import { testDatabaseConfig } from '../config/DbConfig'
 export class PrismaService
   extends PrismaClient
   implements OnModuleInit, OnModuleDestroy {
+  logger: Logger = new Logger('PrismaService', true)
```
webberwang, line 16: Why not inject logger instead of creating a new instance each time

webberwang (changes requested):

libs/backend/src/infrastructure/persistence/prisma/PrismaService.ts
artonio: Sure, i will do that

PR #642, opened 2021-03-11
631 generator misc

### LINKED ISSUE #631: Generator misc
image

take code from `tools-generators-json-schema` for inspiration, we'll add to `tools-generators-crawler`.

- [ ] Implement file crawler & process multiple input files for export into single file
- [x] Fix spec error

artonio:
## Related Issue(s)

Fixes #631


libs/modules/graph/src/core/application/useCases/updateVertex/UpdateVertexInput.ts
```diff
@@ -1,52 +1,243 @@
...
-  @Required()
+  @RjsfGridProp({ row: 0, span: 24, required: true })
   declare vertexId: string
 
   @Field(() => String, { nullable: true })
-  @Enum(VertexType)
+  @RjsfGridProp({ row: 1, span: 24, required: true, enum: VertexType })
   declare type?: VertexType
 
+  @RjsfGridProp({
```
webberwang, line 43: These props should be like CreateVertexInput

libs/alpha/ui/antd/src/components/affix/Affix.input.ts
```diff
@@ -1,38 +1,41 @@
...
+import { Description, } from '@tsed/schema'
 import { AffixProps as AntAffixProps } from 'antd/lib/affix'
+import { RjsfGrid, RjsfGridProp } from '@codelab/tools/generators/form-decorator';
 
-class Props implements Omit<AntAffixProps, 'children'> {
-  @Optional()
-  @Property('string')
-  @Description('Offset from the bottom of the viewport (in pixels)')
+@RjsfGrid({
+  'ui:spacing': 16,
```
webberwang, line 6: Set default in decorator so we can avoid this

webberwang (changes requested):

PR #701, opened 2021-03-29
feat(circleci): include beanstalk deploy in cicd (#700)

svitalii33:
## Current Behavior

There is no beanstalk deployment today.

## Expected Behavior

This will trigger deploying the updates to beanstalk on tag releases.
## Related Issue(s)

Fixes #

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -76,7 +76,7 @@ executors:
   docker-node:
     <<: *executor-config
     docker:
-      - image: cimg/node:14.15.0-browsers
+      - image: cimg/node:14.16.0-browsers
     resource_class: large
   vault-img:
     <<: *executor-config
@@ -130,8 +130,8 @@ commands:
       - run:
           name: 'Set node version'
           command: |
-            nvm install 14.13.0
-            nvm alias default 14.13.0
+            nvm install 14.16.0
+            nvm alias default 14.16.0
             node -v
 
   setup-workspace:
@@ -306,6 +306,35 @@ jobs:
           channel: $SLACK_CHANNEL_UPDATE
           mentions: $SLACK_USER
 
+  deploy-to-ebs:
+    <<: *executor-params
+    steps:
+      - setup-workspace
+      - run:
+          name: Installing dependencies
+          command: |
+            sudo apt-get -y -qq update
+            sudo apt-get install python3-pip awscli build-essential
+            pip3 install awsebcli --upgrade --user
+      - run:
+          name: Create AWS credentials manually
+          command: |
+            mkdir ~/.aws
+            touch ~/.aws/config
+            chmod 600 ~/.aws/config
+            echo "[profile eb-cli]" > ~/.aws/config
+            echo "aws_access_key_id=$AWS_ACCESS_KEY" >> ~/.aws/config
+            echo "aws_secret_access_key=$AWS_SECRET_KEY" >> ~/.aws/config
+      - run:
+          name: Install NPM packages
+          command: yarn install --frozen-lockfile --prod=true
+      - run:
+          name: Deploying to eb
+          command: |
+            eb use app-staging --region ap-east-1 --profile eb-cli
+            eb deploy app-staging --region ap-east-1 --profile eb-cli
+        
+
 workflows:
   app-dev:
     jobs:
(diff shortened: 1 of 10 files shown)


.platform/hooks/prebuild/yarn.sh
```diff
@@ -0,0 +1,19 @@
+#!/bin/bash
+
+# need to install node first to be able to install yarn (as at prebuild no node is present yet)
+sudo curl --silent --location https://rpm.nodesource.com/setup_14.x | sudo bash -
```
webberwang, line 4: could we remove this, and move to circleci ci. we wouldn't want this to trigger locally

package.json
```diff
@@ -1,7 +1,6 @@
 {
   "name": "codelab",
   "engines": {
-    "node": ">=13",
```
webberwang, line 4: we'd still want this requirement, could we make it work on aws

webberwang (changes requested):

.ebignore
```diff
@@ -0,0 +1,25 @@
+/node_modules
```
webberwang, line 1: I think better to ignore all, then re-add required folders

.circleci/config.yml
```diff
@@ -305,6 +310,35 @@ jobs:
...
+            sudo apt-get install python3-pip awscli build-essential
+            pip3 install awsebcli --upgrade --user
+      - run:
+          name: Create AWS credentials manually
+          command: |
+            mkdir ~/.aws
+            touch ~/.aws/config
+            chmod 600 ~/.aws/config
+            echo "[profile eb-cli]" > ~/.aws/config
+            echo "aws_access_key_id=$AWS_ACCESS_KEY" >> ~/.aws/config
```
webberwang, line 330: we would use vault

.circleci/config.yml
```diff
@@ -83,6 +84,10 @@ executors:
     docker:
       - image: joatmon08/circleci-gcp-agent:287.0-1.14.72
     resource_class: small
+  python:
```
webberwang, line 87: is this being used?

webberwang (changes requested):

.platform/hooks/prebuild/yarn.sh
svitalii33: It is required step to run yarn command in beanstalk

webberwang, line 4: Can this file be removed now that we have correct node version?

.circleci/config.yml
```diff
@@ -306,6 +306,35 @@ jobs:
...
+      - run:
+          name: Create AWS credentials manually
+          command: |
+            mkdir ~/.aws
+            touch ~/.aws/config
+            chmod 600 ~/.aws/config
+            echo "[profile eb-cli]" > ~/.aws/config
+            echo "aws_access_key_id=$AWS_ACCESS_KEY" >> ~/.aws/config
+            echo "aws_secret_access_key=$AWS_SECRET_KEY" >> ~/.aws/config
+      - run:
```
webberwang, line 328: We have a job for this right?

webberwang (changes requested):

.platform/hooks/prebuild/yarn.sh
svitalii33: As I mentioned in the comment, above command is required to install yarn in prebuild stage.

.platform/hooks/prebuild/yarn.sh
```diff
@@ -0,0 +1,15 @@
...
+sudo wget https://dl.yarnpkg.com/rpm/yarn.repo -O /etc/yum.repos.d/yarn.repo
+sudo yum -y install yarn
+
+# install
+cd /var/app/staging/
+
+# debugging..
+ls -lah
+
+yarn install --frozen-lockfile --prod=true
```
webberwang, line 13: is this required?

webberwang (changes requested):

.circleci/config.yml
```diff
@@ -306,6 +306,32 @@ jobs:
...
+          name: Deploying to eb
+          command: |
+            eb use app-staging --region ap-east-1 --profile eb-cli
```
webberwang, line 331: We have `AWS_REGION`

webberwang (changes requested):

PR #713, opened 2021-04-03
feat(ci): add app-prod workflow to deploy

svitalii33:
## Related Issue(s)

Fixes #

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -18,10 +18,16 @@ filters-dev: &filters-dev
       ignore: /.*/
 
 filters-staging: &filters-staging
-# https://discuss.circleci.com/t/circle-2-0-workflow-filters-should-be-logical-and-not-or/18231
   filters:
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -330,6 +336,31 @@ jobs:
...
+    steps:
+      - setup-workspace
+      - run:
+          name: Install dependencies
+          command: |
+            sudo apt-get -y -qq update
+            sudo apt-get install python3-pip awscli build-essential
+            pip3 install awsebcli --upgrade --user
+      - run:
+          name: Create AWS credentials manually
```
webberwang, line 351: These should all be created as re-useable commands

webberwang (changes requested):

.circleci/config.yml
webberwang: same for all others

PR #717, opened 2021-04-08
654 hasura webhook

### LINKED ISSUE #654: Hasura event webhook
Send Hasura events to localhost via ngrok tunnel (credentials sent via chat).

`yarn tunnel`

`modules/lambda-api`

Generate `@nrwl/nest:library`

- [ ] Integration specs for Create

No need to test function, just assert the response from aws

- [ ] Integration specs for Delete

Assert response from aws, and should not exists when finding by lambdaId.

- [ ] Integration specs for Read

- Should be able to get the aws function using our lambdaId. We should be able to execute the function while passing in arguments, and assert the return value of the function

- [ ] Integration specs for Update

- Assert response from aws, then re-test function. Re-use the function from Create & Read test.

artonio:
## Related Issue(s)

Fixes #654


.graphqlconfig
```diff
@@ -4,18 +4,18 @@
   "extensions": {
     "endpoints": {
       "Hasura GraphQL Endpoint": {
-        "url": "${env:HASURA_URL_GRAPHQL}",
+        "url": "https://codelab-ai.hasura.app/v1/graphql",
```
webberwang, line 7: How come these are changed?

apps/api-test/tsconfig.spec.json
```diff
@@ -0,0 +1,9 @@
+{
```
webberwang, line 1: this module should be removed using nrwl

libs/backend/src/infrastructure/persistence/aws/AwsLambdaService.ts
```diff
@@ -10,14 +10,15 @@ import {
...
+import { ILambda } from '../../../../../modules/lambda-api/src/lib/interfaces/IEventTrigger';
 
 export class AwsLambdaService extends LambdaClient {
   bucketPrefix = 'codelab-lambda'
 
-  async createFunction(lambda: Lambda) {
+  async createFunction(lambda: ILambda) {
     const params: CreateFunctionRequest = {
       Code: {
-        S3Bucket: `${this.bucketPrefix}-${lambda.libraryId}`, // BUCKET_NAME
```
webberwang, line 20: keep `libraryId`, you can change name in console 
image

libs/backend/src/infrastructure/persistence/aws/AwsLambdaService.ts
```diff
@@ -10,14 +10,15 @@ import {
   UpdateFunctionCodeRequest,
 } from '@aws-sdk/client-lambda'
 import { Lambda } from '@codelab/frontend/shared'
+import { ILambda } from '../../../../../modules/lambda-api/src/lib/interfaces/IEventTrigger';
```
webberwang, line 13: this would trigger a lint error for nrwl boundaries, use project reference import instead

libs/modules/lambda-api/.babelrc
```diff
@@ -0,0 +1,3 @@
+{
+  "presets": ["@nrwl/web/babel"]
```
webberwang, line 2: is this generated using `@nrwl/nest:library`?

webberwang (changes requested):

libs/modules/lambda-api/.babelrc
artonio: yes, that is what i used

libs/backend/src/infrastructure/persistence/aws/AwsLambdaService.ts
artonio: Screenshot from 2021-04-08 19-23-40

cannot change it, this is what hasura passed to the web hook, i checked in setting, graphql field name is: libraryId, but the table field name is getting passed

.graphqlconfig
artonio: i changed them because i did not understand that schema.graphql is generate using .graphqlconfig.yaml. also .graphqlconfig did not read env variables and it did not work

Issue #751, opened 2021-04-19
Cypress spec for Atom


webberwang:
Let me know if any of these require a different use case

- [ ] CreateAtom
- [ ] UpdateAtom
- [ ] DeleteAtom

Write Cypress specs for these

Add any comments here https://airtable.com/tblVnr2UK7V9ZdKQp/viwsvAJUSigZs5ztD?blocks=hide

![image](https://user-images.githubusercontent.com/26264925/115279025-08372800-a0fb-11eb-9ad1-055ba4f76ee1.png)

PR #782, opened 2021-04-24
test(apps-web-e2e): #751 atom  CRUD Tests

antonmihaylov:
CLOSES: #751

### DIFF
--- a/apps/web-e2e/src/integration/atom.spec.ts
+++ b/apps/web-e2e/src/integration/atom.spec.ts
@@ -1,4 +1,8 @@
-import { Atom_Type_Enum, CreateAtomGql } from '@codelab/hasura'
+import {
+  Atom_Type_Enum,
+  CreateAtomGql,
+  DeleteAllAtomsGql,
+} from '@codelab/hasura'
 import { print } from 'graphql'
 import { sample } from 'lodash'
 
@@ -10,19 +14,13 @@ const pageId = '4ccde878-25d3-4b36-a9b8-67fcec986e7b'
 
 const deleteAllAtoms = () => {
   return cy.hasuraAdminRequest({
-    query: `
-        mutation DeleteAllAtoms {
-          delete_atom(where:{}) {
-            affected_rows
-          }
-        }
-      `,
+    query: print(DeleteAllAtomsGql),
   })
 }
 
 const openAtomsTab = () => {
   cy.visit(`/apps/${appId}/pages/${pageId}`)
-  cy.get('[data-test-id=atom-tab-trigger]').click()
+  cy.getByTestId('atom-tab-trigger').click()
 }
 
 const randomAtomType = () =>
@@ -39,6 +37,12 @@ const insertOneAtom = (atomType: Atom_Type_Enum) => {
   })
 }
 
+const getAtomListItem = (atomType?: string) =>
+  cy.getByTestId(
+    'get-atoms-list-item',
+    atomType ? `[data-test-atom-type=${atomType}]` : '',
+  )
+
 describe('Atom', () => {
   before(() => {
     cy.login()
@@ -50,19 +54,17 @@ describe('Atom', () => {
     openAtomsTab()
 
     //We should have no items in the list
-    cy.getByTestId(`get-atoms-list-item`).should('not.exist')
+    getAtomListItem().should('not.exist')
 
     cy.getByTestId(`create-atom-button`).click()
     cy.getByTestId(`create-atom-form`).find('div[name=type] input').click()
 
     cy.get('div#create-atom-form-0000_list + div div[__typename=atom_type]')
       .first()
-      .then((el) => {
-        cy.log(el.attr('type') as any)
+      .then((typeSelectInput) => {
+        const selectedAtomType = typeSelectInput.attr('type')
 
-        const selectedAtomType = el.attr('type')
-
-        el.trigger('click')
+        typeSelectInput.trigger('click')
 
         cy.get('.create-atom-modal button[type=submit]').click()
 
@@ -73,9 +75,7 @@ describe('Atom', () => {
         cy.get('.create-atom-modal').should('not.exist')
 
         //We should have the new item in the list
-        cy.get(
-          `[data-test-id=get-atoms-list-item][data-test-atom-type=${selectedAtomType}]`,
-        )
+        getAtomListItem(selectedAtomType)
       })
   })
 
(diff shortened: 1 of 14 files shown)


apps/web-e2e/src/integration/atom.spec.ts
```diff
@@ -1,14 +1,158 @@
...
+
+//Those are in the seeds, so we know they exist
+//If we remove them from the seeds, need to add them dynamically here
+
 const appId = 'c83654f1-f89f-48a0-8741-acf91e38c888'
 const pageId = '4ccde878-25d3-4b36-a9b8-67fcec986e7b'
 
+const deleteAllAtoms = () => {
+  return cy.hasuraAdminRequest({
+    query: `
```
webberwang, line 13: make graphql file then import generated doc

apps/web-e2e/src/integration/atom.spec.ts
```diff
@@ -1,14 +1,158 @@
...
+            affected_rows
+          }
+        }
+      `,
+  })
+}
+
+const openAtomsTab = () => {
+  cy.visit(`/apps/${appId}/pages/${pageId}`)
+  cy.get('[data-test-id=atom-tab-trigger]').click()
```
webberwang, line 25: getByTestId

apps/web-e2e/src/integration/atom.spec.ts
```diff
@@ -1,14 +1,158 @@
...
+    cy.get('div#create-atom-form-0000_list + div div[__typename=atom_type]')
+      .first()
+      .then((el) => {
+        cy.log(el.attr('type') as any)
+
+        const selectedAtomType = el.attr('type')
+
+        el.trigger('click')
+
+        cy.get('.create-atom-modal button[type=submit]').click()
```
webberwang, line 67: couldn't we use data-test-id on button?

apps/web-e2e/src/integration/atom.spec.ts
```diff
@@ -1,14 +1,158 @@
...
+
+        cy.wait('@graphql')
+
+        //modal should close
+        cy.getByTestId(`create-atom-form`).should('not.exist')
+        cy.get('.create-atom-modal').should('not.exist')
+
+        //We should have the new item in the list
+        cy.get(
+          `[data-test-id=get-atoms-list-item][data-test-atom-type=${selectedAtomType}]`,
```
webberwang, line 77: for my own knowledge, is this a nested selector? or chained on same element?

apps/web-e2e/src/integration/atom.spec.ts
```diff
@@ -1,14 +1,158 @@
...
+        )
+          .find('[data-test-id=atom-delete-button]')
+          .click()
+
+        cy.get('.delete-atom-modal button[type=submit]').click()
+
+        cy.wait('@graphql')
+
+        cy.getByTestId('delete-atom-form').should('not.exist')
+        cy.get('.delete-atom-modal').should('not.exist')
```
webberwang, line 151: possible to use test id? or is this an antd component that doesn't take it?

apps/web-e2e/src/support/commands.ts
```diff
@@ -67,6 +71,22 @@ Cypress.Commands.add('login', () => {
...
+    url: Cypress.env('CODELAB_HASURA_GRAPHQL_ENDPOINT'),
+    method: 'POST',
+    headers: {
+      'x-hasura-admin-secret': Cypress.env('HASURA_GRAPHQL_ADMIN_SECRET'),
+    },
+  })
+})
+
+Cypress.Commands.add('getByTestId', (testId) => {
+  return cy.get(`[data-test-id=${testId}]`)
```
webberwang, line 87: any reason over `data-cy`?

webberwang (changes requested):

apps/web-e2e/src/support/commands.ts
antonmihaylov: more explicit, if you see it you know what it's for

apps/web-e2e/src/integration/atom.spec.ts
antonmihaylov: modal doesn't take data attributes. Maybe I can just wrap it in a div though

apps/web-e2e/src/integration/atom.spec.ts
antonmihaylov: Same element, since there's no space between the two [] selectors, that's why getByTestId won't work here

apps/web-e2e/src/integration/atom.spec.ts
antonmihaylov: I'll check on that

apps/web-e2e/src/integration/atom.spec.ts
antonmihaylov: Missed that one

apps/web-e2e/src/integration/atom.spec.ts
```diff
@@ -1,14 +1,158 @@
...
+      const atomType = randomAtomType()
+
+      insertOneAtom(atomType).then(() => {
+        openAtomsTab()
+
+        //We should have the new atom in the list, click its update button
+        cy.get(
+          `[data-test-id=get-atoms-list-item][data-test-atom-type=${atomType}]`,
+        )
+          .find('[data-test-id=atom-update-button]')
```
webberwang, line 96: findByTestId

webberwang (changes requested):

apps/web-e2e/src/support/commands.ts
webberwang: i mean i would argue data-cy is more explicit, because cypress is a subset of testing. you could have jest dom testing as well

apps/web-e2e/src/integration/atom.spec.ts
antonmihaylov: won't work here, this is a nested selector

apps/web-e2e/src/support/commands.ts
antonmihaylov: but jest would rely on the same ids, why keep them separate?

webberwang: we might as use `data-testid` for re-useability https://testing-library.com/docs/queries/bytestid/

apps/web-e2e/src/integration/atom.spec.ts
webberwang: yeah since we have a modal component anyways, this way less context switching when selecting elements for cy testing

antonmihaylov: doesn't work, the model is rendered somewhere in a portal, so even if we wrap it it won't be displayed under the wrapping parent

apps/web-e2e/src/integration/atom.spec.ts
antonmihaylov: no, we don't control how the button is rendered, we only pass props to the modal and it doesn't allow data- props

PR #820, opened 2021-05-07
feat(libs-frontend-shared): add SelectField with search

aburtsev:
## Related Issue(s)

Fixes #


webberwang:
@aburtsev https://stackoverflow.com/questions/58787934/how-to-override-external-typescript-interface, ideally we don't have to copy the entire interface

webberwang:
We'll keep the branch, but will use tsignore for now

PR #869, opened 2021-05-29
feat: update ci to deploy ecs staging

svitalii33:
## Related Issue(s)

Fixes #


.circleci/config.yml
```diff
@@ -20,12 +20,16 @@
 filters-staging-pr: &filters-staging-pr
```
webberwang, line 20: this doesn't have to do with staging

Issue #884, opened 2021-06-03
Preliminary prop design, entity modeling


webberwang:

antonmihaylov:
### Types and props module overview
- Atom has Interface, which defines the API for the props it takes
- PageElement has Props, which are related to and validated by the Interface of the atom
- If a PageElement has no Atom, it can't have Props
- Interface must be able to reference other Interfaces, making the reusable
- Interface must have a way to include flexible types, like nested Interfaces, Arrays, Enums, Arrays of Interfaces, etc.
- Can't add fields with the same key to the same interface

## Use cases
- [ ] Interface
   - [ ] Create Interface
   - [ ] Get Interface (by Atom, by Id)
   - [ ] Update Interface
   - [ ] Add Field to Interface (separated from UpdateInterface, in order to reduce complexity)
   - [ ] Validate Interface (takes in arbitrary json data and validates in against the interface)
- [ ] Field
   - [ ] Create Field (maybe not needed, if we just use Add Field To Interface?)
   - [ ] 

- Prop CRUD
- Assign / Deassign Interface to Atom
- Assign / Deassign Props to PageElement
- Validate Props based on Interface

PR #890, opened 2021-06-11
884 prop module

antonmihaylov:
1. type-api module
see #884 for implementation details
took a bit more verbose approach with dedicated mapper services and separate dgraph models, but the result is code that is a lot more readable and reusable, for a large module as type-api turned out to be, think that's more important

2. [graphqlRequest helper](https://github.com/codelab-ai/codelab.ai/commit/99b90059d4048f6c48e54bc8746c02c7d84e1c76#diff-3381931ec322dd717ba43df6cdaa0f90568738be21e23cfd82c9ec274c1ac783R15) for integration tests 

3. Dgraph query builder improvements. See type-api/models/Dgraph*.model and type-api/use-cases/*.query for examples. This might be a bit of a stretch, but I spent quite some time fixing things caused by improper configuration of dql queries and their result's types. And that way they are more reusable, otherwise dql fields are all over the place.


webberwang:
definitely quite a bit of code, I'll take a closer look down the line. 

I see quite a bit of code for dql wrapper, are those for type safety reasons?

antonmihaylov:
> definitely quite a bit of code, I'll take a closer look down the line. 
> 
> I see quite a bit of code for dql wrapper, are those for type safety reasons?

The dql wrapper code hase a couple of purposes: 
1. Type safety, yes
2. Mapping from dql models to api models
3. Validation of dql models (there's a zod schema for each one)
4. Type safe and reusable query building

I went without them at first, but found myself messed up in all the queries and mapping from dql models -> api models all over the place, it was confusing while building, let alone after 1 month when revisiting it

PR #913, opened 2021-07-02
910 styling

### LINKED ISSUE #910: Styling
Styling by manually specifying the CSS interface can be too cumbersome.

Some temporary solutions are:
- [ ] ~~Inline JSON string as a style object, that gets parsed using JSON.parse when rendering~~
image

- [X] Element level css editor, see if we can inject it into emotion
- [X] Page level css editor, using classes

antonmihaylov:
fixes #910

Demo:

https://user-images.githubusercontent.com/57956282/124333046-cebf7600-db9b-11eb-8d19-8283307026dc.mp4

- Added CSS editor tab in the meta tab of page elements
- It can be used to edit element-level CSS by setting it on the target element, or as page-level css, by setting it on the page root element and assigning classes to page elements by a prop (or whatever selector you want)
- The css gets passed down to emotion, so anything allowed in emotion is allowed in there too (except js expressions)
- Used Monaco editor for auto completion and syntax highlighting

Problems:
Monaco editor is great, but by default it can't handle similar css-in-js syntax, where css properties are written at the root level, without a selector. That and it gave a weird css error, because next doesn't like it injecting global css appereantly.
I tried wrapping the css around some kind of dummy selector, but it was very flaky and broke easily. So I forked the plugin responsible for css/scss/less processing, modified it to accept the css-in-js syntax and used that instead.
The plugin usually handles language processing by delegating the work to a web worker. But I couldn't get that to work with the modified version, so I removed the web worker part and everything is running on the main thread now, as a temporary solution.
This might cause some lag when loading autocompletion, if it's a problem we can try to get the web worker to work again.


webberwang:
Should we raise an issue in their repo?

antonmihaylov:
> Should we raise an issue in their repo?

Might be helpful, but I'm not sure where the problem is - in the core lib, or, there's another library - monaco-react, which packages the editor for react.

PR #1023, opened 2021-08-10
0 query fix

antonmihaylov:
Fixes issues with the Query component

Note - after this, query props needs to get pass down as object from the lambda, e.g.: 
```
return {
  data: [..]
}
```
instead of:
`return [..data]`


webberwang:
nice

PR #1045, opened 2021-08-16
Fix menu render issue(#1043)

### LINKED ISSUE #1043: Fix rendering and prop mapper issue
- ### Fix the rendering mechanism in the `renderFactory()` method.

> https://github.com/codelab-ai/codelab.ai/blob/64bda14a643dd6b0ca65d8c33106dd45b4a4cffb/libs/frontend/modules/builder/src/renderer/renderFactory.tsx#L159
> 
> The `AntDesignMenuItem` is blinking when rendering. 
> To fix this, it is needed to convert the JSX to normal JavaScript function. 
> image
> This will save the extra wrapper element.

> The reason of blinking: 
> ... > Menu > Wrapper Element > MenuItem

We need to update the Prop Mapper in the way of passing props down as function arguments

angelwong0963:
## Expected Behavior
The render should work without bilking when rendering the `AntDesignMenuItem`.

## Related Issue(s)

> Closes #1043

### DIFF
--- a/libs/frontend/modules/builder/src/renderer/renderFactory.tsx
+++ b/libs/frontend/modules/builder/src/renderer/renderFactory.tsx
@@ -22,7 +22,7 @@ export type RenderHandler<TNode extends RenderNode = RenderNode> = (
       __node: TNode
     } & Record<string, any>
   >,
-) => ReactElement | null
+) => any
 
 //
 // Helpers:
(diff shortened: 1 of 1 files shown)


libs/frontend/modules/builder/src/renderer/renderFactory.tsx
```diff
@@ -22,7 +22,7 @@ export type RenderHandler<TNode extends RenderNode = RenderNode> = (
       __node: TNode
     } & Record<string, any>
   >,
-) => ReactElement | null
+) => any
```
webberwang, line 25: Is there some other type we can use?

antonmihaylov (commented):
Does the Prop Mapper work after this change? I remember I had to switch from plain functions to JSX in order to make it work. The RenderContext wasn't properly passed down otherwise

angelwong0963:
I am checking the State and Prop Mapper with this update. Please delay the merging.

PR #1069, opened 2021-09-01
feat: wip hooks and graphql-let refactor

antonmihaylov:
- Basic hooks skeleton
- Hooks create/delete UI 
- Query hook with config 
- Query hook with lambda - WIP
- Fixed bulder bugs
- ~Refactored for graphql-let:~
    - Moved all domain enums in codelab/shared/abstract/core
    - Removed codelab/shared/graphql
    - Added separate .graphql files wherever is needed (e.g. CreateApp.api.graphql in both frontend/app and backend/app - for spec)
    - ~Workaround for jest issue with graphql-let~
    - Moved element tree code in codelab/shared
    - Added interfaces for Element Tree code
    - Refactored renderer to use element tree interfaces, instead of fragments

- Removed graphql-let in favor of near-file https://www.graphql-code-generator.com/docs/presets/near-operation-file


webberwang:
hackathon mode

PR #1128, opened 2021-09-27
feat: implement render props, react node

### LINKED ISSUE #1106: Component as props
Let's take a look at some Antd components and the signature for components

List.renderItem = `(item) => ReactNode`
Menu.Item = `ReactNode`

Currently component as props is shown as a list of componentId, then this id is passed to the renderer.

We'll want to make sure that the different signatures work.

We could treat all as `ReactNode`, then map to `() => ReactNode` inside prop pipeline (something like this https://github.com/codelab-ai/codelab.ai/blob/1a8cf7998fa9d4d0ccf3b036bc34601c5cc08f01/libs/frontend/view/components/mapper/Mapper.tsx#L24)

But we'd have to put this meta data somewhere, maybe in the prop form? Let me know your thoughts

----

Let's create 2 menus, one using list 1 using menu to test these out.

phmngocnghia:
- implement render props, react node
- Write e2e tests: for create, render those prop types
- Note: We still only have one type at the moment: `RenderPropsType`. I planed to integrate react node type on another PR which will implement add ReactNode, RenderPropType
- TODO: 
  - Add prop mapper for ReactNode, RenderProps type
  - Idea: render atom as render props, react node?

## Implementations:
- Added 2 types `RenderPropsType` and `ReactNodeType`: types, render pipes
- Seed the props using new types
- Prop schema: `{type: <id of either those types>, id: <id of component>`

## Expected Behavior

https://user-images.githubusercontent.com/36730355/134918160-986718c5-d6a4-428e-924d-b91af22f0373.mov

image

## Related Issue(s)

Fixes #1106


webberwang:
very nice work! will give feedback after checking in detail

PR #1182, opened 2021-11-07
refactor: rtk query element module

antonmihaylov:
#1176


libs/frontend/modules/element/src/use-cases/elementEndpoints.ts
```diff
@@ -0,0 +1,69 @@
+import {
```
assimhabeek, line 1: I used `xxx.endpoints.ts` instead of `xxxEndpoints.ts` we should agree on one naming conversion

PR #1233, opened 2021-12-16
fix: props now are updated

### LINKED ISSUE #1224: hookPipe overriding element props
## Current Behavior

When we add hook to element the props of element are all cleared and also this way props binding doesn't work

assimhabeek:
## Related Issue(s)

Fixes #1224


assimhabeek:
Since we are testing and refactoring pipeline this issue should be fixed in a better way

Issue #1236, opened 2021-12-22
Fix LastRenderProps being empty due to HookWrapper


assimhabeek:
Using a component inside `hookPipe` to be able to use hooks and also re-render when ever hooks are added or deleted cause problems when capturing the render restful and extracting last render props.

The new approach is  to lift this component up called `RenderContainer` which will not be specific for hooks but serves the same goal as the previous `HookWrapper`. then extract the props for each element inside the container and pass it to `BuilderSlice`

assimhabeek:
Fixed

PR #1237, opened 2021-12-22
test: write specs for propsPipeline

assimhabeek:
The following changes are made : 
- Add test config for builder module
- Remove `libs/shared/abstract/core/src/props` and move its types into builder module. 
   `IProps` => `RenderProps`
   `PropsByElementId` => `RenderPropsByElementId`
- Rename `useHookFactory` to `useHookResponse`
- Rename `propsWithTypesPipe` to `typedPropsPipe`
- `typedPropsPipe` : remove for support fetching BaseType by id instead it directly uses `typeKind`. @antonmihaylov  is this accepted or we still need to support fetching by id.  
## Related Issue(s)

Fixes #1237 
         #1236


libs/frontend/modules/builder/src/Builder.tsx
```diff
@@ -62,7 +62,7 @@ const BuilderRenderer = ({
     <Renderer
       isComponentRenderer={isComponentBuilder}
       tree={tree}
-      context={{
+      parentContext={{
```
antonmihaylov, line 65: why the name change, parentContext seems a bit confusing, while context follows the react terminology

libs/frontend/modules/builder/src/renderer/hooks/useHookResponse.ts
```diff
@@ -0,0 +1,88 @@
...
+  [AtomType.HookGraphqlQuery]: useGraphqlQueryHook,
+  [AtomType.HookGraphqlMutation]: useGraphqlMutationHook,
+  [AtomType.HookRecoilState]: useRecoilStateHook,
+  [AtomType.HookQueryPage]: useQueryPageHook,
+  [AtomType.HookQueryPages]: useQueryPagesHook,
+  [AtomType.HookRouter]: useRouterHook,
+}
+
+export const useHookResponse = () => {
+  const getHooksResponse = (
```
antonmihaylov, line 33: What happens if you add/remove a hook from an element that's not the root?

Does it crash with a React error message? Usually when you execute a different number of hooks each render it does that. Before the solution was to put a unique key related to the number of hooks to the HookWrapper. This causes the Wrapper to re-render every time the number of hooks changes, so technically it's a new element and it doesn't matter if it has a d different number of hooks.

 Now that the executing fn is in the Renderer, and the key depends on the hook count of the root element, I'm wondering if it will cause an issue

libs/frontend/modules/builder/src/renderer/pipes/persistedPropsPipe.ts
```diff
@@ -1,17 +1,20 @@
...
 export const persistedPropsPipe: RenderPipeFactory =
   (next) => (element, context, props) => {
-    try {
-      return next(
-        element,
-        context,
-        mergeProps(props, JSON.parse(element.props.data)),
-      )
-    } catch (e) {
+    const persistedProps = attempt(JSON.parse, element.props.data)
```
antonmihaylov, line 10: What's the benefit of using attempt vs try-catch?

libs/frontend/modules/builder/src/renderer/renderContainer/RenderContainer.tsx
```diff
@@ -0,0 +1,22 @@
...
+  appendToKey,
+  element,
+  context,
+  props,
+  isRoot = false,
+}: RenderContainerProps) => {
+  const baseKey = `${element.id}-${element.hooks.length}`
+  const key = appendToKey ? `${baseKey}-${appendToKey}` : baseKey
+
+  return (
```
antonmihaylov, line 14: What's the need for a RenderContainer component, couldn't this functionality be done in the renderFactory, like it was before. It's not much overhead, but it's not readable to have 3 similar components: Renderer, RenderContainer, RenderContainerWithKey

libs/frontend/modules/builder/src/renderer/renderContainer/RenderContainerWithKey.tsx
```diff
@@ -0,0 +1,45 @@
...
+
+    const renderedElement = rendered as ReactElement
+    const elementId = getElementId(renderedElement)
+    const renderedElementProps = renderedElement.props
+
+    if (!elementId || !isObjectLike(renderedElementProps)) {
+      return
+    }
+
+    context.onRendered({ elementId, props: renderedElement.props })
```
antonmihaylov, line 41: How does this work with a tree deeper than 1 element? As I see, context.onRendered is called only on the root rendered element.

Also, this is similar to the way it was done before, we're essentially bringing back the old pattern of calling onRendered for every rendered element. This caused a massive lag spike each re-render since on every call we are updating the redux state again and again. With 10+ elements it gets very noticeable.

That's why I changed it to update the redux state in 1 batch

antonmihaylov (changes requested):
Great thinking on the hooks solution, seems cleaner.
Also good use of WeakMap
Tests seem good enough, we can always add more when we encounter a bug

Just a couple of comments

libs/frontend/modules/builder/src/renderer/hooks/useHookResponse.ts
assimhabeek: I didn't change the way we handle that. however I lift the `HookWrapper` As a container. and the container is used to render each element in the tree https://github.com/codelab-ai/codelab.ai/blob/2521ce80c42b161b1bfd1a2a4717e24b86e733fd/libs/frontend/modules/builder/src/renderer/renderContainer/RenderContainer.tsx#L11
The first reason is that now we have access to the props, we also have the ability now to re-render whenever we want by passing `appendToKey`
I tested that while working on MrHost app and it is working fine

libs/frontend/modules/builder/src/renderer/pipes/persistedPropsPipe.ts
assimhabeek: it looks much cleaner, We have one early return then the actually one. 
https://dev.to/coly010/let-s-clean-up-ugly-try-catches-5d4c

libs/frontend/modules/builder/src/renderer/renderContainer/RenderContainer.tsx
assimhabeek: `RenderContainerWithKey` is an internal component that shouldn't be used by any other component. The difference is that  `RenderContainer` is called for each element and once it is done rendering it will register its props to Redux. The previous approach of resolving props from the rendered tree was blocked by the HookWrapper.

libs/frontend/modules/builder/src/renderer/renderContainer/RenderContainerWithKey.tsx
assimhabeek: it does work with deep trees, but I will see how to update state once

libs/frontend/modules/builder/src/renderer/hooks/useHookResponse.ts
antonmihaylov: https://github.com/codelab-ai/codelab.ai/issues/1043
Here's the issue where we removed the wrapper components. Essentially, RenderElement was a react component which acted similarly to how the RenderContainer wraps every element

libs/frontend/modules/builder/src/renderer/pipes/persistedPropsPipe.ts
antonmihaylov: I disagree, try-catch is almost universally understandable by any developer, while I had to look up what does attempt do and how does it work. 
But again, it seems like a preference, no issue as long as it works

libs/frontend/modules/builder/src/renderer/pipes/renderPipeline.tsx
```diff
@@ -0,0 +1,53 @@
...
+  componentPipe,
+)
+
+// (4). Combine the pipelines and add the final renderChildrenPipe
+export const renderPipeline = compose(
+  propsPipeline,
+  propModifiersPipeline,
+  extraPropsPipe, // here are our 'global' builder props, which must override all other props, since we override onClick here
+  elementPipeline,
+)(renderChildrenPipe)
```
webberwang, line 53: The pipeline should be purely js, no concept of React should be needed. 

We can use use `{renderChildrenPipe()}` inside the tsx view

also use `.ts` extension where possible, such as this file

`renderChildrenPipe`

webberwang (changes requested):

libs/frontend/modules/builder/src/renderer/pipes/renderPipeline.tsx
```diff
@@ -0,0 +1,53 @@
...
+import { renderChildrenPipe } from './renderChildrenPipe'
+import { renderPropsPipe } from './renderPropsPipe'
+import { typedPropsPipe } from './typedPropsPipe'
+
+//
+// Construct the pipeline:
+//
+
+// (1). Base props
+export const propsPipeline = compose(
```
webberwang, line 23: this shouldn't be exported if not used outside this file, same with the others

webberwang (changes requested):

libs/frontend/modules/builder/src/renderer/pipes/renderPipeline.tsx
assimhabeek: we need to export it for tests

webberwang: right nvm

libs/shared/utils/src/props/mergeProps.ts
```diff
@@ -39,7 +39,7 @@ export const mergeTwoPropObjects = (
     }
 
     if (key === '__node') {
-      aggregate[key] = valueA
+      aggregate[key] = valueA || valueB
```
webberwang, line 42: could we use https://github.com/jonschlinkert/merge-deep or some lib to achieve the recursive merging?

libs/frontend/modules/builder/src/renderer/pipes/extraElementPropsPipe.ts
```diff
@@ -1,22 +1,14 @@
...
-      context.extraElementProps[element.id] &&
-      typeof context.extraElementProps[element.id] === 'object'
-    ) {
-      return next(
-        element,
-        context,
-        mergeProps(props, context.extraElementProps[element.id]),
-      )
-    }
+    const propsPath = `extraElementProps.${element.id}`
```
webberwang, line 10: constant would be better here, refactor to `const EXTRA_ELEMENT_PROPS_PREFIX = 'EXTRA_ELEMENT_PROPS_PREFIX_'` and do control search on entire project

perhaps put in `frontend/abstract/core` for now

webberwang (changes requested):
- [ ] Similar to extraElementProps, we do the same for `data-component-id`
- [ ] Also this seems more elegant https://stackoverflow.com/a/44313931/2159920, shared/abstract/types
- [ ] https://github.com/codelab-ai/codelab.ai/blob/c926f8836d07d6d9d94e8862c22fdb8a53003a16/libs/frontend/modules/builder/src/renderer/pipes/propMapBindingsPipe.ts#L14 use PropData interface do ctrl find for `Record<string, any>` and see if they could be substituted
- [ ] Use constant 
image

libs/shared/utils/src/props/mergeProps.ts
assimhabeek: yes, any lib will do the trick. I remember it used to be merge-deep. but since we used Lodash a lot so I chose to use `_.merge` which does the same.

PR #1239, opened 2021-12-24
test: prop modifiers pipeline

### LINKED ISSUE #1238: Unit test propModifiersPipeline

assimhabeek:
## Related Issue(s)

Fixes #1238


libs/frontend/modules/builder/src/renderer/pipes/propMapBindingsPipe.ts
```diff
@@ -7,36 +8,29 @@ import { RenderPipeFactory } from './types'
...
 export const propMapBindingsPipe: RenderPipeFactory =
   (next) => (element, context, props) => {
-    const extraProps = {
-      ...context.extraElementProps,
-    }
+    const { extraElementProps } = context
+    const { propMapBindings } = element
 
-    let currentElementProps: Record<string, any> = {}
+    const toTargetElements = propMapBindings
```
antonmihaylov, line 14: Why do we have to loop twice over the array, wasn't the previous implementation more readable?

antonmihaylov (approved):
Looks good

libs/frontend/modules/builder/src/renderer/pipes/propMapBindingsPipe.ts
assimhabeek: Fixed looping twice issue. 
I really don't believe that nested if/for is more readable. Moreover assigning values inside if/else is considered a bad practice. A generally rule I follow is to avoid if/else for/while and try/catch whenever possible    
Please be aware that is my opinion and if you disagree I will restore the old version.

antonmihaylov: Yeah, It's more of a preference I suppose. 
For/if the version is more performant but it shouldn't be significant here. 
No problem leaving it, as long as it works well

PR #1243, opened 2021-12-30
test: unit test render pipeline utils

### LINKED ISSUE #1242: Unit test render pipeline utils

assimhabeek:
## Related Issue(s)

Fixes #1242


webberwang:
[ ] Refactor Entity interface
- Control search all `id: string` type and replace with `Entity` interface. 
- Put under `libs/shared/abstract/types/src/entity.ts`

[ ] Refactor DgraphEntity interface
- Control search all `uid: string` type and replace with `DgraphEntity` interface
- Put under `libs/shared/abstract/types/src/entity.ts`

webberwang (changes requested):
[ ] Refactor Entity interface
- Control search all `id: string` type and replace with `Entity` interface. 
- Put under `libs/shared/abstract/types/src/entity.ts`

[ ] Refactor DgraphEntity interface
- Control search all `uid: string` type and replace with `DgraphEntity` interface
- Put under `libs/shared/abstract/types/src/entity.ts`

libs/shared/utils/src/props/mergeProps.ts
```diff
@@ -1,73 +1,22 @@
...
+import { mergeWith } from 'lodash'
 
-  if (!propsA) {
-    if (propsB) {
-      return { ...propsB }
-    }
+export const mergeTwoPropObjects = (value: any, srcValue: any, key: string) => {
+  if (key === 'className') {
+    const classList = value ? value.split(' ') : []
+    classList.push(srcValue)
```
antonmihaylov, line 6: What's the need to split the className?

antonmihaylov (changes requested):

libs/frontend/modules/builder/src/renderer/utils/applyBinding.ts
```diff
@@ -1,26 +1,24 @@
...
-    value = get(sourceProps, binding.sourceKey)
-  }
+  const value = isSourceKeyWildcard
+    ? sourceProps
+    : get(sourceProps, binding.sourceKey)
 
-  if (binding.targetKey === '*' && typeof value === 'object') {
-    return { ...targetProps, ...value }
+  if (isTargetKeyWildcard && isObjectLike(value)) {
+    return merge(targetProps, value)
```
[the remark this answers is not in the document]
assimhabeek: well `isObjectLike` is shorter than `typeof value === 'object'` and also it check value isn't null.
I am not sure If we really need it but `merge` is safer since it does a deep merge while spread don't.

libs/shared/utils/src/props/mergeProps.ts
assimhabeek: when writing tests I see ` class1 class2` the random spaces at the beginning and the end  causes problem. we want well spaced className to test easily

libs/frontend/modules/builder/src/renderer/utils/applyBinding.ts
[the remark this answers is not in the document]
antonmihaylov: Spreading null or undefined does nothing. Also, my point was - do we need to merge nested values, or shouldn't we override them?
For example - if we have default props: `{data: [{name: 'asd', someOptionalValue: `asd`}]}`, then we apply a binding that results in `{data: [{name: 'otherName'}]}`, we will still have the `someOptionalValue` in the final result, which is not desirable.

libs/shared/utils/src/props/mergeProps.ts
antonmihaylov: isn't it easier to .trim it?

libs/frontend/modules/builder/src/renderer/utils/applyBinding.ts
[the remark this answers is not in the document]
assimhabeek: Fixed. thanks for the catch

libs/shared/utils/src/props/mergeProps.ts
assimhabeek: Restored, 
I was comparing objects, I had to add the space manually for the test to pass 
```    const merged = {
      prop02: {
        nested: 'propB02',
      },
      propB01: 'propB01',
      propA01: 'propA01',
      className: ' classA classB',
    }

webberwang (changes requested):
Testing

webberwang (approved):

PR #1272, opened 2022-01-13
feat(apps): #1262 add demo form page

### LINKED ISSUE #1262: Create Form inside next.js page
This is for internal use mostly. This creates a form from interface, such as our props form in builder.

Create a nextjs page `apps/web/pages/demo/form`, and use the same form component we use for the props form. This will connect to our interface data

We'll use this as a temporary place before migrating to storybook.

We'll test a list of form fields, below come up with a list of fields we should support.

On submit it should console log the data

- [ ] monoca fields
- [ ] array types
- [ ] object types
- [ ] xxx
- [ ] xxx
- [ ] xxx

phmngocnghia:
CLOSES: #1262

## Expected Behavior

Url: `demo/form`
- Primitive Type (Boolean, Float, Integer, String)
- Array Type (String)
- An object that contains 2 fields, a number and a string
- Enum Type
- Lambda Type
- Element Type
- Element Type
- App Type
- React Node Type
- Render Props Type
- Union Type that contains 2 field type, Float and Integer
- Monaco Type (CSS, Javascript)

https://user-images.githubusercontent.com/34818311/150397770-36a5142c-cf95-465d-886d-0eac660da56c.mp4

## Related Issue(s)

Fixes #1262


webberwang (changes requested):
We only need to create the type. Then pull the type data and populate the custom form with it.

I guess we don't have a way to import just types? 

I added an issue https://github.com/codelab-ai/codelab.ai/issues/1273 to resolve this issue

Issue #1254, opened 2022-01-06
Sync Atoms Remotely


webberwang:
- [ ] Add "SyncAtoms" in admin page & pull from remote data 

In the atoms page, we already have import atoms, this actually imports the associated type as well. So we don't need "SyncTypes" as a separate button in admin.

The import behavior also already upserts both the atoms & types. All the core logic should be there already.

This remote data will be source of truth, we will pull all atoms remotely and do an upsert locally. Any associated types will also be upserted.

## Notes
- This replaces our import/export which uses files.
- You shouldn't have to create new import/export use case, re-use the current one

phmngocnghia:
- the same feature could be used by both external & internal dev
- [ ] Create/Set public data source by tagging
   - a checkbox when creating, and updating an app
   - for internal use, so only admin can have this input
   - This checkbox would tag an app as `public`
- [ ] Let’s implement them for internal first:
  - Have a button on the Admin page that is called: “Sync data from data sources”
  - Click on would call the API
  - API gets the data from the configured endpoint (point to staging) and import the data to the environment of the user

PR #1292, opened 2022-01-19
fix(libs-backend-modules-lambda): #1254 fix lambda specs

### LINKED ISSUE #1274: Fix all lambda specs
I turned off all lambda specs, was having some issues before. Coordinate with me in devops I'll get you setup for aws

phmngocnghia:
CLOSES: #1274

## Current Behavior
- All backend lambda module's specs are disabled

## Expected Behavior
- Enable them
- Fix any issues
  - Throw error in mutation `updateLambda` if lamda is pending
  - Update spec to wait until lambda is active to update

## Related Issue(s)

Fixes #1274

### DIFF
--- a/libs/backend/modules/lambda/src/use-cases/update-lambda/tests/update-lambda.i.spec.ts
+++ b/libs/backend/modules/lambda/src/use-cases/update-lambda/tests/update-lambda.i.spec.ts
@@ -5,6 +5,7 @@ import {
   AwsTokens,
 } from '@codelab/backend/infra'
 import { domainRequest } from '@codelab/backend/shared/testing'
+import { updateLambdaInvalidState } from '../../../application/lambda.error'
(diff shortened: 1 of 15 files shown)


libs/backend/modules/lambda/src/use-cases/update-lambda/tests/update-lambda.i.spec.ts
```diff
@@ -51,18 +52,35 @@ describe.skip('UpdateLambda', () => {
...
+          >(
+            testModule.userApp,
+            TestUpdateLambdaGql,
+            updateLambdaInput,
+            undefined,
+            true,
+          )) as any
+
+          if (res.errors) {
+            expect(res.errors[0].message).toBe(expectUpdateLambdaError.message)
```
webberwang, line 73: Interesting solution, so we recursively call if it's an expected update error until jest global time out

Issue #1323, opened 2022-01-24
Add UI to assign Tag to Atom


webberwang:
I think a checkable tree select makes the most sense

<img width="957" alt="Mouse_Highlight_Overlay_and_TreeSelect_-_Ant_Design" src="https://user-images.githubusercontent.com/26264925/150746759-27a8be8e-e8c0-41dc-855b-f1785626d8fc.png">

In the atom crud form, we'll add a new fields tags that allows this selection

PR #1388, opened 2022-01-29
fix: add ui assign to atom(#1323)

### LINKED ISSUE #1286: Atom Table Filters
Currently, hooks are represented as an `Atom` with a type in the backend, this makes it harder to sort the table.

In the atoms table, all library defaults to `Ant Design` currently. But we have `MUI`, and `Hooks`.

Integrate Tag module to tag those atoms. Essentially an Atom can have Tags.

Tags are a self-recursive structure, so you can create any Tag tree, and use tags that way.

I think currently, we can create multiple Tag trees.

- [ ] Use the UI to create a tag tree below

```
UI Frameworks 
\
| \
|   MUI
Ant Design
```

- [ ] Include tags to import/export for atom https://github.com/codelab-ai/web/issues/1425

- for assigning tags to the atom, we will do this by hand (so no code required)

- [ ] Modify Atom table such that we can filter based on tags

---

You can just test the tagging on 1 or 2 atoms

harryjin:
## Expected Behavior

https://user-images.githubusercontent.com/1405854/151656614-7b98c7ff-87b4-4863-ad6e-09e0a94f48d1.mp4

## Related Issue(s)

Fixes #1286
add tags column to atom table
update crud of atom


libs/backend/modules/atom/src/use-cases/create-atom/create-atom.input.ts
```diff
@@ -14,4 +14,7 @@ export class CreateAtomInput {
     nullable: true,
   })
   declare api?: string
+
+  @Field(() => [String], { defaultValue: [], nullable: true })
```
webberwang, line 18: Use `ObjectRef` type, more self documenting

webberwang (changes requested):
Also all tags should show in the table, not just the parent tag

webberwang:
Closed in favor of #1442

PR #1410, opened 2022-02-02
refactor: atom module to neo4j

antonmihaylov:
#1402 

- Refactors the atom module frontend use cases to use neo4j.
- Does not add Atom.api, because we haven't migrated type module yet. We must add it after that #1409 

Haven't re-implemented import/export atoms, we need to come up with a general import/export strategy first. That's why I haven't deleted the backend atom module too - to use it as a reference for that.

Misc:
- Adds a central way to manage options for the v2 api so that we don't have to change each onSubmit
image

- Changed the schema file generation from on-request to generation from the codegen cli


libs/tools/rtk-query/src/config.ts
```diff
@@ -63,7 +63,9 @@ export interface RTKConfig {
 
 export interface RTKQueryRawPluginConfig
   extends RawClientSideBasePluginConfig,
-    RTKConfig {}
+    RTKConfig {
+  extraOptions?: Record<string, any>
```
webberwang, line 67: Add some doc on why we added this, to pass the env for query

webberwang (approved):

PR #1413, opened 2022-02-03
refactor: page module to neo4j

assimhabeek:
## Related Issue(s)

Fixes #

### DIFF
--- a/.docker/docker-compose.yaml
+++ b/.docker/docker-compose.yaml
@@ -9,7 +9,7 @@ version: '3.8'
 services:
   neo4j:
     container_name: codelab-neo4j
-    image: neo4j:4.0
+    image: neo4j:latest
     volumes:
       - codelab-neo4j:/data
       - codelab-neo4j:/plugins
--- a/libs/frontend/modules/page/src/providers/types.ts
+++ b/libs/frontend/modules/page/src/providers/types.ts
@@ -1,9 +1,9 @@
-import { PageBaseFragment } from '../graphql/PageBase.fragment.graphql.gen'
+import { PageFragment } from '../graphql'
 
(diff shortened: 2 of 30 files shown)


libs/frontend/modules/page/src/providers/types.ts
```diff
@@ -1,9 +1,9 @@
-import { PageBaseFragment } from '../graphql/PageBase.fragment.graphql.gen'
+import { PageFragment } from '../graphql'
 
 export type PageProviderProps = {
   pageId: string
 }
 
 export type UseProvideCurrentPage = {
-  currentPage?: PageBaseFragment
```
webberwang, line 8: No need to use base fragment?

.docker/docker-compose.yaml
```diff
@@ -9,7 +9,7 @@ version: '3.8'
 services:
   neo4j:
     container_name: codelab-neo4j
-    image: neo4j:4.0
+    image: neo4j:latest
```
webberwang, line 12: change to latest but use the version

webberwang (approved):

libs/frontend/modules/page/src/providers/types.ts
assimhabeek: here is the old fragment schema  
```
fragment PageBase on Page {
  id
  name
  rootElementId
  # Elements are expensive to get, don't query it every time
}


fragment PageFull on Page {
  ...PageBase
  elements {
    ...ElementGraph
  }
}
```
I am not sure why the old design contain this duplication (elements and rootElementId) since the rootElement will contain its children 
Now we have only 
```
fragment Page on Page {
  id
  name
  rootElement {
    id
    name
  }
}
```

PR #1479, opened 2022-03-11
1445 import export data

### LINKED ISSUE #1445: Import/export data
Instead of separate use cases for each, we combine import/export so no need for separate use cases.

Import/export
- [ ] Tag
- [ ] Atom
- [ ] Type (including base types)

This will be admin import export, we’ll have a separate one for user data.

Keep in mind order matters, tag will be needed before atom, type needed before atom.

railway17:
Fixes #1445


webberwang (changes requested):
- [ ] We don't need graph diff if we assume id's are same across

Can loop through each interface, then upsert all the descendants 

- [ ] We don't need any validation check, we can assume the input data is valid. With upsert, we don't need to check for existence.

Issue #1511, opened 2022-03-26
Add app/page to import/export


webberwang:
Make sure to update e2e specs as well

PR #1514, opened 2022-03-29
1511 import app in admin

railway17:
Feature #1511


apps/web-e2e/src/integration/admin.spec.ignore.ts
```diff
@@ -14,6 +14,9 @@ describe('Admin Import and Export', () => {
           cy.exportAdminData().then((exportResult: any) => {
```
webberwang, line 14: This file is ignored, not running

apps/web/src/graphql/repositories/app/app.repository.ts
```diff
@@ -0,0 +1,42 @@
...
+  CreateAppsMutationResponse,
+} from '../../ogm-types.gen'
+
+export const appRepository = {
+  exportApp: async (): Promise<Array<AppInput>> => {
+    const apps = (await App()).find()
+
+    return apps
+  },
+  importAppFromJson: async (
```
webberwang, line 15: Having a repo method of `importAppFromJson` is too specified and not re-usable.
 
From admin import resolver we should call appRepository.createApp()

apps/web/src/graphql/resolvers/admin/admin.resolvers.ts
```diff
@@ -57,19 +57,22 @@ export const importAdminData: IFieldResolver<
...
+  const apps = JSON.parse(payload)?.apps ?? ([] as Array<AppCreateInput>)
+  // const tags = JSON.parse(payload)?.tags as TagGraph
+  const atoms = JSON.parse(payload)?.atoms ?? ([] as Array<Atom>)
+  // const createTags = await tagRepository.importTagsFromJson(tags)
   const createAtom = await atomRepository.importAtomFromJson(atoms)
+  const createApps = await appRepository.importAppFromJson(apps, auth0Id)
 
   return Promise.resolve({
-    result: !!createTypes && !!createTags && !!createAtom,
+    result: !!createApps && !!createTypes && /* !!createTags &&*/ !!createAtom,
```
webberwang, line 68: should rename to success if we're using boolean

apps/web/src/graphql/resolvers/admin/admin.resolvers.ts
```diff
@@ -57,19 +57,22 @@ export const importAdminData: IFieldResolver<
...
   })
 }
 
 export const exportAdminData: IRxTxnResolver = () => (txn) => {
   return forkJoin({
-    // tags: tagRepository.getTagsGraph(txn),
+    apps: from(appRepository.exportApp()),
```
webberwang, line 74: should be named `getApps()` and `getAtoms()`, naming it too specified makes it not re-usable

Only exportAdminData is named export

webberwang (changes requested):

webberwang (changes requested):
Where is the pages export?

PR #1524, opened 2022-04-04
feat: tag mobx, add tags to atom table

### LINKED ISSUE #1519: Complete Tag Module
A tag is a recursive model, a tag can have many tags

tag_schema_ts_—_web

A tag is used to tag some data, such as our atoms & types. An atom could be a Ant Design Button (which is tagged Ant Design), or a material UI button (which is tagged Material UI) etc.

---

- [x] Tags table on the main area on `/tags` page
- [x] Tags table edit & delete in the action column
- [x] Tag create button in the header, should be able to select the parent tag from dropdown in the form
- [x] Tag tree on the left pane area (Uses Ant Design tree, similar to our builder which you can access by create an app, then create a page & clicking into the page)

- Clicking on a tag tree item then clicking create in the header will pre-select the parent as that item

---

getTagGraphs is the main function for getting tag data. It fetches an array of tagGraphs, by getting the root tag id & it's descendants. (`tag.endpoints.v2.1.graphql`)

Once we get the descendants, we'll then use the GetTags query to fetch all the tags.

For each Tag (`Tag.fragment.v2.1.graphql`), we have info on whether it's a root, and the array of children id's. This is enough information for us to re-construct the tree in the frontend.

---

I'm currently creating a TreeService that composes of Node classes. I'm thinking our Tag model would inherit from this Node class https://mobx-keystone.js.org/class-models#inheritance.

The Node class should have `id, label, children`, this is the general form of a Node for a Tree.

In MobX a class can only have 1 parent (strong reference), all other references need to be used with useRef (weak reference). In TreeService, we have a root that holds a strong reference to the Node, then each subsequent Nodes will hold strong reference to other Nodes.

The TreeService will also contain a map of Nodes, but those are weak references using Ref, since a Node can only have 1 parent according to MobX

I've discussed this a bit with another team member https://www.loom.com/share/fee6719cf88c49d080746b35f4eb7bca

---

The TreeService should have a method called getAntdTree similar to https://github.com/codelab-ai/web/blob/9d2d756fbe3378f7532bfd543e4395ed520bd2a4/libs/shared/core/src/tree/tree.service.ts#L101

---

An example if inheritance we currently use is the `import { ModalService } from '@codelab/frontend/shared/utils'` that all our entity specific modal extends.

Essentially our Node base class would give the `id, label, children` to the child, since those are all common across a Tree Node. As long as the Node has these properties, the TreeService would use those during traversal.

---

We have 2 different ways of displaying data (table, & tree). The Tree gives us hierarchy info, while the table gives us more room to show data.

---

If the MobX inheritance proves too difficult, you can create a TagTree that contains Tag nodes as well. This could be a good starting point before we refactor to a generalized TreeService that we can use for other tree-like models as well.

scientiststwin:
## Related Issue(s)

Fixes #1519


apps/web/pages/tags/index.tsx
```diff
@@ -41,7 +39,10 @@ const TagPage: CodelabPage<DashboardTemplateProps> = observer(() => {
       <CreateTagModal tagService={store.tagService} />
       <UpdateTagModal tagService={store.tagService} />
       <DeleteTagsModal tagService={store.tagService} />
-      <GetTagsTable tagService={store.tagService} />
+
+      <ContentSection>
```
webberwang, line 43: Move logic inside table & use the `loading` prop

libs/frontend/modules/tag/src/store/tag.service.ts
```diff
@@ -32,14 +32,37 @@ export class TagService extends Model({
     return [...this.tags.values()]
   }
 
+  @computed
+  get TagsListOptions() {
```
webberwang, line 36: Note the case should be `tagsListOptions`

libs/frontend/modules/tag/src/store/tag.service.ts
```diff
@@ -32,14 +32,37 @@ export class TagService extends Model({
...
+    return this.tagsList.map((tag) => ({
+      label: tag.name,
+      value: tag.id,
+    }))
+  }
+
   @modelFlow
   @transaction
   create = _async(function* (this: TagService, input: CreateTagInput) {
+    const parentSetStatement = input?.parentTagId && {
```
webberwang, line 46: Use `const connectParentWhere`

libs/frontend/modules/tag/src/store/tag.service.ts
```diff
@@ -125,4 +155,35 @@ export class TagService extends Model({
       this.tags.set(tag.id, tagModel)
     })
   })
+
+  @modelFlow
+  @transaction
+  getTags = _async(function* (this: TagService) {
+    const { tags } = yield* _await(tagApi.GetTags())
+
+    tags.sort((a, b) => {
```
webberwang, line 164: Why is this needed again?

libs/frontend/modules/tag/src/store/tag.service.ts
```diff
@@ -125,4 +155,35 @@ export class TagService extends Model({
...
+      this.tags.set(tag.id, tagModel)
+    })
+  })
+
+  @modelFlow
+  @transaction
+  getTagDescendants = _async(function* (this: TagService, tagId: string) {
+    const { tagGraphs } = yield* _await(tagApi.GetTagGraphs())
+
+    const TagWithItsDescendants = tagGraphs.find(
```
webberwang, line 183: the convention is to use lower camel case, `tagsWithDescendants` is good note tags is plural

webberwang (changes requested):

libs/frontend/modules/tag/src/store/tag.service.ts
scientiststwin: @webberwang  the variable holds the single tag with its descendants, Should I use plural noun?

webberwang: singular looks correct actually

libs/frontend/modules/tag/src/store/tag.service.ts
scientiststwin: Actually, I don't know. if I remove the sort function get an error.

scientiststwin: @webberwang I found the issue, let's discuss it in slack.

PR #1530, opened 2022-04-06
feat: import-export type and atom

### LINKED ISSUE #1523: Import Export Type
- [ ] Remove `admin.resolver`

## `AdminService.export`

On the frontend we'll fetch `Array<InterfaceType>`, and another query to get the array of descendant id types. Then store query results as json.

## `AdminService.import`

We'll parse the array of types and create them, starting with the types first before creating the interface.

---

Some notes

- I think generating uuid in frontend seems like the way to go
- There is some old code for import/export that would need to be cleaned up, such as getTypeGraph cypher etc, `graph: TypeGraph!` on the type.schema itself

antonmihaylov:
fixes #1523 #1532


apps/web/src/graphql/repositories/admin/exportAdminData.cypher
```diff
@@ -11,7 +11,7 @@ MATCH (this)-[r]-() WHERE labels(this)[0] ENDS WITH 'Type' OR labels(this)[0]='E
     vertices:  [node IN nodes | apoc.map.merge(properties(node), {__resolveType: labels(node)[0], typeKind: labels(node)[0]})],
     edges: apoc.coll.union(
         [rel IN relationships WHERE labels(startNode(rel))[0] = 'InterfaceType' |
-                apoc.map.merge(properties(rel), { source: startNode(rel).id, target: endNode(rel).id, __resolveType: 'InterfaceTypeEdge'})],
+                apoc.map.merge(properties(rel), { source: startNode(rel).id, target: endNode(rel).id, __resolveType: 'Field'})],
```
webberwang, line 14: Actually thinking of changing this graph structure to using descendent id approach, so we don't need to have a complicated cypher query like this

libs/frontend/modules/type/src/store/type-import.service.ts
```diff
@@ -0,0 +1,335 @@
...
+        return foundPrimitive
+      }
+    }
+
+    // if not -get by id
+    return typeService.type(importedType.id)
+  }
+
+  /// Makes the create input without any of the type-type relations
+  private makeCreateInput(
```
webberwang, line 201: I would assume these can be re-used for normal type create? We could have a shared TypeInputFactory?

webberwang (changes requested):

apps/web/src/graphql/repositories/admin/exportAdminData.cypher
antonmihaylov: I didn't use this at all, only regular generated graphql endpoints. I.e. the descendantIds approach

libs/frontend/modules/type/src/store/type-import.service.ts
antonmihaylov: Good idea!

apps/web/src/graphql/repositories/admin/exportAdminData.cypher
webberwang: yeah let's remove it

PR #1545, opened 2022-04-13
feat: cascade delete added for app

### LINKED ISSUE #1535: Authorization & cascade delete
- [x] Fetch only user's data where appropriate

Currently I think we're fetching all apps, we need to fetch only the current user's app.

I think we need to pass the owner id to the graphql query in app.service, then we check for authorization in the backend 

Mouse_Highlight_Overlay

App is just one example, could you go through all the models and double check it's fetching only user id (where appropriate)

---

- [x] Casacade delete

https://neo4j.com/docs/graphql-manual/current/mutations/delete/#_nested_delete

For example, when deleting app, we'll want to delete pages as well, along with elements that are attached to the pages.

This is the only scenario I can think of

scientiststwin:
## Related Issue(s)

Fixes #1535


libs/backend/src/schema/app.schema.ts
```diff
@@ -10,13 +10,16 @@ export const appSchema = gql`
...
+  extend type App
+    @auth(
+      rules: [
+        {
+          operations: [READ, UPDATE, CREATE, DELETE]
+          roles: ["User", "Guest"]
+          where: { owner: { auth0Id: "$jwt.sub" } }
+          bind: { owner: { auth0Id: "$jwt.sub" } }
+        }
+        { operations: [READ, UPDATE, CREATE, DELETE], roles: ["Admin"] }
```
webberwang, line 22: this means admin created apps won't have owner id?

webberwang (changes requested):

apps/web-e2e/src/integration/pages.spec.ts
```diff
@@ -1,23 +1,20 @@
...
   const pageName = 'new useful page'
   const updatedPageName = 'updated page'
 
   before(() => {
     cy.resetDatabase().then(() => {
       cy.login().then(() => {
-        cy.createApp().then((apps) => {
-          appId = apps[0].id
-        })
+        createAppWithCypress(appName)
```
webberwang, line 12: Here we can just use a graphql query to create the app (no need to test the UI again)

webberwang (changes requested):

PR #1552, opened 2022-04-23
1535 authorization cascade delete nghia

### LINKED ISSUE #1535: Authorization & cascade delete

phmngocnghia:
## Related Issue(s)

Fixes #1535


libs/frontend/modules/tag/src/use-cases/create-tag/CreateTagModal.tsx
```diff
@@ -8,7 +9,16 @@ import { WithTagService } from '../../store/tag.service'
...
 export const CreateTagModal = observer<WithTagService>(({ tagService }) => {
-  const onSubmit = (input: ICreateTagDTO) => tagService.create({ ...input })
+  const { user } = useUser()
+
+  const onSubmit = (input: ICreateTagDTO) => {
+    if (!user?.sub) {
+      throw new Error('Missing user sub')
+    }
+
+    return tagService.create({ ...input }, user?.sub)
```
webberwang, line 19: Should move the ownerId to ICreateTagDTO

webberwang (changes requested):

apps/web/pages/apps/[appId]/pages/[pageId]/builder.tsx
```diff
@@ -66,7 +69,12 @@ const PageBuilder: CodelabPage<any> = observer(() => {
       await builderService.builderRenderer.init(
         elementService.elementTree,
         providerElementService.elementTree,
-        createMobxState(storeTree, apps, pages, router),
+        createMobxState(
+          storeTree as Nullish<Store>,
```
Comment (webberwang) [inline apps/web/pages/apps/[appId]/pages/[pageId]/builder.tsx:73]: Import the interfaces instead

apps/web/pages/apps/[appId]/pages/[pageId]/index.tsx
```diff
@@ -61,7 +64,12 @@ const PageRenderer: CodelabPage<any> = observer(() => {
       await renderService.init(
         elementService.elementTree,
         providerElementService.elementTree,
-        createMobxState(storeTree, apps, pages, router),
+        createMobxState(
+          storeTree as Nullish<Store>,
+          apps as Array<App>,
+          pages as Array<Page>,
```
Comment (webberwang) [inline apps/web/pages/apps/[appId]/pages/[pageId]/index.tsx:70]: There shouldn't be any casting, you can change the argument type of createMobxState if needed

apps/web/pages/tags/index.tsx
```diff
@@ -73,3 +78,16 @@ TagPage.Layout = observer((page) => {
     </DashboardTemplate>
   )
 })
+
+export const getServerSideProps = withPageAuthRequired({
+  getServerSideProps: async (context: GetServerSidePropsContext) => {
+    // await setClientAuthHeaders(context)
```
webberwang, line 84: can remove getServerSideProps here

libs/backend/src/schema/app.schema.ts
```diff
@@ -9,23 +9,21 @@ export const appSchema = gql`
...
-        #        }
-        #        {
-        #          operations: [READ, UPDATE, CREATE, DELETE]
-        #          roles: ["Admin"]
-        #          where: { owner: { auth0Id: "$jwt.sub" } }
-        #          bind: { owner: { auth0Id: "$jwt.sub" } }
-        #        }
+        { operations: [READ], roles: [] }
+        {
+          operations: [READ, UPDATE, CREATE, DELETE]
```
webberwang, line 17: Why is READ required here?

libs/frontend/modules/tag/src/use-cases/create-tag/CreateTagModal.tsx
```diff
@@ -7,46 +12,56 @@ import React from 'react'
...
-    //   label: tag.name,
-    //   value: tag.id,
-    // }))
-    const options = tagService.tagsSelectOptions
-    const defaultOption = tagService.selectedOption
-    const closeModal = () => tagService.createModal.close()
+export const CreateTagModal = observer<
+  WithServices<TAG_SERVICE | USER_SERVICE>
+>(({ tagService, userService }) => {
+  const { user } = useUser()
```
webberwang, line 18: We can remove this & use userService instead

webberwang (changes requested):

Issue #1553, opened 2022-04-25
Improve props UI


antonmihaylov:
Props UI research in similar platforms: https://www.craft.do/s/HNiVGURe7sjZgs

- Refactor interface form
- Remove uniforms from it (keep it in all other non-generated forms)
- Handle 4 types of 'simple' fields:
  - Text (CodeMirror) - for string, int, float
  - Select - for enums and reference types (App, RenderProps, etc.)
  - Checkbox (or toggle) for booleans
  - List field - a combination of the above and a '+' button - for arrays
- Each field has a 'JS' button next to it that transforms it into a CodeMirror field (except the Text CodeMirror field of course)
- The CodeMirror field auto-suggests you values from the state with the syntax of `{{ someState }}`
- You can also enter js expressions inside `{{ }}`

PR #1579, opened 2022-05-12
feat: props form fields

### LINKED ISSUE #1580: Allow Component to be assigned to props
For prop types of ReactNodeType, we should be able to assign a Component to it

image

---

For the PropForm, we will use a select dropdown (not code mirror field), and list all components that are available. We'll store the component id as the prop value.

Label will use component name, and value will use the id

assimhabeek:
- A lot of bug fixes and other improvements

## Related Issue(s)

Fixes #1580
#1553


webberwang:
@assimhabeek looking good

PR #1625, opened 2022-06-12
refactor: action model

assimhabeek:
## Related Issue(s)

Fixes #


libs/frontend/modules/store/src/store/models/resource-action.model.ts
```diff
@@ -0,0 +1,147 @@
...
+    errorAction: prop<Ref<IAnyAction>>(),
+  })
+  implements IResourceAction
+{
+  static hydrate = hydrate
+
+  @modelAction
+  restFetch(client: AxiosInstance, config: IRestActionConfig) {
+    return client.request({
+      method: config.method as Method,
```
antonmihaylov, line 61: Should we add headers in there too? I'd imagine it would be a common thing to need. But likely a task for another PR

libs/frontend/modules/store/src/store/models/resource-action.model.ts
```diff
@@ -0,0 +1,147 @@
...
+      // eslint-disable-next-line @typescript-eslint/ban-types
+      let errorQueue: Array<Function> = []
+
+      if (this.errorAction.current) {
+        errorQueue = yield* _await(this.successAction.current.getQueue())
+      }
+
+      return [
+        // eslint-disable-next-line no-new-func
+        new Function(`this.${this.name}.error=${JSON.stringify(error)}`),
```
antonmihaylov, line 100: This will throw too if somehow the Function code is invalid, is this something that could happen?

libs/frontend/modules/store/src/store/models/resource-action.model.ts
```diff
@@ -0,0 +1,147 @@
...
+    } catch (error) {
+      // eslint-disable-next-line @typescript-eslint/ban-types
+      let errorQueue: Array<Function> = []
+
+      if (this.errorAction.current) {
+        errorQueue = yield* _await(this.successAction.current.getQueue())
+      }
+
+      return [
+        // eslint-disable-next-line no-new-func
```
antonmihaylov, line 133: Same as above

antonmihaylov (changes requested):
Looks great! The actions are well thought out and flexible. The ActionTypedValueTransformer is a neat idea too. 

Left some minor comments on error handling - tell me what you think

libs/frontend/modules/store/src/store/models/resource-action.model.ts
antonmihaylov: I suppose only if this.name is invalid, e.g. imagine `this.some invalid identifier.error =...`

libs/frontend/modules/store/src/store/models/resource-action.model.ts
assimhabeek: After looking deeper, I remember we don't support headers at action level we only allow them at resource level, guess leaving it for another PR would be better.

antonmihaylov: Ah makes sense, leave it out

libs/frontend/modules/store/src/store/models/resource-action.model.ts
assimhabeek: `this` is a reference to our global state object. while name is the name of the action. we are sure that they both exist because we create them at runtime.

assimhabeek: [here](https://github.com/codelab-ai/web/blob/5ebd088df7ea267c2d96d5220a25bb9373c001ad/libs/frontend/modules/store/src/store/models/store.model.ts#L83) we set the action name into the global state. after running an action we add response/error to this action object

libs/frontend/modules/store/src/store/models/custom-action.model.ts
```diff
@@ -0,0 +1,36 @@
...
+  extends ExtendedModel(createActionBase(IActionKind.CustomAction), {
+    code: prop<string>(),
+  })
+  implements ICustomAction
+{
+  static hydrate = hydrate
+
+  getQueue() {
+    // eslint-disable-next-line  no-new-func
+    return Promise.resolve([new Function(this.code)])
```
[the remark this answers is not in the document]
assimhabeek: I agree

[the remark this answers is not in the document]
assimhabeek: I will create a helper function that handles all cases above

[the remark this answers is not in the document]
antonmihaylov: Good idea. The queue processor should handle empty arrays right?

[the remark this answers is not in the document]
assimhabeek: sure

PR #1627, opened 2022-06-21
Component circular reference (Backend solution) DONT MERGE

### LINKED ISSUE #1624: MobX Cycle detected in computation Element
## Current Behavior

When selecting the parent component of an element as its component when creating it or from the right panel. `Uncaught Error: [mobx] Cycle detected in computation` error happens

## Expected Behavior

There should be no error.

## More details
So basically, what is happening is that when a component is a child of itself, either direct child, or a child of another component, which is in turn a child of the first component. In other words we're having a cycle in the component tree. Due to this cycle the renderer goes into an infinite loop. In other words, the rendering starts with `renderRoot()`  in `renderer.model.ts` which calls `renderElement()` that creates an `ElementWrapper`. `ElementWrapper` finds the children of the element and calls `renderChildren`, which calls `renderElement()` for each child and from there the recursion continues.

### Two potential solutions exist
1. Modify the renderer so that a component is not rendered more than once if a cycle exists. To do this some metadata can be kept in the rendering pipe or a property can be added to components, but we need to find a way to reset these values after the rendering has finished.
2. Eliminate the components from the drop down in the right panel. This should involve removing every component that has the parent of the currently selected element as a child. Basically for each option we have to check if there is a cycle.

## Steps to Reproduce

Follow the videos:
- first way

https://user-images.githubusercontent.com/51242349/172781487-eb976311-a40b-4329-a9f3-40a666f4aa2e.mp4

- second way

https://user-images.githubusercontent.com/51242349/172781531-1c55ed6b-240e-4f2a-9d1f-0b75f07adc07.mp4

webberwang:
## Related Issue(s)

Fixes #1624

### DIFF
--- a/libs/frontend/modules/element/src/store/element.model.ts
+++ b/libs/frontend/modules/element/src/store/element.model.ts
@@ -111,6 +111,8 @@ export class Element
   })
   implements IElement
 {
+  calculatingDescendants = false
+
   @computed
   get childrenSorted(): Array<IElement> {
     return [...this.children.values()].map((x) => x.current).sort(compareOrder)
@@ -149,11 +151,17 @@ export class Element
   @computed
   get descendants(): Array<IElement> {
     const descendants: Array<IElement> = []
+    this.calculatingDescendants = true
 
     for (const child of this.childrenSorted) {
       descendants.push(child)
-      descendants.push(...child.descendants)
+
+      if (!child.calculatingDescendants) {
+        descendants.push(...child.descendants)
+      }
     }
+
+    this.calculatingDescendants = false
 
     return descendants
   }
--- a/libs/frontend/modules/renderer/src/renderPipes/componentRenderPipe.ts
+++ b/libs/frontend/modules/renderer/src/renderPipes/componentRenderPipe.ts
@@ -32,7 +32,7 @@ export class ComponentRenderPipe
     const elementService = getElementService(this)
     const rootElement = elementService.elements.get(component.rootElementId)
 
-    if (!rootElement) {
+    if (!rootElement || rootElement.id === element.id) {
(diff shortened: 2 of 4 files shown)


libs/frontend/modules/element/src/store/element.model.ts
```diff
@@ -149,12 +151,18 @@ export class Element
...
     const descendants: Array<IElement> = []
+    this.calculatingDescendants = true
 
     for (const child of this.childrenSorted) {
       descendants.push(child)
-      descendants.push(...child.descendants)
+
+      if (!child.calculatingDescendants) {
+        descendants.push(...child.descendants)
+      }
```
mohasarc, line 161: This goes into an infinite recursion when creating an element that refers its parent component even before rendering. That's why I added the check here.

libs/frontend/modules/renderer/src/renderPipes/componentRenderPipe.ts
```diff
@@ -32,7 +32,7 @@ export class ComponentRenderPipe
     const elementService = getElementService(this)
     const rootElement = elementService.elements.get(component.rootElementId)
 
-    if (!rootElement) {
+    if (!rootElement || rootElement.id === element.id) {
```
mohasarc, line 35: When the root of a component is an instance of the same component it keeps starting the pipe over and over, so I added this check to stop that.

mohasarc, line 35: When the root of a component is an instance of the same component it keeps starting the pipe over and over, so I added this check to stop that.

webberwang:
We'll use the UI solution for now https://github.com/codelab-ai/web/pull/1629

PR #1628, opened 2022-06-23
Custom domains for Next.js pages

### LINKED ISSUE #1617: Enable custom domains for apps

---

Branch: `1617-custom-domains`

---

- [x] Enable custom domains

---

---

phmngocnghia:
## Description (Optional)
- Implement next middle to redirect domain assigned in vercel to app at `_sites/user/[user]/[app][page]`
- CRUD domain entity

https://user-images.githubusercontent.com/36730355/180426666-4ec2d4e1-b38d-4409-8495-8e0c83a4c3ea.mp4
https://user-images.githubusercontent.com/36730355/180426659-a702580c-a150-4bc2-870b-ee53ebba2de0.mp4

## Related Issue(s)

Fixes #1617


libs/backend/src/schema/schema.ts
```diff
@@ -13,6 +14,18 @@ const escapeDotPathKeys = (key: string) => {
   return key.replace(/\./g, '\\.')
 }
 
+const slugCallback: Neo4jGraphQLCallback = async (parent: any) => {
+  return `${parent.name.toLowerCase()}`
```
antonmihaylov, line 18: Use something like https://www.npmjs.com/package/slugify toLowerCase doesn't produce valid url slugs

apps/web/pages/_sites/user/[user]/[app]/[page]/index.tsx
```diff
@@ -0,0 +1,178 @@
...
+    paths,
+    // fallback true allows sites to be generated using ISR
+    fallback: true,
+  }
+}
+
+export const getStaticProps: GetStaticProps = async (context) => {
+  console.log('getStaticProps', context)
+
+  return {
```
Comment (antonmihaylov) [inline apps/web/pages/_sites/user/[user]/[app]/[page]/index.tsx:161]: This doesn't seem functional, but I suppose it will be in another PR, right?

schema.api.graphql
```diff
@@ -695,6 +697,7 @@ type AppElementRootElementNodeAggregateSelection {
 }
 
 input AppOnCreateInput {
+  domains: [String!]!
```
webberwang, line 700: Wouldn't an app only have 1 domain?

apps/web/pages/_sites/user/[user]/[app]/[page]/index.tsx
```diff
@@ -0,0 +1,177 @@
...
+
+  const [, { isLoading, error, data, isDone }] = useStatefulExecutor(
+    async () => {
+      const [app] = await appService.getAll({
+        slug: appSlug as string,
+      })
+
+      console.log('app', app)
+
+      const [page] = await pageService.getAll({
```
Comment (antonmihaylov) [inline apps/web/pages/_sites/user/[user]/[app]/[page]/index.tsx:49]: Can we avoid fetching anything on the client side at all? I imagine we can move all of this in getStaticProps and pass it down as props to the page

libs/frontend/modules/app/src/store/app.service.ts
```diff
@@ -96,7 +97,7 @@ export class AppService
 
     const appModel = App.hydrate(updatedApp)
 
-    this.apps.set(app.id, appModel)
+    this.apps.set(app.id, appModel as any)
```
webberwang, line 100: how come cast is needed here?

libs/frontend/modules/app/src/use-cases/get-apps/ItemDropdown.tsx
```diff
@@ -27,8 +29,12 @@ const menuItemIconStyle: CSSProperties = {
 }
 
 export const ItemDropdown = observer<ItemMenuProps>(({ app, appService }) => {
-  const onEditClick = () => appService.updateModal.open(appRef(app.id))
-  const onDeleteClick = () => appService.deleteModal.open(appRef(app.id))
+  const onEditClick = () => appService.updateModal.open(appRef(app.id) as any)
```
webberwang, line 32: Also here

libs/frontend/modules/domain/src/use-cases/create-domain/CreateDomainModal.tsx
```diff
@@ -0,0 +1,56 @@
...
+}>(({ domainService, userService }) => {
+  const currentAppId = useCurrentAppId()
+
+  const model = {
+    auth0Id: userService.user?.auth0Id,
+    appId: currentAppId,
+  }
+
+  const onSubmit = (data: ICreateDomainDTO) => {
+    return domainService.create(data as any)
```
webberwang, line 27: also here

libs/frontend/modules/domain/src/use-cases/get-domain/GetDomainsItem/GetDomainsItem.tsx
```diff
@@ -0,0 +1,57 @@
...
+}
+
+export const GetDomainItem = observer<GetAppsItemProps>(
+  ({ domain, domainService }) => {
+    const {
+      domainConfig: { misconfigured },
+      projectDomainData: { verified },
+    } = domain
+
+    const url = `https://${domain.name.replace('https://', '')}`
```
webberwang, line 24: I think better to convert to `new URL()` object then access the parts you need

webberwang (changes requested):
I took a quick look some minor things

libs/frontend/modules/domain/src/use-cases/get-domain/GetDomainsItem/GetDomainsItem.tsx
phmngocnghia: @webberwang the domain name is raw. I'm going to validate them using https://www.npmjs.com/package/is-valid-domain
I think we should expect the domain is loaded with `https` protocol

apps/cli/src/repository/app.repo.ts
```diff
@@ -99,6 +99,7 @@ export const createApp = async (app: IAppExport, selectedUser: string) => {
         id: app.id,
         name: app.name,
         owner: { connect: { where: { node: { id: selectedUser } } } },
+        slug: app.name,
```
antonmihaylov, line 102: Is this right, shouldn't it be app.slug?

apps/web/pages/_middleware.ts
```diff
@@ -0,0 +1,75 @@
...
+  const publicRootDomains =
+    process.env.NEXT_PUBLIC_ROOT_DOMAINS?.split(',') || []
+
+  const matchedPublicDomains = publicRootDomains.find((domain) =>
+    hostname?.includes(domain),
+  )
+
+  // vercel domain is for previewing, dev only
+  const vercelURL = String(process.env.VERCEL_URL)
+  const matchedVercelDomain = hostname?.includes(vercelURL)
```
antonmihaylov, line 17: String(undefined) produces 'undefined', so this will actually match if the hostname for some reason includes the word undefined

apps/web/pages/_middleware.ts
```diff
@@ -0,0 +1,75 @@
...
+  const isRootHostName = Boolean(matchedPublicDomains)
```
antonmihaylov, line 18: Is there a reason to use .find and Boolean instead of .some()?

antonmihaylov (changes requested):
Again, massive work, just small comments

libs/backend/src/schema/schema.ts
phmngocnghia: I removed this. The slugs can be inputted on the modal when creating/updating the domain name. The input will be slugified using `slugify`

PR #1629, opened 2022-07-01
Component circular referance (UI solution)

### LINKED ISSUE #1624: MobX Cycle detected in computation Element

mohasarc:
CLOSES: #1624

## Related Issue(s)
Fixes #1624

### DIFF
--- a/libs/backend/src/repositories/component/getDescendantComponentIds.cypher
+++ b/libs/backend/src/repositories/component/getDescendantComponentIds.cypher
@@ -7,6 +7,6 @@ CALL apoc.path.subgraphAll(
 
 UNWIND descendants AS descendant
   WITH descendant
-    WHERE NOT 'Element' IN LABELS(descendant)
+    WHERE 'Component' IN LABELS(descendant)
 
 RETURN collect(DISTINCT descendant.id)
--- a/libs/backend/src/repositories/component/getDescendantComponentIds.cypher
+++ b/libs/backend/src/repositories/component/getDescendantComponentIds.cypher
@@ -1,0 +1,12 @@
+CALL apoc.path.subgraphAll(
+  this,
+  {
+    relationshipFilter: 'COMPONENT_ROOT>|PARENT_OF_ELEMENT>|INSTANCE_OF_COMPONENT>'
+  }
+) YIELD nodes AS descendants
+
+UNWIND descendants AS descendant
+  WITH descendant
+    WHERE NOT 'Element' IN LABELS(descendant)
+
+RETURN collect(DISTINCT descendant.id)
--- a/libs/frontend/modules/type/src/interface-form/fields/select-component/SelectComponent.tsx
+++ b/libs/frontend/modules/type/src/interface-form/fields/select-component/SelectComponent.tsx
@@ -4,9 +4,15 @@ import { HTMLFieldProps } from 'uniforms'
 import { SelectField, SelectFieldProps } from 'uniforms-antd'
 import { interfaceFormApi } from '../../../store'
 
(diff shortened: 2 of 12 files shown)


libs/frontend/modules/type/src/interface-form/fields/select-component/SelectComponent.tsx
```diff
@@ -15,8 +21,14 @@ export const SelectComponent = ({ name, error }: SelectComponentProps) => {
     interfaceFormApi.InterfaceForm_GetComponents(),
   )
 
+  // remove the components that refer the current component to avoid creating circular references
+  const filteredComponents = data?.components?.filter(
+    (component) =>
```
antonmihaylov, line 26: I'm not sure, but do we need to check if `component.id !== activeComponentId` too? I think we do since subgraphAll doesn't return the id of the component itself

antonmihaylov (changes requested):
Overall looking good, just need to check for component.id in SelectComponent and some other minor comments

libs/backend/src/repositories/component/getDescendantComponentIds.cypher
```diff
@@ -0,0 +1,12 @@
+CALL apoc.path.subgraphAll(
+  this,
+  {
+    relationshipFilter: 'COMPONENT_ROOT>|PARENT_OF_ELEMENT>|INSTANCE_OF_COMPONENT>'
+  }
+) YIELD nodes AS descendants
+
+UNWIND descendants AS descendant
+  WITH descendant
+    WHERE NOT 'Element' IN LABELS(descendant)
```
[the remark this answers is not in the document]
mohasarc: Yeah, that makes sense.

libs/frontend/modules/type/src/interface-form/fields/select-component/SelectComponent.tsx
mohasarc: Actually, subgraphAll does return the current component id as well.

antonmihaylov: Oh good, in the examples in the docs they don't have it, but maybe they've omitted it for illustration

antonmihaylov (approved):

Issue #1630, opened 2022-07-03
Builder issue list


mohasarc:
## A list of issues with the builder.
**These issues are broken down into 3 categories:**
`[CRASH/ERROR]` causing a crash or an error.
`[WRONG BEHAVIOR]` not causing an error but a feature not working or producing a wrong result.
`[ANNOYING]` just annoying or inconvenient.

They are also ordered in terms of what I think could be more important to least important.

-----
### 1. `[WRONG BEHAVIOR]` Changing the parent element of any element doesn’t work.

- [x] @webberwang #1631  

-----
### 2. `[WRONG BEHAVIOR]` Changing the order of any element doesn’t work either by drag-drop or by changing the value of the `order` input.

- [x] @webberwang #1631  

-----
### 3. `[WRONG BEHAVIOR]` Removing the value of INSTANCE_OF_COMPONENT of an element can leave residue sometimes.

The element becomes the parent of the children of the removed component. This is fixed by refreshing the page. Similar to this issue, when a component is deleted, any element that is an instance of it becomes the parent of the component's children temporarily (until the page is refreshed).

Steps to reproduce:
- Make an element an instance of some component.
- Refresh the page
- Revert the element to its previous state i.e. not an instance of the component.
- The elements under the component will become the children of this element temporarily

https://user-images.githubusercontent.com/51242349/177056835-c4f100b7-a4d7-4f20-8ca3-f70c29599a66.mp4

- [x] @mohasarc #1634 

-----
### 4. `[ERROR/CRASH]` Changing the component of an element (INSTANCE_OF_COMPONENT) doesn’t work directly.

Need to remove the current component, then select a new one. `The Error: Error while updating element Element.instanceOfComponent must be less than or equal to one`.

- [x] @mohasarc #1633

-----
### 5. `[WRONG BEHAVIOR]` `[CRASH/ERROR]` When converting an element to a component, the new component doesn't show up complete and sometimes throws an error when clicked on, Fixed after refresh.

https://user-images.githubusercontent.com/51242349/177056900-63b00a2f-c8ed-42d1-bb5a-827e2cd18644.mp4

- [x] @mohasarc #1637

-----
### 6. `[ANNOYING]` Selecting an element for the first time causes the component tree to collapse.

https://user-images.githubusercontent.com/51242349/177056934-b91977da-7e14-408d-b2e2-61812e5ac550.mp4

- [ ] @mohasarc #1639

-----
#### c. The duplicate element has the same order value as the original.

[...]

PR #1644, opened 2022-07-12
fix Edit Component option redirects to 404

mohasarc:
## Description (Optional)
fixes: [CRASH/ERROR] (Right click -> Edit Component) on an element that is an instance of component redirects to 404 page.

## Related Issue(s)
Related: #1630


antonmihaylov:
✅ will merge after the other ones

Issue #1645, opened 2022-07-12
UI for common `CSS` cases


mohasarc:
# Description
A number of the most common `CSS` usecases are listed bellow. The goal is to make a simple GUI to modify the css of these common cases.

# Common `CSS` Usecases
### Display

| Option   | Type      | Values
|:----------|:-------------|:---------------
| display | `select-one` | `block` `flex` |
| flex-direction | `select-one` | `row` `row-reverse` `column` `column-reverse` |
| align-items | `select-one` | `flex-start` `center` `flex-end` `stretch` `baseline`   |   
| justify-content | `select-one`  |`flex-start` `center` `flex-end` `space-between` `space-around` | `space-evenly`
| flex | `number` |
| flex-wrap | `select-one` | `wrap` `nowrap` `wrap-reverse`

### Spacing
<table>
<tr><th>Margins</th><th>Padding</th></tr>
<tr><td>

| Option   |      Type | Units
|:----------|:-------------|:---------------
| margin-top | `number` | `px` `%` `em` `rem` `ch` `vw` `vh`
| margin-bottom | `number` | `px` `%` `em` `rem` `ch` `vw` `vh`
| margin-left | `number` | `px` `%` `em` `rem` `ch` `vw` `vh`
| margin-right | `number` | `px` `%` `em` `rem` `ch` `vw` `vh`

</td><td>

| Option   |      Type | Units
|:----------|:-------------|:-----------
| padding-top | `number` | `px` `%` `em` `rem` `ch` `vw` `vh`
| padding-bottom | `number` | `px` `%` `em` `rem` `ch` `vw` `vh`
| padding-left | `number` | `px` `%` `em` `rem` `ch` `vw` `vh`
| padding-right | `number` | `px` `%` `em` `rem` `ch` `vw` `vh`

</td></tr> </table>

### Size
| Option   |      Type | Units
|:----------|:-------------|:----------|
| height | `number` `auto` | `px` `%` `em` `rem` `ch` `vw` `vh`
| width | `number` `auto` | `px` `%` `em` `rem` `ch` `vw` `vh`
| min-height | `number` `auto` | `px` `%` `em` `rem` `ch` `vw` `vh`
| min-width | `number` `auto` | `px` `%` `em` `rem` `ch` `vw` `vh`
| max-height | `number` `auto` | `px` `%` `em` `rem` `ch` `vw` `vh`
| max-width | `number` `auto` | `px` `%` `em` `rem` `ch` `vw` `vh`

[...]

PR #1651, opened 2022-07-17
CSS Layout GUI

mohasarc:
## Description (Optional)
This PR adds support for GUI css alongside custom css. Additionally it implements support for layout related Css as listed in #1645

## Demo

https://user-images.githubusercontent.com/51242349/180513942-e0b49ed7-b3e2-46f5-baba-eb7e74c4d179.mp4

## Related Issue(s)
Related: #1645


libs/backend/src/schema/element.schema.ts
```diff
@@ -44,7 +44,8 @@ export const elementSchema = gql`
       )
     # fixedId: String!, // This could be no longer needed with neo4j
     name: String
-    css: String
+    customCss: String
```
antonmihaylov, line 47: Descriptive names, nice. But could you still add a 1-line comment desribing which is what

libs/frontend/modules/element/src/css-editor/ElementCssEditor.tsx
```diff
@@ -41,36 +56,61 @@ export const ElementCssEditor = observer(
...
       return () => {
-        updateCss(cssStringRef.current).then()
+        updateCustomCss(customCssStringRef.current).then()
       }
-    }, [updateCss])
+    }, [updateCustomCss])
 
     /*
      * Debounce autosave, otherwise it will be too quick
      * Getting a dgraph  error if this is too fast, like 500ms
```
antonmihaylov, line 65: This comment is way outdated, we don't use dgraph anymore 😆

libs/frontend/modules/element/src/css-editor/ElementCssEditor.tsx
```diff
@@ -41,36 +56,61 @@ export const ElementCssEditor = observer(
...
-        onChange={(value) => setCssString(value)}
-        value={cssString}
-      />
+      <>
+        <EmotionCssEditor
+          height="100%"
+          onChange={(value) => setCustomCssString(value)}
+          value={customCssString}
+        />
+        <Divider orientation="left" plain>
```
antonmihaylov, line 109: Will there be a field for the type of display? E.g. if the user selects display: block -> don't display the FlexBoxEditor; if the user selects display: flex -> show FlexBoxEditor

libs/frontend/modules/element/src/css-editor/css-layout-editor/FlexBoxEditor.tsx
```diff
@@ -0,0 +1,101 @@
...
+import React, { useEffect, useState } from 'react'
+import { makeMenu } from '../utils/makeMenu'
+import { updateGuiCssProperty } from '../utils/updateGuiCssProperty'
+
+type FlexBoxEditorProps = {
+  element: IElement
+}
+
+export const FlexBoxEditor = observer(({ element }: FlexBoxEditorProps) => {
+  const [guiCssObj, setGuiCssObj] = useState<cssMap>(
```
antonmihaylov, line 14: Isn't it better to lift this up to ElementCssEditor, you'll need it for other forms too

libs/frontend/modules/renderer/src/renderPipes/atomRenderPipe.ts
```diff
@@ -42,7 +42,9 @@ export class AtomRenderPipe
       return this.next.render(element, props)
     }
 
-    const elCss = element.css ? css(evalCss(element.css)) : undefined
+    const elCss = element.customCss
```
antonmihaylov, line 45: You likely missed `| element.guiCss` here

antonmihaylov (commented):
Good work, just small comments

libs/backend/src/schema/element.schema.ts
mohasarc: Yeah, that makes sense. I'll do it.

libs/frontend/modules/element/src/css-editor/ElementCssEditor.tsx
mohasarc: I think we still need to debounce regardless? because for example the guiCss is not debounced for now; although it is working, sometimes glitches happen and the result in the builder looses sync with the selected value.

libs/frontend/modules/element/src/css-editor/ElementCssEditor.tsx
mohasarc: Currently when `flex-direction` is set to `none`, the field for display is deleted so it reset to default value. But we can also add `display` field.

libs/frontend/modules/renderer/src/renderPipes/atomRenderPipe.ts
mohasarc: Good catch, thanks :)

libs/frontend/modules/element/src/css-editor/ElementCssEditor.tsx
antonmihaylov: Yeah, debouncing makes sense. Could it be optimistic?

mohasarc: I'm not sure I understand, can you clarify?

webberwang:
good separation of files & distinction between gui & custom css

PR #1693, opened 2022-08-04
1692 import export app domains

### LINKED ISSUE #1692: Include app domains on export/import app
## Description
- Currently, domains are not included when exporting/importing an app

phmngocnghia:
## Related Issue(s)

Fixes #1692


apps/cli/src/use-cases/import/import-domains.ts
```diff
@@ -0,0 +1,32 @@
...
+  createVercelDomainIfNotExist,
+} from '../../repository/domain.repo'
+
+export const importDomains = async (domains: Array<IDomainExport> = []) => {
+  console.log('Importing domains...')
+
+  const verificationPromises: Array<Promise<any>> = []
+  const verifiedDomains: Array<IDomainExport> = []
+
+  domains.forEach((domain) => {
```
antonmihaylov, line 13: Why don't you use `verificationDomains = await Promise.all(domains.map(...))`

phmngocnghia: @antonmihaylov because some of the promises may fail. I'll add log incase they fail

webberwang: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/allSettled

PR #1703, opened 2022-08-06
InnerHtml prop editor

### LINKED ISSUE #1672: Remove the need to use a separate element for adding text
## Description

Adding text is a very common operation, using another element and attaching an atom of type text is unintuitive.
#1669  outlines other problems with it. 

## Proposed solution
- Add a text field here that will directly edit element.props.children as a text.  
image

- Dynamic state variable replacement should work naturally as it's just another prop, but test it

- We should disable it for some AtomTypes, for example adding it to input, other void HTML tags, or some AntD/MUI atoms that don't take in any children doesn't make sense.

- Check out how it will interact if we have both element.props.children and regular children. My guess is that the regular children will override the text, which sounds reasonable.
  e.g.: 
```
Parent {props: {children: "Some text"}}
    \_ Child
```
Will this render 
`<Parent><Child/></Parent>` (this seems more appropriate to me)
Or 
`<Parent>Some text</Parent>`

## Future ideas
- Double-clicking an element in the builder triggers it to become contenteditable and directly edits props.children that way

mohasarc:
#### demo

https://user-images.githubusercontent.com/51242349/183255261-3c41f9c3-462a-49bd-b8c1-384c94635f39.mp4

### TODO

- [x] Add the filed `InnerHtml` to update `props.dangerouslySetInnerHTML.__html` of an element
- [x] Prevent crash after adding elements by removing the value of `dangerouslySetInnerHTML` and disabling `InnerHtml`
- [x] Add HTML extension for syntax highlighting and autocompletion
- [ ] Remove Text Atom

## Related Issue(s)

Fixes #1672


libs/frontend/modules/element/src/use-cases/element/update-inner-html/UpdateInnerHtmlForm.tsx
```diff
@@ -0,0 +1,74 @@
...
+
+    return (
+      <Row align="middle">
+        <Col span={4}>
+          <p style={{ color: inEditMode() ? '#000000' : '#D1D1D1' }}>
+            InnerHtml
+          </p>
+        </Col>
+        <Col span={20}>
+          <CodeMirrorInput
```
antonmihaylov, line 57: Why don't you add the HTML extension for syntax highlighting and autocompletion https://github.com/codemirror/lang-html

antonmihaylov (changes requested):

libs/frontend/modules/element/src/use-cases/element/update-inner-html/UpdateInnerHtmlForm.tsx
mohasarc: That makes perfect sense. I'll do that.

PR #1711, opened 2022-08-10
Shortcut to Atom's interface

### LINKED ISSUE #1704: A shortcut button to edit/add props to an atom API
## Description
when opening the props tab of an element whose atom has no props, it is a bit confusing and unclear where/how to add a prop. It would be very helpful to have a button to directly add a prop to the atom API from the props tab. An alternative appproach would be just to add some text clarifying why is the tab empty and how to add a prop.

Screenshot from 2022-08-07 17-48-52

## Suggested Implementation
A button when clicked opens `typeService.fieldCreateModal`.

mohasarc:
## Description (Optional)
A shortcut button in props sidebar tab that redirects to the interface of the selected atom.

https://user-images.githubusercontent.com/51242349/183854198-fe7ee3b1-d74d-45f1-9dec-c96c19b65a6b.mp4

## Related Issue(s)
Fixes #1704


antonmihaylov:
I think a Link would be better since it would indicate better that it takes you to another page. Perhaps a small "Edit API" link?

PR #1714, opened 2022-08-10
refactor: actions and fix related issues

assimhabeek:
## Related Issue(s)

Fixes #


antonmihaylov:
Great, looks clean.
A downside that I notice is that if the pipeline contains async actions it won't wait for them to complete, i.e. they will all run in parallel. 

It seems like a rare edge case and we could deal with it if we encounter it, what do you think?

assimhabeek:
> Great, looks clean. A downside that I notice is that if the pipeline contains async actions it won't wait for them to complete, i.e. they will all run in parallel.
> 
> It seems like it would be a rare edge case if that causes an issue and we could deal with it if we encounter it, what do you think?

I tried to add an `await` if the action result is promise.

antonmihaylov:
That doesn't changr much, because they're in a .map and nothing handles the promises resulting from the map. You'd have to change it to a for..of for it to work. But still, I don't think it will be a common issue

PR #1719, opened 2022-08-12
Integrate WYSIWYG

### LINKED ISSUE #1710: Integrate WYSIWYG editor to customize texts
## Description

Replace the current text section of css editor and innerHtml of props tab with a text editor that handles all text related customization/styling.

## Use Case
Customize the looks of any string in the builder.

## Suggested Implementation
Integrate WYSIWYG editor

mohasarc:
## Description (Optional)
This PR: 
- Integrates the Quill Editor. We use some of the default formats; however, we customize the default stylesheet because it has some undesired styles like padding and overflow on the container.
- Removes Typography panel from Css Editor as it is replaced with Quill Editor.

https://user-images.githubusercontent.com/51242349/184859304-932740b0-2812-45cc-9528-a675b0cf2b64.mp4

## Related Issue(s)

Fixes #1710


apps/builder/pages/_app.tsx
```diff
@@ -1,6 +1,8 @@
 // import '../src/wdyr'
 import 'reflect-metadata'
 import '../src/styles/antd-theme.less'
+// This stylesheet is used to customize some of the default Quill editor's styles.
+import '../src/styles/quill.snow.custom.css'
```
webberwang, line 5: Any reason to not import quill.snow.css directly like the docs?

webberwang (changes requested):

apps/builder/pages/_app.tsx
mohasarc: Yes, the reason is that the default stylesheet adds some unwanted styles like padding and overflow. This means, for example, when some text is added into a button, the text will be padded and scroll bar will appear if it doesn't fit.

libs/frontend/modules/renderer/src/element/wrapper.utils.ts
```diff
@@ -35,10 +34,12 @@ export const getReactComponent = (renderOutput: IRenderOutput) =>
   // Render the atom if it exists, otherwise use fragment
   renderOutput.atomType ? getAtom(renderOutput.atomType) ?? Fragment : Fragment
 
-export const makeChildrenPropElement = (props: IPropData) =>
-  React.createElement('span', {
+export const makeCustomTextContainer = (customText: string) =>
+  React.createElement('div', {
```
antonmihaylov, line 38: Is there a reason not to use a span wrapper here again?

libs/frontend/modules/renderer/src/element/ElementWrapper.ts
```diff
@@ -51,9 +51,10 @@ export const ElementWrapper = observer<ElementWrapperProps>(
       if (
         hasNoChildren &&
         renderOutput.props &&
+        renderOutput.props['customText'] &&
         element.atom?.current.allowCustomTextInjection
       ) {
-        children = makeChildrenPropElement(renderOutput.props)
+        children = makeCustomTextContainer(renderOutput.props['customText'])
```
antonmihaylov, line 57: I think it would be better to refactor "customText" to a constant and refer to it both here and in the form

antonmihaylov (changes requested):

apps/builder/pages/_app.tsx
webberwang: what about importing the default sheet, then adding an override css file, so we only add rules that we need

libs/frontend/modules/renderer/src/element/wrapper.utils.ts
mohasarc: The reason is that quill uses `div` for some block elements like `blockquote` and `code-block`.

apps/builder/pages/_app.tsx
mohasarc: I've done some version of that, I've also imported the default stylesheet (into `libs/frontend/modules/element/src/use-cases/element/update-inner-html/ReactQuill.tsx`), but instead of overriding styles, I made the customText's container use the custom stylesheet. Now I see it actually make sense to only override the styles we want to change. I'll try to do that.

PR #1720, opened 2022-08-12
1698 dnd comp to builder el tree

### LINKED ISSUE #1698: Re-enable drag & drop, optimize for better use-ability
@phmngocnghia let me know your thoughts on this

---

Drag & drop can refer to multiple contexts:

## Component

We decided the component tab to go on the right panel (Config Pane)
- [ ] Drag & drop from component to the page should work
- [x] Drag & drop from component to the element page tree should also work

## Element

Dragging a component to the page/tree essentially creates a new element of type component. We should be also able to move existing elements around

- [ ] Drag & drop within the tree, this should be built-in Ant Design feature
- [ ] Drag & drop using the builder

## Bounding Box

- [ ] The parent bounding box should be highlighted, and will tell us which parent to add the node under.

phmngocnghia:
## Description
- drag and drop as a child, or descendants of target el
- visual only, no business features have been imp yet

https://user-images.githubusercontent.com/36730355/186071756-49c9534e-91f2-4c58-94f7-343ba98d0e55.mp4

## Related Issue(s)

Fixes #1698


apps/cli/src/use-cases/export/export-atom.ts
```diff
@@ -24,5 +24,5 @@ export const exportAtom = async (): Promise<ExportAtomData> => {
   //   return { atoms }
   // }
 
-  return { atoms }
```
webberwang, line 27: `as any` cast doesn't seem good, anything else we can use?

webberwang (changes requested):

PR #1723, opened 2022-08-14
feat: dynamic npm packages integration

### LINKED ISSUE #1696: Research dynamic NPM library integration
Users may want to integrate third party NPM libraries that we have not provided. See if this is possible & how we would achieve it. Explain the architecture & the steps we would need to take to get there.

assimhabeek:
## Related Issue(s)

Fixes #1696


libs/frontend/platform/atoms/src/custom/codelab-script/CodelabScript.tsx
```diff
@@ -0,0 +1,9 @@
+import Script, { ScriptProps } from 'next/script'
+import React from 'react'
+
+export const CodelabScript = (props: ScriptProps) => (
+  // eslint-disable-next-line react/jsx-props-no-spreading
+  <Script {...props} strategy="afterInteractive" />
```
antonmihaylov, line 6: Why do you override the strategy, what if the user wants to set it

antonmihaylov (commented):
Looks good. I think Webber was trying to say for users to create their own atoms + atom types. But this turned out well with the additional Script atom. 
And if they want to make it reusable they can always make it into a component.

libs/frontend/platform/atoms/src/custom/codelab-script/CodelabScript.tsx
assimhabeek: `beforeInteractive` is a server side loading which doesn't work for us. 
`worker` is still experimental and it didn't work when I tested it out.
we have `lazyOnload` and `afterInteractive` for now I forced the second one. but if we need on the  future we may change

PR #1733, opened 2022-08-16
1698 refactor element tree

### LINKED ISSUE #1698: Re-enable drag & drop, optimize for better use-ability

phmngocnghia:
## Description (Optional)
- Refactor element order to use linked list
- Children will form a list by linking to each other using next/prev property
- A root of a tree can be accessed by using `childrenRoot` property of a parent element

## Related Issue(s)

Fixes #1698


libs/frontend/modules/element/src/store/element.service.ts
```diff
@@ -162,7 +165,51 @@ export class ElementService
       throw new Error('No elements created')
     }
 
-    return this.hydrateOrUpdateCache(elements)
+    const hydratedElements = this.hydrateOrUpdateCache(elements)
```
webberwang, line 168: Could extract logic to 3 functions at the top level

`addChildElement` and `addNextSibling` or `addPrevSibling`

Then the 2 siblings functions would share code from a third function

webberwang (changes requested):

libs/backend/src/schema/element.schema.ts
```diff
@@ -12,12 +12,19 @@ export const elementSchema = gql`
 
   type Element {
     id: ID! @id(autogenerate: false)
+    nextSibling: Element @relationship(type: "TREE_NODE", direction: IN)
```
webberwang, line 15: shouldn't the relationship be "NODE_SIBLING"

"TREE_NODE" doesn't give us any indication about sibling relationship

libs/backend/src/schema/element.schema.ts
```diff
@@ -12,12 +12,19 @@ export const elementSchema = gql`
...
+    prevSibling: Element @relationship(type: "TREE_NODE", direction: OUT)
+
+    childrenRoot: Element @relationship(type: "TREE_ROOT", direction: IN)
```
webberwang, line 18: this should be "TREE_SUB_ROOT", since a tree only has 1 root

libs/frontend/modules/element/src/store/element.model.ts
```diff
@@ -410,6 +456,64 @@ export class Element
...
+
+    // next -> prev
+    if (this.nextSibling) {
+      // link next to prev if any
+      this.nextSibling.prevSiblingId = this.prevSibling?.id ?? null
+    }
+  }
+
+  @modelAction
+  linkSiblings({
```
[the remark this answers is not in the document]
webberwang: To clarify, we still need the conditional somewhere. We're just pushing it to the top most calling level, so as close to the event source as possible.

libs/frontend/modules/element/src/store/element.model.ts
```diff
@@ -410,6 +456,64 @@ export class Element
...
+    nextSiblingId,
+    parentElementId,
+  }: Parameters<IElement['linkSiblings']>[0]) {
+    this.prevSiblingId = prevSiblingId ?? null
+    this.nextSiblingId = nextSiblingId ?? null
+    this.parentId = parentElementId ?? null
+  }
+
+  @modelAction
+  syncLinkedSiblings() {
```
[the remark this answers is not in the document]
phmngocnghia: this fn is for orchestrating update state when moving an element, so adding, and removing element could happen at the same time and result in a batched state change, and prevent layout shift https://github.com/codelab-ai/builder/blob/347547fee4fc5ea3c803ab64c87d0a0dc53554fe/libs/frontend/modules/element/src/store/element.service.ts#L558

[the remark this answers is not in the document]
webberwang: Ah i see you left a comment there. Could you

- [ ] Move the comment to the class method (I think this is a better place to add it)
- [ ] Change name to something more appropriate, such as `syncCacheForSiblingLinks`

LinkedSiblings is more about the siblings, but SiblingLinks is focusing on the link. Ins this case I think it's more about the links

webberwang (changes requested):

libs/frontend/modules/element/src/use-cases/element/create-element/CreateElementModal.tsx
```diff
@@ -46,6 +47,29 @@ export const CreateElementModal = observer<CreateElementModalProps>(
   }) => {
     const onSubmit = async (data: ICreateElementDTO) => {
       const [element] = await elementService.create([data])
+      const { prevSiblingId, parentElementId } = data
+
+      const prevSibling = prevSiblingId
```
webberwang, line 52: why can't we use prevSiblingId directly?

libs/frontend/modules/element/src/use-cases/element/move-element/MoveElementForm.tsx
```diff
@@ -33,29 +32,67 @@ export const MoveElementForm = observer<MoveElementFormProps>(
...
+    }, [element.parentId, element.prevSiblingId])
+
     const onSubmit = (data: MoveData) => {
-      const promise = elementService.moveElement(
-        element.id,
-        data.parentElementId,
-        data.order,
-      )
+      const {
+        prevSiblingId: currentPrevsiblingId,
```
webberwang, line 46: note the camel case

libs/frontend/modules/element/src/use-cases/element/move-element/MoveElementForm.tsx
```diff
@@ -33,29 +32,67 @@ export const MoveElementForm = observer<MoveElementFormProps>(
...
+        parentElementId: currentParentElementId,
+      } = model
+
+      const { parentElementId, prevSiblingId } = data
+      const changePrevSibling = currentPrevsiblingId !== prevSiblingId
+      const changeParent = currentParentElementId !== parentElementId
+
+      // check this first because change parent could trigger change in prev to undefined
+      // because we move the target element to the begining of the tree
+      if (changeParent && parentElementId) {
```
webberwang, line 56: We should have a domain specific function that takes in parameters like sibling, parent etc

`shouldMoveAsSubRoot` etc

libs/shared/abstract/core/src/domain/element/element.service.interface.ts
```diff
@@ -55,6 +55,15 @@ export interface IElementService
       ICRUDModalService<Ref<IElement>, { element: Maybe<IElement> }>,
       'createModal'
     > {
+  // link element
+  linkElement(props: {
```
webberwang, line 59: this no longer exists right?

package.json
```diff
@@ -14,7 +14,7 @@
     "cli": "node dist/apps/cli/main.js",
     "docker:up": "scripts/docker/up.sh",
     "codegen": "yarn graphql-codegen",
-    "ogmgen": "cross-env GENERATE_OGM_TYPES=true nx serve builder",
+    "ogmgen:builder": "cross-env GENERATE_OGM_TYPES=true nx serve builder",
```
webberwang, line 17: builder doesn't seem necessary since landing would never have ogm types

webberwang (changes requested):

libs/frontend/modules/element/src/store/element.service.ts
```diff
@@ -261,54 +270,148 @@ export class ElementService
...
   })
 
   /**
-   * Moves an element to a different parent and/or order
+   * Moves an element to the next postion of target element
    */
+  @modelFlow
   @transaction
-  moveElement = (elementId: string, newParentId: string, newOrder?: number) => {
+  moveElementNextTo = _async(function* (
```
webberwang, line 303: `moveElementAsNextSibling` sounds more fitting to our domain

libs/frontend/modules/element/src/use-cases/element/create-element/CreateElementModal.tsx
```diff
@@ -46,6 +47,29 @@ export const CreateElementModal = observer<CreateElementModalProps>(
   }) => {
     const onSubmit = async (data: ICreateElementDTO) => {
       const [element] = await elementService.create([data])
```
webberwang, line 49: The create here should be called inside our "Use Case Driven Function" `createElementAsNextSibling`, `createElementAsRootChild`, `moveElementAsRootChild`, `moveElementAsNextSibling` etc

webberwang (changes requested):

libs/frontend/modules/element/src/store/element.model.ts
```diff
@@ -410,6 +451,88 @@ export class Element
...
+    if (this.parentElement.childrenRootId === this.id) {
+      this.parentElement.childrenRootId = this.nextSiblingId
+    }
+
+    this.parentElement.removeChild(this)
+    this.parentId = null
+  }
+
+  @modelAction
+  attachToParent(parentElementId: string) {
```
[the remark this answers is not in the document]
phmngocnghia: yes, in case we don't need to attach to subroot. We won't need to attach to parent if we make parent as a computed property. This is just for the current implementation context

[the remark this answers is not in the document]
webberwang: Which use case would we not attach as subRoot?

webberwang:
think we've refactored enough, I'll check take a look down the line

PR #1741, opened 2022-08-20
Enum type dropdown

### LINKED ISSUE #1670: Interface Form Enum
Enum value should be dropdown option

Home___Builder___Codelab

Mouse_Highlight_Overlay

mohasarc:
## Related Issue(s)

Fixes #1670

### DIFF
--- a/libs/frontend/modules/type/src/props-form/form-controls/Field.tsx
+++ b/libs/frontend/modules/type/src/props-form/form-controls/Field.tsx
@@ -12,6 +12,7 @@ import { CheckboxField } from './CheckboxField'
 import { CodeMirrorField } from './CodeMirror'
 import { SelectActionField } from './SelectActionField'
 import { SelectComponentField } from './SelectComponentField'
+import { SelectEnumField } from './SelectEnumField'
 
 export type FieldProps = {
   field: IField
@@ -65,6 +66,10 @@ export const Field = observer(({ field, form, context }: FieldProps) => {
       ) : (
         <CodeMirrorField context={context} field={field} form={form} />
       )
+    }
+
+    case ITypeKind.EnumType: {
+      return <SelectEnumField field={field} form={form} />
     }
 
     default:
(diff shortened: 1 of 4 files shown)


libs/frontend/modules/type/src/props-form/form-controls/Field.tsx
```diff
@@ -12,6 +12,7 @@ import { CheckboxField } from './CheckboxField'
 import { CodeMirrorField } from './CodeMirror'
 import { SelectActionField } from './SelectActionField'
 import { SelectComponentField } from './SelectComponentField'
+import { SelectEnumField } from './SelectEnumField'
 
 export type FieldProps = {
   field: IField
```
webberwang, line 18: We could make `IField<T = IAnyType>`, so it allows us to pass a more specific subtype so we don't have to cast

webberwang (changes requested):

libs/frontend/modules/type/src/props-form/form-controls/Field.tsx
mohasarc: Making `IField` generic alone did not fix the issue because aparently typescript does not infer the type in switch statements over nested properties (https://stackoverflow.com/questions/65139900/why-doesnt-typescript-infer-type-in-switch-statements-over-deep-nested-properti). So I had to change the switch statement into if/return clauses applying the workaround outlined in the same stackoverflow link.

libs/frontend/modules/type/src/props-form/form-controls/Field.tsx
```diff
@@ -15,54 +23,61 @@ import { SelectComponentField } from './SelectComponentField'
...
+  }
 
-    default:
-      return <CodeMirrorField context={context} field={field} form={form} />
+  if (isOfTypeKind<IField<IEnumType>>(field, ITypeKind.EnumType)) {
+    return <SelectEnumField field={field} form={form} />
   }
+
+  return <CodeMirrorField context={context} field={field} form={form} />
 })
```
mohasarc, line 83: @webberwang This is the change.

webberwang: yeah this is a good solution, thanks

PR #1750, opened 2022-08-23
feat: import/export tag in seed data

### LINKED ISSUE #1747: Support tags for import/export
- add-antd-use-case-tags & add-antd-atoms-icons should be extracted to a new function with prefix `__` similar to `__seedData`.
- Comment this function after the seed-data.json is updated with the appropriate data

phmngocnghia:
## Related Issue(s)

Fixes #1747


webberwang (changes requested):
Our import tag logic should upsert tag using existing id, if we use name, we won't be able to update tag names.

App upsert uses id which is correct

apps/cli/src/use-cases/parser/ant-design.ts
```diff
@@ -11,6 +18,35 @@ export interface AntdDesignApi {
   isEnum: boolean
 }
 
+export const createAntdesignTagsData = async () => {
```
webberwang, line 21: Can we add a signature here

apps/cli/src/commands/import/import-seed-data/import-seed-data.ts
```diff
@@ -36,11 +37,16 @@ export const importSeedData = async (
  * Once data is in JSON format, we use that to import
  */
 export const __seedData = async (selectedUser: string) => {
```
webberwang, line 39: This should be broken into 2 functions `__seedAtomFromParser` & `__seedAtomWithTags` etc

The `__seedAtomFromParser` is already ran by admin and likely won't need to be run again

apps/cli/src/commands/import/import-seed-data/import-seed-data.ts
```diff
@@ -6,7 +6,10 @@ import { builderComponentUsecaseTag } from '../../../data/tag'
 import { importAtom } from '../../../use-cases/import/import-atom'
 import { importTags } from '../../../use-cases/import/import-tags'
 import { importType } from '../../../use-cases/import/import-type'
-import { createAntDesignAtomsData } from '../../../use-cases/parser/ant-design'
+import {
+  createAntDesignAtomsData,
+  createAntdesignTagsData,
```
webberwang, line 11: use `AntDesign` for naming

apps/cli/src/repository/atom.repo.ts
```diff
@@ -33,25 +33,28 @@ export const upsertAtom = async (atom: IAtomExport, userId: string) => {
...
     ...baseInput,
-    tags: [
-      {
-        connectOrCreate: connectOrCreateTags,
-      },
-    ],
+    tags: [{ connect: connectTags }],
   }
 
   if (!idExisting.length) {
```
webberwang, line 60: `existingId` reads better or `existingAtomId`

apps/cli/src/repository/tag.repo.ts
```diff
@@ -2,62 +2,70 @@ import { TagOGM } from '@codelab/backend'
...
+  selectedUser: string,
+) => {
+  const Tag = await TagOGM()
 
-      if (tag.parent) {
-        createInput.parent = {
-          connectOrCreate: connectOrCreateParent,
-        }
-      }
+  const nameExisting = await Tag.find({
```
webberwang, line 36: `existingName` is more grammatically correct

apps/cli/src/use-cases/import/import-tags.ts
```diff
@@ -1,19 +1,31 @@
...
+import { createOrUpdateTag, linkTag } from '../../repository/tag.repo'
 
 export const importTags = async (
   tags: Array<ITagExport> = [],
   selectedUser: string,
 ) => {
   console.log('Importing tag...\n')
 
-  for (const tag of tags) {
+  const promiseCreateTags = tags.map((tag) => {
```
webberwang, line 11: `createTagsOperations`

apps/cli/src/use-cases/import/import-tags.ts
```diff
@@ -1,19 +1,31 @@
...
     console.log('\n')
 
-    await upsertTag(tag, selectedUser)
-  }
+    return createOrUpdateTag(tag, selectedUser)
+  })
+
+  await Promise.all(promiseCreateTags)
+
+  const promiseSyncTags = tags.map((tag) => {
```
webberwang, line 22: `syncTagsOperations`, the fact it's a promise shouldn't have to be reflected in the time, it's just a technicality, not a domain concern

apps/cli/src/use-cases/parser/ant-design.ts
```diff
@@ -1,6 +1,13 @@
...
+import {
+  AtomOGM,
+  atomSelectionSet,
+  TagOGM,
+  tagSelectionSet,
+} from '@codelab/backend'
 import { IAtomExport, IAtomType } from '@codelab/shared/abstract/core'
+import { componentTagName } from '@codelab/shared/data'
 import { v4 } from 'uuid'
+import { antTagNames } from '../../commands/import/import-seed-data/add-antd-use-case-tags'
```
webberwang, line 10: `add-antd-tags` is good

apps/cli/src/use-cases/parser/ant-design.ts
webberwang: `AntDesign` naming

libs/shared/abstract/core/src/domain/tag/tag.dto.interface.ts
```diff
@@ -19,7 +19,8 @@ export type ITagGraphDTO = ITagDTO & {
 }
 
 export type ITagExport = {
-  name: string
-  children: Array<Pick<ICreateTagDTO, 'name'>>
-  parent?: Pick<ICreateTagDTO, 'name'>
+  id?: string
+  name?: string
```
webberwang, line 23: why couldn't we make `id` & `name` required?

webberwang (changes requested):

apps/cli/src/commands/import/import-seed-data/import-seed-data.ts
phmngocnghia: image

@webberwang do you mean something like this? `__seedAtomFromParser ` is old fn, and `__seedAtomWithTags ` is a new fn accepting atoms as parameter, transform, and update them

webberwang: I mean usage would be like this, we can individually comment out different types of seeder

builder_–_import_command_ts

apps/cli/src/repository/tag.repo.ts
```diff
@@ -2,62 +2,70 @@ import { TagOGM } from '@codelab/backend'
...
-  try {
-    const Tag = await TagOGM()
-
-    const nameExisting = await Tag.find({
-      where: {
-        name: tag.name,
-      },
-    })
+export const linkTag = async (tag: ITagExport) => {
+  const connectChildrens =
```
webberwang, line 6: children is already the plural

apps/cli/src/repository/tag.repo.ts
```diff
@@ -2,62 +2,70 @@ import { TagOGM } from '@codelab/backend'
...
-export const upsertTag = async (tag: ITagExport, selectedUser: string) => {
-  try {
-    const Tag = await TagOGM()
-
-    const nameExisting = await Tag.find({
-      where: {
-        name: tag.name,
-      },
-    })
+export const linkTag = async (tag: ITagExport) => {
```
webberwang, line 5: The function name could be bit more descriptive, linkTag to what?

something like `connectChildTagToParent` would be better. long function names are okay

webberwang (changes requested):

apps/cli/src/commands/import/import.command.ts
```diff
@@ -64,7 +66,9 @@ export const importCommand: CommandModule<any, any> = {
     }
 
     // Only used by admin
-    // await __seedData(selectedUser)
+    // await __seedData(selectedUser, antdAtomsFactory)
```
webberwang, line 69: This function no longer exists right

webberwang (changes requested):

PR #1752, opened 2022-08-24
Stuck panel border fix

### LINKED ISSUE #1668: Border stuck when re-sizing panels
Happens for both the vertical & horizontal panels. When re-sizing, then border stays in the same place at times.

web_–_Text_tsx_and_Home___Builder___Codelab_and_codelab-ai_web__Web_application_builder

## Steps to reproduce:
- Load the builder page.
- Resize the browser window or open dev tools.
- Resize a vertical or horizontal pane.

mohasarc:
## Description

After window resize, aparently, motion decides to translate the border with the same delta when resizing a panel. The current solution is just forcing 0px translation.

## Related Issue

Fixes #1668

### DIFF
--- a/libs/frontend/view/components/resizable/useResizable.ts
+++ b/libs/frontend/view/components/resizable/useResizable.ts
@@ -119,6 +119,7 @@ export const useResizable = ({
     xDragHandleProps: {
       onDrag: handleXDrag,
       style: {
+        translateX: '0px !important',
         cursor: 'col-resize',
       },
       onDragEnd: () => setIsDraggingX(false),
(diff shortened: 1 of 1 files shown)


webberwang (changes requested):
Before I merge, could you explain how translateX/Y fixed this issue?

Is it because on resize the translateX/Y moves the border?

mohasarc:
> Before I merge, could you explain how translateX/Y fixed this issue?
> 
> Is it because on resize the translateX/Y moves the border?

Yes exatcly, it translates the border with the same delta away from the original location.

PR #1754, opened 2022-08-26
1698 refactor element link service

### LINKED ISSUE #1770: Extract out element link service

phmngocnghia:
## Related Issue(s)

Fixes #1770


libs/frontend/modules/element/src/store/element.link.service.ts
```diff
@@ -0,0 +1,456 @@
+import { getElementService } from '@codelab/frontend/presenter/container'
```
webberwang, line 1: should be named `element-link.service.ts`

PR #1757, opened 2022-08-28
Renderer Error Handling Module

### LINKED ISSUE #1214: Prop validation and fallbacks
When we first add an element to a tree, if it has required props, it could throw an error, which requires reloading the page to go away.

1 fix would be to validate the props before rendering the element and if it doesn't match (e.g. a required prop is not included) - to not render it and/or to display a warning.

image

For this to be executed properly we need to first store all/most of the API interfaces

mohasarc:
## Description

This PR introduces prop validation to the rendering process by introducing errors into elements. The `renderer.model` would check if there are any errors in the element. Errors include missing required props for example.

Note: had to move `renderer/src/atoms` to `atom/src/atoms` to avoid circular dependencies.

## Related Issue

Fixes #1214


mohasarc:
Closed in favor of #1863 and #1897

PR #1776, opened 2022-09-04
1755 disallow deleting root element

melhamin:
fix: disable deleting the root element using hotkeys

## Description (Optional)

This PR disables deleting the root element of a page using hotkeys. 
It also includes fix for the bug where pressing `del` on a non-root element would open delete modal for the root element.

## Related Issue(s)

Fixes #1775 @melhamin make sure this format


webberwang:
looks good!

PR #1783, opened 2022-09-05
1698 integrate drag and drop component picker

### LINKED ISSUE #1697: Improve component picker
## Description
- Current component picker is implemented using dropdown to select either atom, or component which is not intuitive. We need a better component picker

## Suggested Implementation
- Organize them using group. Eg: Form, Layout, Navigation
- Add icon to improve the component's intuitive
- Merge both atom, and component to list. Add a tag to differentiate them

Refs:

image

phmngocnghia:
## Related Issue(s)

Fixes #1697

### DIFF
--- a/libs/frontend/modules/builder/src/dnd/useBuilderDnd.ts
+++ b/libs/frontend/modules/builder/src/dnd/useBuilderDnd.ts
@@ -20,7 +20,7 @@ export interface UseBuilderDnd {
 export const useBuilderDnd = (
   builderService: IBuilderService,
   elementService: IElementService,
-  elementTree: IElementTree,
+  elementTree?: IElementTree,
 ): UseBuilderDnd => {
   const onDragStart = useCallback(
     (e: DragStartEvent) => {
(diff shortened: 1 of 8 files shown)


libs/frontend/modules/builder/src/dnd/useBuilderDnd.ts
```diff
@@ -17,6 +20,7 @@ export interface UseBuilderDnd {
 export const useBuilderDnd = (
   builderService: IBuilderService,
   elementService: IElementService,
+  elementTree?: IElementTree,
```
webberwang, line 23: should this be required?

webberwang (changes requested):

libs/frontend/modules/builder/src/dnd/useBuilderDnd.ts
phmngocnghia: It can be undefined https://github.com/codelab-app/builder/pull/1783/files/fead5bd3a7d264051e864152d37253255c959910#diff-2de6b809f6667793afc53f7f609f97cc771d4c23575703701e77f84b3ceba486R174

If it's undefined, I don't want to pause the operation

webberwang: the pageTree should be defined before the component renders, it only shows as undefined because we use `.get()` on the object map

webberwang (approved):

PR #1789, opened 2022-09-06
refactor: element service

### LINKED ISSUE #1786: ElementService refactor
- [x] `ElementService.deleteElementSubgraph` why call `detachElementFromTreeSiblings` if we already have all the elements that need to be deleted
- [x] Using bind makes it more difficult to track where the parameters are coming from, the function signature gives a lot of information on what it does
- [x] Why do we need to push here? This pattern isn't great because it scatters the code & is unnecessary here
image
- [x] You could always detach everything & we wouldn't need so many functions. I don't think we would need to extract to these 3 functions since it's not re-used anywhere else
```
element.makeDetachParentInput(),
element.makeDetachNextSiblingInput(),
element.makeDetachPrevSiblingInput(),
```
- [x] Leave more comments, basically write out your thought process for more complex parts

image

For example, here we can say that `If we want to attach an element as subRoot to a parent that already has a subRoot, we prepend the new element as prevSibling to the parent's existing subRoot`

It's much faster to understand the motivation rather than having to read the code

- [x] `detachElementFromTreeSiblings` naming seems odd, reading that it makes you think it only unlinks from siblings, but it does also unlink from subRoot
- [x] `Element.childrenRoot` should be `Element.firstChild`, our domain has the concept of ordering with siblings, so first child is always the first one
- [x] I recommended `rootOf` to `subRootOf` but didn't look a the full picture. It should really just be `parent` in this case.

We still have `parent`, but only a firstChild can have a parent, other elements have prevSiblings

- [x] Do we have a type here we can use
image
- [x] I don't think we need a custom function for updates, the default UpdateElement already takes in an array, you can see the signature here
image
- [x] Try to leave more comments, especially before `if` statements or use case functions. These type of code are core logic and could use more insight for new developers on what is happening (without having to go deep into the code)
- [x] https://github.com/codelab-ai/builder/blob/3e654dfdcfc314663e7ab9e18ea8a34642321ef6/libs/frontend/modules/builder/src/store/builder.service.ts#L72 

The camel case is off, the naming of the variables are confusing, casting should be removed, don't use push if not needed. The forEach block doesn't seem to do anything??
- [x] You can filter out nulls or undefined so you don't need to cast, I noticed several places with this. 
`BuilderService.tagsWithComponents`, `BuilderService.componentTags`
https://stackoverflow.com/questions/43118692/typescript-filter-out-nulls-from-an-array
https://dev.to/ylerjen/typescript-cast-is-a-type-breaker-1jbh
- [x] Don't use cast if you can avoid it, having a single object here is also more readable
image
This happened many times in cypress, but I've changed it now

- Don't use cast, push or let when you can avoid it

phmngocnghia:
## Description (Optional)
TODO: make `parentElement` computed in client side

## Related Issue(s)

Fixes #1786


libs/frontend/modules/element/src/store/element.model.ts
```diff
@@ -598,12 +611,16 @@ export class Element
     const sibling = this.elementService.element(siblingId)
 
     if (!sibling) {
-      throw new Error(`sibling element ${siblingId} not ond`)
+      throw new Error(`sibling element ${siblingId} not found`)
     }
 
+    // sibling - next sibling
+    // sibling - [element]
     return makePatchElementInput(sibling, {
```
webberwang, line 619: what is different between patch and update? why not call it `makeUpdateElementInput`?

webberwang (changes requested):

libs/frontend/modules/element/src/store/element.model.ts
phmngocnghia: It accepts `element` param to fill in the `where`

webberwang: is update name already taken?

libs/frontend/modules/element/src/graphql/element.endpoints.graphql.custom.ts
```diff
@@ -19,15 +17,15 @@ export type SdkFunctionWrapper = <T>(
 export const getSdk = (client: GraphQLClient) => {
   return {
     BatchUpdateElements(
```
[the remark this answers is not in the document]
phmngocnghia: image

we can update many elements with multiple changes but with only one joined condition containing multiple where. Like

Update "a=1" where "b=1 or c=2"....

I need a function that would update
- a=1 when b=1
- b=2 when c =1
- ... etc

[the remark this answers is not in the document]
webberwang: okay my mistake, I thought batching mutations had issues?

webberwang (changes requested):
some rebase issue, also don't use abbreviations, use `service` instead of `svc`

phmngocnghia:
https://github.com/codelab-ai/builder/pull/1789/files#diff-d8d2d09015339cd1dec0bd1332df7fafb6073251273d4a1608109e04cd284c94 this file is not used anywhere

PR #1794, opened 2022-09-08
fix: prop data left over after delete

### LINKED ISSUE #1758: Residual left after prop is deleted
## Current Behavior

although a prop is deleted from the atom's api, the elements of that prop keep having that prop and its last value before deletion.

## Expected Behavior

No residual should be left.

## Steps to Reproduce

https://user-images.githubusercontent.com/51242349/187069078-58a993be-74c7-4582-a51e-60daf0ffd3dc.mp4

melhamin:
## Description (Optional)
This PR fixes prop data left over in the elements after the prop is deleted from an atom API. 

## Related Issue(s)

Fixes #1758

### DIFF
--- a/libs/frontend/modules/element/src/store/element.service.ts
+++ b/libs/frontend/modules/element/src/store/element.service.ts
@@ -22,6 +22,7 @@ import {
   IUpdatePropMapBindingDTO,
 } from '@codelab/shared/abstract/core'
 import { IEntity, Nullable } from '@codelab/shared/abstract/types'
+import { omit } from 'lodash'
 import {
   _async,
   _await,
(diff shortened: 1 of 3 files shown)


libs/frontend/modules/element/src/store/element.service.ts
```diff
@@ -778,4 +779,33 @@ export class ElementService
...
+    this: ElementService,
+    interfaceType: IInterfaceType,
+    propKey: string,
+  ) {
+    const elementsThatUseTheProp = yield* _await(
+      this.getAll({ atom: { api: { id: interfaceType.id } } }),
+    )
+
+    const promises = elementsThatUseTheProp.map((element) => {
+      const currentProps = { ...element.props?.data }
```
webberwang, line 795: I think something like https://lodash.com/docs/4.17.15#omit is more clear, the performance isn't a huge concern here

webberwang (changes requested):

PR #1819, opened 2022-09-13
refactor: make parent element computed

phmngocnghia:
## Related Issue(s)

#1698

### DIFF
--- a/apps/builder-e2e/project.json
+++ b/apps/builder-e2e/project.json
@@ -8,6 +8,7 @@
       "options": {
         "cypressExecutor": "e2e"
       },
+      // Need this empty block to forward the configurations
       "configurations": {
         "ci": {},
         "test": {},
--- a/apps/builder-e2e/project.json
+++ b/apps/builder-e2e/project.json
@@ -8,7 +8,6 @@
       "options": {
         "cypressExecutor": "e2e"
       },
-      // Need this empty block to forward the configurations
       "configurations": {
         "ci": {},
         "test": {},
@@ -87,7 +86,7 @@
               "command": "npx wait-on 'http://127.0.0.1:3001' && echo 'READY'"
             },
             {
-              "command": "nx serve-test builder -c test",
+              "command": "nx run builder:serve-test:dev",
               "forwardAllArgs": false
             }
           ]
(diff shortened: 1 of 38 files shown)


phmngocnghia:
image

apps/builder-e2e/project.json
```diff
@@ -8,7 +8,6 @@
       "options": {
         "cypressExecutor": "e2e"
       },
-      // Need this empty block to forward the configurations
```
webberwang, line 11: This comment should be left here

libs/frontend/modules/element/src/store/element-tree.model.ts
```diff
@@ -83,19 +91,18 @@ export class ElementTree
...
-    for (const element of elements) {
-      /**
-       * For ElementTree that is already initialized, we need this to update its elements
-       */
-      if (!this._elements.find((el) => el.current.id === element.id)) {
-        this._elements.push(elementRef(element))
-      }
+  addElements(elements: Array<IElement>) {
+    // filter elements that are already on the tree
+    const isElementNotInTree = (element: IElement) =>
```
webberwang, line 96: instead of filtering out existing elements, we can call `updateCache` on existing elements

libs/frontend/modules/element/src/store/element-tree.model.ts
```diff
@@ -83,19 +91,18 @@ export class ElementTree
...
-      if (!element.parentElement?.id) {
-        this.set_root(elementRef(element))
-      }
+    const newElements = elements.filter(isElementNotInTree)
+
+    // add reference to new element
+    newElements.forEach((element) => this._elements.push(elementRef(element)))
 
+    // validate component meta data
+    newElements.forEach((element) => {
```
webberwang, line 105: couldn't we share the same forEach loop?

webberwang (changes requested):

libs/shared/abstract/core/src/domain/element/element.model.interface.ts
```diff
@@ -38,23 +39,22 @@ export interface IElement
   hooks: Array<IHook>
   parentId: Nullable<string>
   parentElement: Maybe<IElement>
+  parentElementFromId: Maybe<IElement>
```
webberwang, line 42: `parentElementId` is good

webberwang (changes requested):

PR #1822, opened 2022-09-13
Fix errors when re-ordering using dnd

### LINKED ISSUE #1780: Elements disappear after drag-drop and app becomes undeletable.
## Current Behavior

Elements disappear from the builder after rearranging them using drag and drop. 
Also the app becomes undeletable. To delete the app, the broken page should be deleted first then the app.

## Expected Behavior

Rearranging elements using dnd should not cause any issues.

## Steps to Reproduce

https://user-images.githubusercontent.com/48331678/188312044-5c00af10-2eed-4dbd-9aae-1f1e918838fe.mov

melhamin:
This PR fixes errors while re-ordering elements using dnd.

## Related Issue(s)

Fixes #1780


libs/frontend/modules/element/src/store/element.service.ts
```diff
@@ -302,51 +300,31 @@ parent
...
+      // Detach from prev sibling
       element.makeDetachPrevSiblingInput(),
     ]
 
-    updateElementCacheFns.push(
+    const updateElementCacheFns: Array<() => void> = [
+      // Detach from parent
       element.detachParent(),
+      // Attach next to prev
+      element.attachPrevToNextSibling(),
```
webberwang, line 315: For all these, if we used `element.attachPrevToNextSibling` and removed the `() => {}` inside the method, would that work?

webberwang (changes requested):

libs/frontend/modules/element/src/store/element.service.ts
melhamin: I believe it should work.

melhamin: No sorry it doesn't work

webberwang (changes requested):
Could remove the custom element endpoint for the batch mutation since that can cause issues

PR #1839, opened 2022-09-17
fix: cannot see/edit props in UpdateElementPropsForm

### LINKED ISSUE #1837: Atom props don't show up on the Inspector->Props tab
## Current Behavior

Atom props cannot be seen/set in the Inspector->Props tab. An error is thrown when the props inspector tab is opened.

## Expected Behavior
Atom props should be seen and set in the props inspector tab.

## Steps to Reproduce

https://user-images.githubusercontent.com/48331678/190680012-db43123f-08ac-42d2-affa-9afcdbe7c93a.mov

melhamin:
## Related Issue(s)

Fixes #1837

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -164,7 +164,7 @@ jobs:
       - run:
           name: 'Build CI'
           # Can't use `yarn cli` doesn't exist yet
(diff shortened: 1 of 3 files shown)


webberwang:
Ah yes it was the useAsyc refactor

PR #1853, opened 2022-09-20
refactor: migrate init element tree v1 to v2

### LINKED ISSUE #1850: Migrate ElementTree.initTree to ElementTree.initTreeV2

phmngocnghia:
## Related Issue(s)

Fixes #1850


apps/builder/pages/apps/[appId]/pages/[pageId]/index.tsx
```diff
@@ -70,10 +69,6 @@ const PageRenderer: CodelabPage<any> = observer(() => {
       owner: { auth0Id: userService.auth0Id },
     })
```
Comment (phmngocnghia) [inline apps/builder/pages/apps/[appId]/pages/[pageId]/index.tsx:71]: components are loaded in `elementService.getAll`

webberwang (changes requested):
Looking good so far, we'll want to replace all instances of `initTree` with `initTreeV2`.

The main motivation was that `initTree` creates an API call, so we would have waterfall API calls reducing performance, where as `initTreeV2` could allow us to optimize the query call such that we have a single call at the top level.

PR #1858, opened 2022-09-22
feat: migrate prop from to use uniform

### LINKED ISSUE #1854: Migrate Propsform to use uniforms
## Current Implementation Concern

Currently `PropsForm` uses `react-hook-form` to handle inputs. while the rest of the platform uses `uniforms`. one the of main issues is that its custom Fields like `SelectComponentField` are not re-usesable inside other Modals.

phmngocnghia:
## Description
- Use `interfaceForm` in `PropForms`
- Sync CodeMirror behaviors from old `PropField`:
  - Add auto-complete context
  - Primative types use CodeMirror
  - Enum select is searchable
  - 
- The expand button hides the editor's content
image
- CodeMirror interprets data as string, so inputting non-string types (int,float) results validation error
image

Fixing by putting it on the right side to reduce the probability of it blocks the input contents, and add opacity to the expand button so we can see behind it
image

## Notes:

- Atoms page should load all types if types are not available

## Related Issue(s)

Fixes #1854


libs/frontend/modules/builder/src/sections/config-pane/ConfigPane-InspectorTabContainer/ConfigPane-InspectorTabContainer.tsx
```diff
@@ -111,7 +111,7 @@ export const ConfigPaneInspectorTabContainer = observer<MetaPaneBuilderProps>(
           />
         </div>
 
-        <Tabs defaultActiveKey="tab1" size="small">
+        <Tabs defaultActiveKey="tab2" size="small">
```
webberwang, line 114: we could refactor to use more descriptive key names

libs/frontend/modules/type/src/interface-form/type-schema.factory.ts
```diff
@@ -177,8 +178,8 @@ export class TypeSchemaFactory {
     return { type: primitives[type.primitiveKind], ...extra }
   }
 
-  fromEnumType(type: IEnumType): JsonSchema {
-    const extra = this.getExtraProperties(type)
+  fromEnumType(type: IEnumType, context?: any): JsonSchema {
```
webberwang, line 181: possible to type this? along with all other any context

webberwang (changes requested):

PR #1861, opened 2022-09-22
Supports union type and enum type

### LINKED ISSUE #1814: Support union types
Relates: #1809 

When we run `yarn cli parse`, this inserts seed data to our data, which we can export as JSON. Currently union types are not supported.

## Simple Union

Take button for example...

`Button.shape: default | circle | round`
`Button.size: large | middle | small`

These 2 are simple union types, we could have a generator that parses the above.

## Complex Union

`Button.loading: boolean | { delay: number }`

This is a more complex union type, which contains a nested interface, we don't have to support this just yet.

---

- [ ] Support simple union type with only strings (no nested interfaces type

hauleIT:
## Related Issue(s)

Fixes #1814


apps/cli/src/use-cases/parser/type-map.ts
```diff
@@ -35,18 +59,177 @@ export const getTypeForApi = async (
   const ReactNodeType = await ReactNodeTypeOGM()
   const RenderPropsType = await RenderPropsTypeOGM()
   const EnumType = await EnumTypeOGM()
+  const UnionType = await UnionTypeOGM()
+  const values = apiField.type.split('|').map((value) => value.trim())
+  const isBaseCondition = apiField.type.indexOf('|') > -1
```
webberwang, line 64: `includes` works too?

apps/cli/src/use-cases/parser/type-map.ts
```diff
@@ -35,18 +59,177 @@ export const getTypeForApi = async (
...
+
+    if (choose === 'name') {
+      return key
+    } else {
+      return value
+    }
+  }
+
+  const isEnumType =
+    values.includes('small') ||
```
webberwang, line 95: I thought we used the CODE logic for enum, this hard coding words isn't good

webberwang (changes requested):

apps/cli/src/use-cases/parser/type-map.ts
```diff
@@ -148,8 +373,21 @@ export const getTypeForApi = async (
       }
     }
 
-    case 'number | string':
-    case 'string | number':
+    case 'number | string': {
```
webberwang, line 376: for all these strings, we should split as array, then use a single function to check instead of having separate cases

apps/cli/src/use-cases/parser/type-map.ts
```diff
@@ -38,18 +85,99 @@ export const getTypeForApi = async (
...
+      }
+    }
+
+    return {
+      existingId: existingUnion.id,
+    }
+  }
+
+  // Check and Create Enum Type
+  if (apiField.isEnum) {
```
webberwang, line 170: we should have a file for each type, then import related functions here

webberwang (changes requested):

apps/cli/src/use-cases/parser/type-map.ts
```diff
@@ -123,7 +247,108 @@ export const getTypeForApi = async (
     }
   }
 
-  switch (type) {
+  const a = values
```
webberwang, line 250: use a more description variable

webberwang (changes requested):

apps/cli/src/use-cases/parser/type-map.ts
```diff
@@ -26,151 +19,37 @@ type TypeRef = {
...
 ): Promise<TypeRef> => {
   logTask('Get Type For API', atom.name, apiField)
 
-  const type = apiField.type.trim()
-  const PrimitiveType = await PrimitiveTypeOGM()
-  const ReactNodeType = await ReactNodeTypeOGM()
-  const RenderPropsType = await RenderPropsTypeOGM()
-  const EnumType = await EnumTypeOGM()
+  const values = apiField.type.split('|').map((value: string) => value.trim())
+  const isBaseCondition = apiField.type.indexOf('|') > -1
```
webberwang, line 31: why not use includes instead?

webberwang (changes requested):

apps/cli/src/use-cases/parser/type-map.ts
webberwang: @hauleIT any thoughts on this?

apps/cli/src/use-cases/parser/type-map.ts
webberwang: @hauleIT thoughts?

apps/cli/src/use-cases/parser/types/union-type-map.ts
```diff
@@ -0,0 +1,150 @@
...
+
+export const getUnionTypeForApi = async (
+  apiField: AntdDesignApi,
+  atom: IAtomExport,
+  userId: string,
+  values: Array<string>,
+): Promise<TypeRef> => {
+  const UnionType = await UnionTypeOGM()
+
+  const getKeyObjInUnion = (data: string, choose: string) => {
```
webberwang, line 29: this can be replaced by learning regex, it's a bit hard but much more concise

webberwang (changes requested):

apps/cli/src/use-cases/parser/type-map.ts
hauleIT: Yes, I will try

apps/cli/src/use-cases/parser/types/union-type-map.ts
hauleIT: I got it, I will research regex

apps/cli/src/use-cases/parser/type-map.ts
```diff
@@ -26,151 +19,42 @@ type TypeRef = {
...
-          name: IPrimitiveTypeKind.String,
-        },
-      })
-
-      return {
-        existingId: stringType.id,
-      }
-    }
+  // Check if type is Complex Union Type
+  if (isBaseCondition) {
```
webberwang, line 49: why not name it to `isUnionType`? Also we should have `!isEnum` requirement as well. We are relying on `isEnum` check being first (which isn't ideal).

webberwang (changes requested):
- [ ] any instances of primitive string types should use an enum we create specifically for the ant design docs string format

apps/cli/src/use-cases/seed/types/union-type-map.ts
```diff
@@ -0,0 +1,258 @@
...
+      return null
+    }
+  }
+
+  //
+  await checkPrimitiveOfUnionType(values)
+
+  const isInterfaceTypeRegex = /^\{.+.}$/
+
+  const getKeyObjInUnion = (data: string, choose: string) => {
```
webberwang, line 138: Could you comment what this function does? Can be replaced with a regex

webberwang (changes requested):

PR #1873, opened 2022-09-26
feat: add pagination to types table

### LINKED ISSUE #1820: Add pagination to types page using Option 2

phmngocnghia:
## Related Issue(s)

Fixes #1820


libs/backend/adapter/neo4j/src/schema/type/type.schema.ts
```diff
@@ -46,13 +52,17 @@ export const typeSchema = gql`
     """
     getTypeReferences(typeId: ID!): [TypeReference!]
       @cypher(statement: """${getTypeReferences}""")
+
+   typesOfTypesPage(
+    options: TypesOfTypesPageOptions
+   ): [TypesPageAnyType!]!
```
webberwang, line 58: Why can't this return `TypeBase`?

libs/backend/adapter/neo4j/src/schema/type/type.schema.ts
```diff
@@ -74,6 +84,40 @@ export const typeSchema = gql`
...
+ } 
+
+  type TypesPageArrayType implements TypeBase {
+   id: ID!
+   kind: TypeKind! 
+   name: String! @unique
+   owner: User!
+ } 
+
+ union TypesPageAnyType = TypesPageEnumType | TypesPageUnionType | TypesPageTypeBase
```
webberwang, line 119: Even if we do use union, why can't `TypesPageEnumType` be replaced with `EnumType`?

webberwang (changes requested):

libs/backend/adapter/neo4j/src/schema/type/type.schema.ts
phmngocnghia: It can be replaced, I'll address this

PR #1874, opened 2022-09-26
fix: store and actions issues

### LINKED ISSUE #1869: State modal shouldn't open with empty edit
[LOOM DEMO](https://www.loom.com/share/7dbd19d7e7854a82a870c35c12f3b00c)

assimhabeek:
## Related Issue(s)

Fixes #1869


libs/frontend/modules/store/src/store/apis/api.utils.ts
```diff
@@ -65,6 +65,9 @@ export const makeActionCreateInput = (
             ApiAction: {
               connect: { where: { node: { id: action.errorActionId } } },
```
webberwang, line 66: we can use `ApiAction: connectNode(action.errorActionId)` helper

libs/frontend/modules/store/src/store/models/store.model.ts
```diff
@@ -34,17 +30,31 @@ export class Store
...
   @modelAction
-  writeCache({ id, name, actions, api }: IStoreDTO) {
+  updateState(data: IPropData) {
+    this._state = merge(this._state, data)
+  }
+
+  @computed
+  get state() {
+    const state: IPropData = merge(
+      { ...this._api.defaults },
```
webberwang, line 46: don't have to spread here right?

libs/frontend/modules/store/src/store/models/store.model.ts
```diff
@@ -58,28 +68,21 @@ export class Store
...
-      this._api.defaults,
-      this._storeActions,
-    )
-
-    for (const key of keys(this._storeActions)) {
-      state[key].run = createActionFn(state[key].action, state)
-    }
-
-    return state
+  get _actionsRunners() {
```
webberwang, line 78: I think making private is good here, and we could remove from the `IStore` interface, this way the interfaces are only public methods which simplifies it. same for all classes when applicable

webberwang (changes requested):

libs/frontend/modules/store/src/store/models/store.model.ts
assimhabeek: we do have to spread to clone or else the error will occur 
[image](https://user-images.githubusercontent.com/33001357/193041235-50bb4fba-d371-4753-81d2-eb0ff2ed8499.png)

PR #1881, opened 2022-09-30
feat: resolver get base types

### LINKED ISSUE #1820: Add pagination to types page using Option 2

phmngocnghia:
## Description 
- TypeBase should be a type, implement ITypeBase (should be created) because we can't return interface structure
- Other types implement ITypebase too

```graphql
  input GetBaseTypesOptions {
    limit: Int
    offset: Int
  }

  type GetBaseTypesReturn {
    items: [BaseType!]!
    totalCount: Int!
  }

query {
   baseTypes(
    options: GetBaseTypesOptions
   ):GetBaseTypesReturn!
}
```

```graphql
query ($options: GetBaseTypesOptions) {
	baseTypes(options: $options) {
		totalCount
		items {
	    id
			name
			kind
			owner {
				id
			}
		}
	}
}
```

## Related Issue(s)

Fixes #1820

### DIFF
--- a/libs/backend/application/src/repositories/type/type.repo.ts
+++ b/libs/backend/application/src/repositories/type/type.repo.ts
@@ -1,11 +1,9 @@
 import { getBaseTypes } from '@codelab/backend/infra/adapter/neo4j'
 import {
-  BaseType,
   GetBaseTypesReturn,
   QueryBaseTypesArgs,
 } from '@codelab/shared/abstract/codegen'
-import { int, Record, Transaction } from 'neo4j-driver'
-import { GetBaseTypesRecord } from './types'
+import { int, Transaction } from 'neo4j-driver'
 
 export const typeRepository = {
   baseTypes: async (
@@ -23,32 +21,15 @@ export const typeRepository = {
     const totalCountRecord = getTypesRecords[0]?.get('totalCount')
     const totalCount = totalCountRecord ? int(totalCountRecord).toNumber() : 0
 
-    const withOwner = (
-      data: BaseType,
-      record: Record<GetBaseTypesRecord>,
-    ): BaseType => {
+    const items = getTypesRecords.map((record) => {
+      const type = record.get('type').properties
       const owner = record.get('owner').properties
 
       return {
-        ...data,
-        owner,
-      }
-    }
-
-    const dataMapper = (record: Record<GetBaseTypesRecord>): BaseType => {
-      const type = record.get('type').properties
-
-      return {
         ...type,
+        owner,
         __typename: 'BaseType',
       }
-    }
-
-    const items = getTypesRecords.map((record) => {
-      return withOwner(
-        dataMapper(record as Record<GetBaseTypesRecord>),
-        record as Record<GetBaseTypesRecord>,
-      )
     })
 
     return {
--- a/libs/backend/application/src/repositories/type/type.repo.ts
+++ b/libs/backend/application/src/repositories/type/type.repo.ts
@@ -11,11 +11,12 @@ export const typeRepository = {
     params: QueryBaseTypesArgs,
   ): Promise<GetBaseTypesReturn> => {
     const { options } = params
-    const { limit = 10, offset = 0 } = options || {}
+    const limit = options?.limit ?? 10
+    const offset = options?.offset ?? 0
 
     const { records: getTypesRecords } = await txn.run(getBaseTypes, {
-      limit: int(Number(limit)),
-      skip: int(Number(offset)),
+      limit: int(limit),
+      skip: int(offset),
     })
 
     const totalCountRecord = getTypesRecords[0]?.get('totalCount')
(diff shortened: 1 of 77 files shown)


libs/shared/abstract/core/src/domain/type/fragments/array-type.fragment.graphql
```diff
@@ -1,7 +1,7 @@
 fragment ArrayType on ArrayType {
   ...TypeBase
```
webberwang, line 2: `BaseType`?

webberwang (changes requested):

libs/backend/application/src/repositories/type/type.repo.ts
```diff
@@ -0,0 +1,64 @@
...
+      skip: int(Number(offset)),
+    })
+
+    const totalCountNeo4j = countTypesRecords[0].get('count(type)')
+    const totalCount = int(totalCountNeo4j).toNumber()
+
+    return totalCount
+  },
+
+  baseTypes: async (
```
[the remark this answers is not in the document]
phmngocnghia: image

OGM doesn't work with the interface. ^ won't work

I query using custom `Type` label

libs/backend/application/src/repositories/type/type.repo.ts
```diff
@@ -0,0 +1,64 @@
+import { countBaseTypes, getBaseTypes } from '@codelab/backend/adapter/neo4j'
+import { BaseType } from '@codelab/shared/abstract/codegen'
+import { int, Record, Transaction } from 'neo4j-driver'
+import { GetBaseTypesRecord } from './types'
+
+export const typeRepository = {
+  countBaseTypes: async (txn: Transaction, offset = 0): Promise<number> => {
```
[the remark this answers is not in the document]
phmngocnghia: https://github.com/codelab-app/builder/pull/1881#discussion_r985037505

libs/backend/application/src/repositories/type/type.repo.ts
```diff
@@ -0,0 +1,68 @@
...
+
+    const totalCountNeo4j = countTypesRecords[0].get('count(type)')
+    const totalCount = int(totalCountNeo4j).toNumber()
+
+    return totalCount
+  },
+
+  baseTypes: async (
+    txn: Transaction,
+    params: any,
```
webberwang, line 24: type it with generated code

libs/backend/application/src/repositories/type/type.repo.ts
```diff
@@ -0,0 +1,68 @@
+import {
+  countBaseTypes,
+  getBaseTypes,
+} from '@codelab/backend/infra/adapter/neo4j'
+import { int, Record, Transaction } from 'neo4j-driver'
+import { GetBaseTypesRecord } from './types'
+
+type BaseType = any
```
webberwang, line 8: shouldn't use any type

libs/backend/application/src/repositories/type/type.repo.ts
```diff
@@ -0,0 +1,68 @@
...
+
+      return {
+        ...type,
+        __typename: 'BaseType',
+      }
+    }
+
+    const items = getTypesRecords.flatMap((record) => {
+      return withOwner(
+        dataMapper(record as Record<GetBaseTypesRecord>),
```
webberwang, line 61: avoid casting if possible

libs/backend/graphql/src/resolvers/type/query/baseTypes.resolver.ts
```diff
@@ -0,0 +1,18 @@
+// rename
+import { typeRepository } from '@codelab/backend/application'
+import { QueryBaseTypesArgs } from '@codelab/shared/abstract/codegen'
+import { IFieldResolver } from '@graphql-tools/utils'
+import { Transaction } from 'neo4j-driver'
+
+export const baseTypes: IFieldResolver<any, any, QueryBaseTypesArgs> =
```
webberwang, line 7: don't use any, we have generated types

webberwang (changes requested):

libs/backend/application/src/repositories/type/type.repo.ts
phmngocnghia: cast is needed because the type definition of the library doesn't have a way to define the generic type of running transaction

webberwang: check out our other resolvers, we don't use any casting there

libs/backend/application/src/repositories/type/type.repo.ts
webberwang: don't use any here

libs/backend/application/src/repositories/type/type.repo.ts
webberwang: don't use any here

libs/backend/graphql/src/resolvers/type/query/baseTypes.resolver.ts
webberwang: don't use any here

webberwang (changes requested):

PR #1904, opened 2022-10-06
feat: paginate type page frontend

### LINKED ISSUE #1820: Add pagination to types page using Option 2

phmngocnghia:
## Related Issue(s)

Fixes #1820


libs/frontend/abstract/core/src/domain/type/type.service.interface.ts
```diff
@@ -70,4 +69,12 @@ export interface ITypeService
...
   getAllWithDescendants(ids: Array<string>): Promise<Array<IAnyType>>
   load(types: GetTypesQuery): Array<IAnyType>
+
+  getTypesTableCurrentPage: number
+  getTypesTablePageSize: number
+  getTypesTableTypeIds: Array<string>
+  getTypesTableTypes: Array<IAnyType>
+  getTypesTableTotalCount: number
+  queryGetTypesTableTypes(page: number, pageSize: number): Promise<void>
+  isTypeDepdenciesLoading?: boolean
```
webberwang, line 79: `isTypeDepdenciesLoading` why is this needed?

we should have a single object for getTypes instead of this many, try to mirror the api schema

webberwang (changes requested):

Issue #1915, opened 2022-10-08
Error after deleting an internal element


melhamin:
## Current Behavior
An error is thrown after deleting the first child of an element. Deleting other children work fine. 

## Steps to Reproduce

https://user-images.githubusercontent.com/48331678/194724165-d01f8940-139e-4075-867a-2708cbbbb2a0.mov

mohasarc:
I've encountered the same issue. It doesn't have to be the first element - deleting any internal element (there is another element linked to it) cause the issue.

PR #1919, opened 2022-10-09
fix: element duplication, conversion into component, and moving between components

### LINKED ISSUE #1913: Element duplication, conversion into component, and moving between components not working
## Current Behavior

1. When an element is drag/dropped from one component to another, only a referance to the original element is created (the element is not duplicated)
2. Converting an element into a component doesn't work
3. Dupplicating an element doesn't work

## Expected Behavior

1. one of the following options: 
a. A whole new copy of the element should be created.
b. Do not allow the operation.
c. Detach the element from the first component and attach it to the second. (seems the best option)

## Steps to Reproduce

1.

https://user-images.githubusercontent.com/51242349/194549039-90714a1a-367b-4349-a94c-f35827141d34.mp4

2.    

https://user-images.githubusercontent.com/51242349/194748477-9255ec16-f15d-42a3-ac57-494cbf8729e2.mp4

### LINKED ISSUE #1918: Apps are not rendered after deleting an element that uses a component as prop
## Current Behavior
After deleting an element that uses a component in its props, the page and the whole app is not rendered anymore. The error is thrown even for the newly created apps. 

## Steps to Reproduce

https://user-images.githubusercontent.com/48331678/194728190-22b6d7c5-5b9a-4b5b-9711-d9716d2ebfd5.mov

mohasarc:
## Description (Optional)

TODO: 

- [x] fix element duplication regression.
- [x] fix moving elements between components.
- [x] fix converting elements to components.
- [x] fix lint errors
- [x] fix rendering error thrown when converting `first child element` to component.
- [x] ~~fix error thrown when deleting an element that is an instance of a component (created using `convert to component`)~~ known issue #1915 

## Related Issue(s)

Fixes #1913
Fixes #1918 based on https://github.com/codelab-app/builder/issues/1918#issuecomment-1275070169

### DIFF
--- a/libs/frontend/domain/builder/src/hooks/useElementTreeDrop/useElementTreeDrop.ts
+++ b/libs/frontend/domain/builder/src/hooks/useElementTreeDrop/useElementTreeDrop.ts
@@ -3,7 +3,6 @@ import {
   IElementService,
   IElementTree,
 } from '@codelab/frontend/abstract/core'
-import { Element } from '@codelab/frontend/domain/element'
 import { Nullable } from '@codelab/shared/abstract/types'
 import { TreeProps } from 'antd/lib/tree'
 import {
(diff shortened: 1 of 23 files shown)


libs/frontend/domain/builder/src/hooks/useElementTreeDrop/useElementTreeDrop.ts
```diff
@@ -16,9 +21,49 @@ export interface UseElementTreeDropProps {
...
-  const handleDrop: TreeProps['onDrop'] = (info) => {
+  const handleDrop: TreeProps<IBuilderDataNode>['onDrop'] = async (info) => {
     const dragNodeId = info.dragNode.key.toString()
     const dropNodeId = info.node.key.toString()
+    // Check if the dropNode lives at a different component
+    // If so, we need to deep clone the element into the other component
+    const dropElement = elementService.element(dropNodeId)
+    const dragElement = elementService.element(dragNodeId)
+
+    if (dropElement && dragElement) {
```
webberwang, line 32: 4 level deep conditional, possible to flatten this or make it more readable?

webberwang (changes requested):

libs/frontend/domain/builder/src/hooks/useElementTreeDrop/useElementTreeDrop.ts
mohasarc: to simplify the implementation in `handleDrop`, I've added `rootId` to `IBuilderDataNode` and moved `moveElementToAnotherTree` to `element.service`

PR #1928, opened 2022-10-12
fix: filter out props keys that are not defined in atom api

### LINKED ISSUE #1758: Residual left after prop is deleted

melhamin:
## Description (Optional)

This PR includes a frontend solution for removing props that are left over in elements props after deleting it from atom's api. Replaces existing solution that updates element element props after the prop is deleted from atom's api.

- [x] Move filtering out logic to the `Prop` model
- [x] Fix props not filtered when navigating back to the `Builder` tab after deleting prop from Atom API (filtered after page refresh)

## Related Issue(s)

Fixes #1758

### DIFF
--- a/libs/frontend/domain/prop/src/store/prop.model.ts
+++ b/libs/frontend/domain/prop/src/store/prop.model.ts
@@ -1,4 +1,5 @@
-import type {
+import {
+  CUSTOM_TEXT_PROP_KEY,
   IInterfaceType,
   IProp,
   IPropData,
@@ -42,10 +43,20 @@ export class Prop
       const apiPropsMap = this.apiRef.current.fields.items
 
       const apiPropsByKey = values(apiPropsMap)
-        .map((propModel) => ({ [propModel.key]: propModel }))
+        .map((propModel) => ({ [propModel.current.key]: propModel }))
         .reduce(merge, {})
 
-      return omitBy(this.data.data, (_, key) => !apiPropsByKey[key])
+      console.log('apiRef', this.apiRef.current.fieldsList, this.data.data)
+
+      return omitBy(this.data.data, (_, key) => {
+        // CUSTOM_TEXT_PROP_KEY is a special case, it's an element prop
+        // that is not part of the api
+        if (key === CUSTOM_TEXT_PROP_KEY) {
+          return false
+        }
+
+        return !apiPropsByKey[key]
+      })
     }
 
     return { ...this.data.data }
--- a/libs/frontend/domain/prop/src/store/prop.model.ts
+++ b/libs/frontend/domain/prop/src/store/prop.model.ts
@@ -40,13 +40,11 @@ export class Prop
   @computed
   get values() {
     if (this.apiRef) {
-      const apiPropsMap = this.apiRef.current.fields.items
+      const apiPropsMap = this.apiRef.current.fields
 
       const apiPropsByKey = values(apiPropsMap)
-        .map((propModel) => ({ [propModel.current.key]: propModel }))
+        .map((propModel) => ({ [propModel.key]: propModel }))
         .reduce(merge, {})
-
-      console.log('apiRef', this.apiRef.current.fieldsList, this.data.data)
 
       return omitBy(this.data.data, (_, key) => {
         // CUSTOM_TEXT_PROP_KEY is a special case, it's an element prop
--- a/libs/frontend/domain/prop/src/store/prop.model.ts
+++ b/libs/frontend/domain/prop/src/store/prop.model.ts
@@ -1,10 +1,15 @@
 import type {
+  IInterfaceType,
   IProp,
   IPropData,
   IPropDTO,
 } from '@codelab/frontend/abstract/core'
+import { Maybe } from '@codelab/shared/abstract/types'
 import { mergeProps, propSafeStringify } from '@codelab/shared/utils'
+import merge from 'lodash/merge'
 import omit from 'lodash/omit'
+import omitBy from 'lodash/omitBy'
+import values from 'lodash/values'
 import { computed } from 'mobx'
 import {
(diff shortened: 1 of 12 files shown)


libs/frontend/domain/renderer/src/element/ElementWrapper.ts
```diff
@@ -117,4 +126,19 @@ export const ElementWrapper = observer<ElementWrapperProps>(
   },
 )
 
+// Removes keys from element props that are not defined in the atom API
+// This is to prevent rendering props left over after deleting it from the atom API
+const removeLeftOverPropsNotDefinedInAtomAPI = (element: IElement) => {
+  if (element.atom?.current.api.current) {
```
webberwang, line 132: unit specs failing, try `maybeCurrent`

libs/frontend/domain/renderer/src/element/ElementWrapper.ts
```diff
@@ -117,4 +126,19 @@ export const ElementWrapper = observer<ElementWrapperProps>(
...
+    const atomApiInterface = element.atom.current.api.current as IInterfaceType
+    const atomApiPropsMap = atomApiInterface.fields.items
+    const atomApiPropsKeys = values(atomApiPropsMap).map((field) => field.key)
+    const elementPropsMap = element.props?.data.data
+    const elementPropsKeys = keys(elementPropsMap)
+    const keysToOmit = difference(elementPropsKeys, atomApiPropsKeys)
+
+    keysToOmit.forEach((key) => element.props?.delete(key))
```
webberwang, line 140: I'm thinking the prop model should have a reference to an interface type? This way we can move all the logic there

That would make sense domain wise, since dats shape is dependent on the interface

webberwang (changes requested):

libs/frontend/domain/renderer/src/element/ElementWrapper.ts
melhamin: Yeah it makes sense, if we have the interface type in prop model. Will try that

libs/frontend/domain/prop/src/store/prop.model.ts
```diff
@@ -1,9 +1,13 @@
 import type {
+  IInterfaceType,
   IProp,
   IPropData,
   IPropDTO,
 } from '@codelab/frontend/abstract/core'
+import { Maybe } from '@codelab/shared/abstract/types'
 import { mergeProps, propSafeStringify } from '@codelab/shared/utils'
+// eslint-disable-next-line lodash/import-scope
+import { difference, keys, values } from 'lodash'
```
webberwang, line 10: here we do want to import using `lodash/*` etc

libs/frontend/domain/renderer/src/element/ElementWrapper.ts
```diff
@@ -3,6 +3,7 @@ import { CUSTOM_TEXT_PROP_KEY } from '@codelab/frontend/domain/element'
 import { Nullish } from '@codelab/shared/abstract/types'
 import { mergeProps } from '@codelab/shared/utils'
 import { jsx } from '@emotion/react'
+// eslint-disable-next-line lodash/import-scope
```
webberwang, line 6: remove this

libs/frontend/domain/prop/src/store/prop.model.ts
```diff
@@ -14,23 +18,40 @@ import {
...
   get values() {
+    // Filter out keys inside props that are not defined in the atom API
+    // This is to prevent rendering props left over after deleting them from the atom API
+    if (this.api) {
+      const apiPropsMap = this.api.current.fields.items
+      const apiPropsKeys = values(apiPropsMap).map((field) => field.key)
+      const currentPropsKeys = keys(this.data.data)
+      const keysToOmit = difference(currentPropsKeys, apiPropsKeys)
+
+      keysToOmit.forEach((key) => this.delete(key))
```
webberwang, line 52: add https://typescript-eslint.io/rules/no-dynamic-delete/ to root eslint, let's use a map reduce strategy

https://www.npmjs.com/package/omit-deep-lodash

Mouse_Highlight_Overlay

webberwang (changes requested):

libs/frontend/domain/prop/src/store/prop.model.ts
melhamin: Why do we need to add no-dynamic-delete rule?

webberwang: don't think it's good practice as it's mutable
no-dynamic-delete___TypeScript_ESLint

melhamin: We are not using dynamic delete here

webberwang: oh nevermind, also it's already added

melhamin: Should still use https://www.npmjs.com/package/omit-deep-lodash? both would be the same, just one extra package dependency.

webberwang: Could try this https://stackoverflow.com/a/53398058/2159920

melhamin: I want to use [omitBy](https://lodash.com/docs/4.17.15#omitBy) instead. Seem better and cleaner solution than omit.

PR #1985, opened 2022-11-08
refactor: refactor App provider tree to use Page model (#1785)

### LINKED ISSUE #1785: Refactor App provider tree to use Page model
I got started on this before but isn't done yet

## Description

Makes more sense to model an App provider tree using Page.

## Use Case

In Next.js `_app.tsx` is basically a page.

## Suggested Implementation

Add `Page.isProvider`

## Additional Tasks

- [x] Provide a screen recording showing that provider works using Ant Design config provider
- [x] Create a Cypress e2e spec automating the process above

vladyslav-polishchuk:
## Description (Optional)

This #1785 task requires at least these steps to be done:
1. ✅ Create `_app` page automatically when app is created in the builder
2. ✅ Add `isProvider` property to the page schema so that we can programmatically distinguish it from others
3. ✅ Update autogenerated backend endpoints to return new "isProvider" field to client
4. ✅ Disable the ability to rename or delete this page
5. ✅ Pass the page to a renderer so that it is used as a wrapper over all other pages.
    Looks like there are 3 places where it is used: 
        `apps/builder/pages/_sites/user/[user]/[app]/[page]/index.tsx`
        `apps/builder/pages/apps/[appId]/pages/[pageId]/builder.tsx`
        `apps/builder/pages/apps/[appId]/pages/[pageId]/index.tsx`
6. ✅ Probably some refactoring to reduce copy-pasted code (there is big piece of code that looks the same in the files above)

## Related Issue(s)

Fixes #1785

### DIFF
--- a/apps/builder-e2e/src/e2e/admin/assert.ts
+++ b/apps/builder-e2e/src/e2e/admin/assert.ts
@@ -5,11 +5,13 @@ import { AntdTag } from 'libs/shared/data/src/tag/antd-tags.data'
 import difference from 'lodash/difference'
 import * as path from 'path'
 
+const TIMEOUT = 140000
+
 export const seedData = () => {
   cy.log('yarn cli seed').exec(
     'yarn cli data seed --stage test --email cypress@codelab.app',
     {
-      timeout: 90000,
+      timeout: TIMEOUT,
     },
   )
   // .its('stdout')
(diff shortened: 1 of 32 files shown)


libs/frontend/presenter/container/src/pageHooks/usePageRenderingData.ts
```diff
@@ -0,0 +1,87 @@
+import { useAsync } from 'react-use'
+import { useStore } from '../providers'
+import { useCurrentAppId, useCurrentPageId } from '../routerHooks'
+
+export const usePageRenderingData = () => {
```
phmngocnghia, line 5: I found this name a little bit confusing. I can suggest something like `useLoadRenderedPage`

PR #1993, opened 2022-11-09
fix: conflicting transactions

### LINKED ISSUE #1969: Elements disapear sometimes when drag dropped
## Current Behavior

When drap dropping an element into a location between itself and the next element, it disappears. Additionally, the entire page freezes.

## Expected Behavior
No elements should be lost when drag dropped.

## Steps to Reproduce

https://user-images.githubusercontent.com/51242349/199079318-d15008ce-1169-473e-94e7-e584fd414b48.mp4

mohasarc:
## Description (Optional)

This PR introduces a way to run transactions that are potentially conflicting is a sequential manner.

To do that, I've introduced a function called `runSequentially`, which, given a generator function and a `key`, will run this function after waiting for other functions that were run before the current one on the same `key`.

https://user-images.githubusercontent.com/51242349/200827003-97993a54-26c5-44e4-8c20-a0d4a7cd36d6.mp4

## Related Issue(s)

Fixes #1969


webberwang:
Where would the code be that uses this function?

mohasarc:
> Where would the code be that uses this function?

It wraps the functions that we want to run sequentially, I've used it in `ElementService` for `moveElementAsNextSibling` and some other potentially problematic functions.

PR #1999, opened 2022-11-10
fix: button href default value issue

mehmetsefabalik:
## Related Issue(s)

Fixes #


webberwang:
We should opt for the seeder default value method

libs/frontend/domain/element/src/use-cases/element/update-element-props/UpdateElementPropsForm.tsx
```diff
@@ -52,13 +52,21 @@ export const UpdateElementPropsForm = observer<UpdateElementPropsFormProps>(
     )
 
     const onSubmit = (data: IPropData) => {
-      console.log('Submitting: ', data)
+      const updateData = JSON.parse(JSON.stringify(data))
```
webberwang, line 55: why do you parse then stringify here?

PR #2013, opened 2022-11-14
feat: screen size preview for builder page

### LINKED ISSUE #1981: Screen size preview for builder page
Add 5 icons at the top bar for:

- mobile
- tablet vertical
- tablet horizontal
- desktop
- custom

Clicking on each should resize to the preset width (you can choose the 4 breakpoint values, use webflow as guideline)

Allow resize (but we won't have any snapping to breakpoints like webflow), once we resize, the custom icon will be active. Clicking on the active icon should bring up a integer input box which has the custom value.

https://user-images.githubusercontent.com/26264925/200336047-c447baa5-faa7-4934-b003-2e349e6dcd49.mov

tinhw3:
## Related Issue(s)

Fixes #1981

### DIFF
--- a/libs/frontend/domain/builder/src/hooks/useBuilderResize.ts
+++ b/libs/frontend/domain/builder/src/hooks/useBuilderResize.ts
@@ -1,0 +1,132 @@
+import { IBuilderService } from '@codelab/frontend/abstract/core'
+import {
+  MotionProps,
+  MotionValue,
+  PanInfo,
+  useMotionValue,
+} from 'framer-motion'
+import { useCallback, useEffect, useState } from 'react'
+
+export interface MinMaxValue {
+  min?: number
+  max?: number
+  default?: number
+}
+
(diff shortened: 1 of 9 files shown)


libs/frontend/domain/builder/src/hooks/useBuilderResize.ts
```diff
@@ -0,0 +1,129 @@
...
+    (event: MouseEvent | TouchEvent | PointerEvent, info: PanInfo) => {
+      setResizingMainContent(true)
+      clampSet(mWidth, info.delta.x, width)
+      setMainResizingContentWidth(Math.round(mWidth.get()))
+    },
+    [mWidth, width],
+  )
+
+  useEffect(() => {
+    if (width?.default && width.max && width.min) {
```
webberwang, line 81: more clear if we do not or and return early

webberwang (changes requested):

PR #2022, opened 2022-11-16
fix: select component issue when inside scrollable container

### LINKED ISSUE #1983: Dropdown menus do not stick to container on page scroll
## Current Behavior

On builder page (most probably in other places as well) dropdown menu which is displayed below or under select field does not stick to it when page is scrolled

## Expected Behavior

Should either stick or close

## Steps to Reproduce

https://user-images.githubusercontent.com/74900868/200548495-a3f7af2a-0382-45c9-8aef-3cbfd6aeb860.mov

vladyslav-polishchuk:
## Description

There is an issue with `antd` Select component when it is rendered inside a scrollable container.
In official documentation for this component (https://ant.design/components/select/) as a valid solution suggested to customize `getPopupContainer` prop which defines where the dropdown menu will be rendered (default is body).
As a fix - the dropdown menu should render next to the select component in DOM.

This fix was applied to all places where there was a chance to get this issue.

Note: It is possible to define `getPopupContainer` method on `ConfigProvider` node inside of `_app.tsx`, but this does not work for us. Most of the Select components are rendered in forms using `teraform-antd`. This package uses the Select component from `antd/lib/select` location, and this component does not take into account settings from `ConfigProvider`. Only Select components imported from root `antd` inherit settings defined on `ConfigProvider`. That is why the fix was duplicated to all possible places instead of defining it once on a `ConfigProvider`

Fixes #1983

### DIFF
--- a/libs/frontend/domain/type/src/interface-form/fields/select-action/SelectAction.tsx
+++ b/libs/frontend/domain/type/src/interface-form/fields/select-action/SelectAction.tsx
@@ -43,6 +43,7 @@ export const SelectAction = ({ name, label, error }: SelectActionProps) => {
   return (
     <SelectField
       error={error || queryError}
+      getPopupContainer={(triggerNode) => triggerNode.parentElement}
       label={label}
       loading={isLoading}
       name={name}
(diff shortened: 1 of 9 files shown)


webberwang:
good catch thanks

PR #2032, opened 2022-11-18
test(apps-builder-e2e): stabilize tags.cy e2e test (#2015)

### LINKED ISSUE #2015: Stabilize "tag.cy.ts" spec
## Current Implementation Concern

The current implementation of the spec is unstable. Sometimes (~2-3 times out of 10 runs) the test fails in both environments: local and CI. Failed the same step with the same error: 
Снимок экрана 2022-11-15 в 13 04 49

vladyslav-polishchuk:
## Description (Optional)

Rework names of the tags that are used in the tags.cy.ts spec. The failing test was trying to delete the tag with the name "Delete-Tag-0", but tags with the names "Delete-Tag-0-0" and "Delete-Tag-0-1" also matched the "contains" condition. In the "before" hook, depending on what tag is created first - determined final tags order in the grid. And sometimes the wrong tag gets deleted ("Delete-Tag-0-1" instead of "Delete-tag-0") and the assertion fails. With new names - there is no way some different tag will match the selector

## Related Issue(s)

Fixes #2015

### DIFF
--- a/apps/builder-e2e/src/data/tag.ts
+++ b/apps/builder-e2e/src/data/tag.ts
@@ -1,25 +1,25 @@
 export const createData = {
   // Parent
-  tag_0: 'Create-Tag-0',
+  tag_0: 'Create-Tag-0.',
   // Child
-  tag_0_0: 'Create-Tag-0-0',
+  tag_0_0: 'Create-Tag-0-0.',
 }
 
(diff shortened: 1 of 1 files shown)


webberwang:
good catch that makes sense

PR #2044, opened 2022-11-23
fix: default state name with key

### LINKED ISSUE #2040: Default state name with key
image

tinhw3:
## Description (Optional)

https://user-images.githubusercontent.com/117329462/203505173-b4f26fdd-fb69-4b41-a65a-23edf88ed997.mov

## Related Issue(s)

Fixes #2040


libs/frontend/domain/store/src/use-cases/state/get-state/GetStateItem.tsx
```diff
@@ -29,7 +29,7 @@ export const GetStateItem = observer<StateTreeItemProp>(
       <List.Item style={{ padding: 8 }}>
         <Space style={{ width: '100%' }}>
           <FileOutlined />
-          {field.name}
+          {field.name ? field.name : field.key}
```
phmngocnghia, line 32: nit: we can rewrite this as `field.name || field.key` to make the statement even shorter, not sure if this would lead to easier to read

PR #2051, opened 2022-11-25
feat: debouncer for text props

### LINKED ISSUE #2034: Props panel debouncer for text props
Currently, QraphQL request will be sent each time a new symbol is entered into a text prop field.
For example, if the button is added to the page, then when entering the button label - requests are sent for each new letter:

Снимок экрана 2022-11-18 в 13 39 02

Either a debouncer may be added (which will wait ~100-200 milliseconds of user idle and only then will send the request), or send it on focusout.

tinhw3:
## Description (Optional)

https://user-images.githubusercontent.com/117329462/203890293-52fb5c78-aa23-4af1-a5db-c64d7e8ea1d2.mov

> Local update instantly while api delayed

https://user-images.githubusercontent.com/117329462/204436371-bd63012c-3877-404c-89bb-1c9cae0930ff.mov

## Related Issue(s)

Fixes #2034


webberwang:
for the debounce is it possible to have the local update instantly while the graphql api delayed?

PR #2073, opened 2022-12-04
fix atom table filter (server side filtering)

### LINKED ISSUE #1991: Filtering atoms return empty table
https://user-images.githubusercontent.com/33001357/200635692-96090af1-98c1-49cb-9cac-aeb4bc39b9c3.mp4

mohasarc:
## Description (Optional)

TODO:
- [x] Fix custom atom resolver not accepting where arguments
- [x] Send a query with name_CONTAINS argument each time the filter is changed and update local cache.
- [x] Fix pagination (currently always shows the number of pages based on all atoms regardless of the filer). We should see the number of pages based on how many items available given the provided filter.
- [x] Make the search case-insensitive?

## Related Issue(s)
Fixes #1991

### DIFF
--- a/libs/backend/application/src/repositories/atom/atom.repo.ts
+++ b/libs/backend/application/src/repositories/atom/atom.repo.ts
@@ -1,42 +1,41 @@
 import {
-  atoms,
   atomSelectionSet,
-  atomsWithLimit,
   Repository,
   tagSelectionSet,
 } from '@codelab/backend/infra/adapter/neo4j'
 import { Atom, GetAtomsQueryVariables } from '@codelab/shared/abstract/codegen'
-import { int, Transaction } from 'neo4j-driver'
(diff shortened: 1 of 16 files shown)


libs/backend/application/src/repositories/atom/atom.repo.ts
```diff
@@ -1,42 +1,41 @@
...
 
-    /**
-     * We can still use the same query, but we get ID from context instead
-     */
-    const { records: atomsRecords } = await txn.run(cypher, {
-      limit: int(limit),
-      skip: int(offset),
+    const filteredItemIds = await AtomInstance.find({
+      // eslint-disable-next-line @typescript-eslint/no-explicit-any
+      where: params.where as any,
```
webberwang, line 25: builder_–_atom_repo_ts

would this be better for typesafety?

PR #2084, opened 2022-12-06
fix(libs-frontend): set the highest precedence to gui css rules (#2048)

### LINKED ISSUE #2048: Can't change size of ant design input
## Current Behavior
- Create input element on a page
- Go to `CSS` tab
- Change the width of the input
- New CSS applied by the layout editor doesn't take precedence against the base style of ant design input

## Expected behavior
- Ant design atom should be tweaked to make its width easier to override

## Video Showing Bug
image

vladyslav-polishchuk:
## Description
Since we support lots of components in a different UI frameworks (antd, mui) it is important to set higher precedence for our custom styles, so that they for sure are applied and not overriden by default styles of components

## Related Issue(s)

Fixes #2048

### DIFF
--- a/libs/frontend/domain/renderer/src/test/increaseCssPrecedence.spec.ts
+++ b/libs/frontend/domain/renderer/src/test/increaseCssPrecedence.spec.ts
@@ -1,0 +1,26 @@
+import { increaseCssPrecedence } from '../utils/increaseCssPrecedence'
+
+describe('increaseCssPrecedence', () => {
+  const testCss = {
+    display: 'flex',
+    padding: 0,
+  }
+
+  it('should increase precedence for each prop when parameter is css string', async () => {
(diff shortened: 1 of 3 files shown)


libs/frontend/domain/renderer/src/test/increaseCssPrecedence.spec.ts
```diff
@@ -0,0 +1,26 @@
+import { increaseCssPrecedence } from '../utils/increaseCssPrecedence'
+
+describe('increaseCssPrecedence', () => {
```
webberwang, line 3: maybe something like `prependCSSImportance`?

webberwang:
Could we import user stylesheets before the ui framework so it has higher precedence that way?

webberwang (changes requested):

PR #2087, opened 2022-12-07
feat: adding auth to curd app

### LINKED ISSUE #1963: Crud app with supabase
- [x] #1986
- [ ] #1987
- [x] #1988
- [x] #1989
- [x] #1990
- [x] #1991
- [x] #2017
- [x] #1996
- [x] #2018
- [x] #2062
- [x] #2063
- [ ] #2064

assimhabeek:
## Description (Optional)

- Inject `getServerSidePropsData` into state
- Fix `ApiAction` import (add missing success/error actions connections)
- Fix `Store` import because `app` is required in store and vice-versa. store and app must be created in one query.
- Replace `renderIfPropKey` with `renderIfExpression` which uses state expressions 
- Fix `Element` import, missing properties `firstChild`,`nextSibling`,`prevSibling`,`parent`
- Add `stop` to `IRenderOutput` which prevents rendering children if `renderIfExpression = false` in parent element
- Restore `forwardedRef` even-though it causes warnings in console, because direct `ref` doesn't work for loadable components  
- Rewrite `ConditionalRenderPipe` to use state expressions
- Move custom text injection from `ElementWrapper` to `render.model` inside renderChildren
- Modify unit tests to work with current updates
 

## Related Issue(s)

Fixes #1963
#2119
#2112

### DIFF
--- a/apps/builder/pages/apps/[appId]/pages/[pageId]/builder.tsx
+++ b/apps/builder/pages/apps/[appId]/pages/[pageId]/builder.tsx
@@ -76,7 +76,10 @@ const PageBuilder: CodelabPage<IPageProps> = observer(
       })
 
       appStore.state.setMany(appService.appsJson)
-      appStore.state.setMany(getServerSidePropsData)
+
+      if (getServerSidePropsData) {
+        appStore.state.setMany(getServerSidePropsData)
+      }
 
       return {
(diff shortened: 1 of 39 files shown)


webberwang:
Thanks for the summary helps a lot

PR #2120, opened 2022-12-21
refactor: improve performance (#2101)

### LINKED ISSUE #2101: Optimize page builder loading

1. GraphQL waterfall calls

Some places may have waterfall API calls that chain together to take a while. We could combine those queries at the top level if they don't depend on each other. You can use console inspector to check those calls.

If one GraphQL call depends on a previous, we may consider creating a new GraphQL query so we only require a single call.

2. Caching components

We could optimize our components with React hooks such as `useMemo`, & `useCallback`, `React.lazy` etc.

`useMemo` could be good for complex components that have many interactions that is causing it to re-render
`useCallback` for complex functions
`React.lazy` for reducing bundle size

3. Check why components render

We could enable https://github.com/welldone-software/why-did-you-render by un-commnenting the import to see why components are re-rendering. This is generally a laborious process of going through each component, so we will focus on complex components that are more render intensive.

builder_–__app_tsx

---

A use case of the above is changing between pages. It's quite fast in UIBakery

https://user-images.githubusercontent.com/26264925/206811138-1bea709d-9935-4b7d-9e64-34419c579acf.mov

Ours is a tad slower, actually not that bad after testing it. Issue is mostly 

- #2083

https://user-images.githubusercontent.com/26264925/206811536-307db6b7-a1b9-4414-84ec-7916ca143cc5.mov

vladyslav-polishchuk:
## Description
Note: Recommended to review per-commit

Before | After
Screenshot 2022-12-25 at 16 22 25|
Screenshot 2022-12-25 at 16 21 39|

## Related Issue(s)

Fixes #2101


webberwang:
Is there an after photo?

apps/builder-e2e/src/e2e/components-tab.cy.ts
```diff
@@ -7,8 +7,12 @@ import { v4 } from 'uuid'
 import { createAppInput } from '../support/database/app'
 import { createPageInput } from '../support/database/page'
 
-const COMPONENTS_TAB_ID = 'rc-tabs-1-tab-component'
-const COMPONENTS_TAB_PANEL_ID = 'rc-tabs-1-panel-component'
+// antd tabs use internal id counter and tab "id" is computed based on initialization order.
+// so the ids of tabs may change depending on the order in which they are rendered:
+// rc-tabs-0-tab, rc-tabs-1-tab, rc-tabs-2-tab, etc...
+// so the id below may change anytime when components rendering order is changed.
+const COMPONENTS_TAB_ID = 'rc-tabs-4-tab-component'
```
webberwang, line 14: It's odd that antd doesn't use the tab pane key to compute the id

libs/frontend/domain/builder/src/sections/config-pane/ConfigPane-InspectorTabContainer/ConfigPane-InspectorTabContainer.tsx
```diff
@@ -92,12 +92,12 @@ export const ConfigPaneInspectorTabContainer = observer<MetaPaneBuilderProps>(
     const trackPromises = useTrackLoadingPromises()
 
     if (!selectedNode) {
-      return null
+      return <Spin />
```
webberwang, line 95: This seems relevant https://beta.nextjs.org/docs/routing/loading-ui and could possible add as a separate issue in the future.

libs/frontend/presenter/container/src/pageHooks/useLoadRenderedPage.ts
```diff
@@ -45,7 +46,29 @@ export const useLoadRenderedPage = () => {
     const { pageElementTree: pageTree, page } = currentPage
     const { pageElementTree: appTree } = providerPage ?? {}
 
-    typeService.load({ ...types })
+    // type loading is quiet a heavy operation which takes up to 500ms of blocking time.
+    // split types loading into many chunks and queue each of them as a macrotask.
+    // this will unblock UI and allow other js to execute between them, which makes UI much more responsive.
+    await new Promise((resolve) =>
```
webberwang, line 52: This seems a much needed improvement

libs/frontend/domain/builder/src/sections/config-pane/ConfigPane-InspectorTabContainer/ConfigPane-InspectorTabContainer.tsx
vladyslav-polishchuk: I will create an issue for that so that we do not forget to check if we can benefit from this. Thanks

vladyslav-polishchuk: https://github.com/codelab-app/builder/issues/2127

Issue #2064, opened 2022-11-29
Builder could not resolve an object ElementTreeRef 


assimhabeek:
## Current Behavior
Reloading builder **sometimes** produce the following error 
## Video Showing Bug
![image](https://user-images.githubusercontent.com/33001357/204542372-fe35d9b4-7517-4ac4-a980-20ccb4320879.png)

assimhabeek:
I don't believe https://github.com/codelab-app/builder/pull/2128#issuecomment-1365850093

assimhabeek:
can't reproduce error anymore. mostly gone with builder layout refactor

assimhabeek:
This issue is hard to reproduce, but I still face it on preview mode

PR #2128, opened 2022-12-27
fix(apps): fix script error (#2101)

vladyslav-polishchuk:
## Description (Optional)
because of more async UI loading - some parts of data may not be available yet and cause script error during initialization


assimhabeek:
@vladyslav-polishchuk is this fixing #2064

vladyslav-polishchuk:
> #2064



> @vladyslav-polishchuk is this fixing #2064 ?

I don't think so. This fixes the regression after my Pr, which was merged today

PR #2132, opened 2023-01-04
fix(libs): fix empty type and kind column on types table (#2117)

### LINKED ISSUE #2090: Error editing type
## Current Behavior
Editing `Type` field on a type throws error.

## Video Showing Bug

https://user-images.githubusercontent.com/48331678/206287591-675bec6e-bb1e-4078-b555-ca2fca020c41.mov

### LINKED ISSUE #2117: Types table - nested table's `Type` and `Kind` fields are empty
## Current Behavior

Although the values exist they are not visible under `Type` and `Kind` columns

## Video Showing Bug
image

jethrocabaluna:
## Description (Optional)

- load sub types of current page only when expanded
- make typesList alphabetical

## Related Issue(s)

Fixes #2117 
Fixes #2090

### DIFF
--- a/libs/frontend/domain/type/src/store/type.service.ts
+++ b/libs/frontend/domain/type/src/store/type.service.ts
@@ -7,7 +7,6 @@ import type {
 import { IAnyType, ITypeDTO } from '@codelab/frontend/abstract/core'
 import { ModalService } from '@codelab/frontend/shared/utils'
 import type {
-  BaseTypeOptions,
   BaseTypeWhere,
   FieldFragment,
 } from '@codelab/shared/abstract/codegen'
@@ -98,7 +97,12 @@ export class TypeService
 
   @computed
   get typesList() {
-    return [...this.types.values()]
+    // loading sub types messes up the order of the next page
+    // we need to sort here to make sure the types on the
+    // table are always sorted alphabetically
+    return [...this.types.values()].sort((typeA, typeB) =>
+      typeA.name.toLowerCase() < typeB.name.toLowerCase() ? -1 : 1,
+    )
   }
 
   @modelAction
@@ -190,23 +194,9 @@ export class TypeService
 
   @modelFlow
   @transaction
-  getAll = _async(function* (
-    this: TypeService,
-    where?: BaseTypeWhere,
-    options?: BaseTypeOptions,
-  ) {
-    const ids = yield* _await(
-      this.getBaseTypes({
-        where: {
-          name: where?.name,
-        },
-        offset: options?.offset,
-        limit: options?.limit,
-      }),
-    )
-
-    const allIds = [...ids, ...(where?.id_IN || [])]
-    const types = yield* _await(getAllTypes(allIds))
+  getAll = _async(function* (this: TypeService, where?: BaseTypeWhere) {
+    const ids = where?.id_IN ?? undefined
+    const types = yield* _await(getAllTypes(ids))
 
     return types.map((type) => {
       return this.writeCache(type)
--- a/libs/frontend/domain/type/src/use-cases/types/get-types/useTypesTableData.ts
+++ b/libs/frontend/domain/type/src/use-cases/types/get-types/useTypesTableData.ts
@@ -2,13 +2,25 @@ import type { ITypeService } from '@codelab/frontend/abstract/core'
 import { useAsyncFn } from 'react-use'
 
 export const useTypesTableData = (typeService: ITypeService) => {
-  const [{ loading: isLoadingAllTypes }, getAllTypes] = useAsyncFn(
-    typeService.getAll.bind(typeService),
-    [],
+  typeService.getBaseTypes.bind(typeService)
+  typeService.getAll.bind(typeService)
+
+  /**
+   * Get the base types of the current page and load the fields and write to cache
+   */
+  const setupCurrentPageTypes = async (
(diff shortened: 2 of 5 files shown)


libs/frontend/domain/type/src/store/type.service.ts
```diff
@@ -98,7 +97,12 @@ export class TypeService
 
   @computed
   get typesList() {
-    return [...this.types.values()]
+    // loading sub types messes up the order of the next page
```
webberwang, line 100: how come the backend sort by name isn't enough?

libs/frontend/domain/type/src/store/type.service.ts
```diff
@@ -190,23 +194,9 @@ export class TypeService
...
-          name: where?.name,
-        },
-        offset: options?.offset,
-        limit: options?.limit,
-      }),
-    )
-
-    const allIds = [...ids, ...(where?.id_IN || [])]
-    const types = yield* _await(getAllTypes(allIds))
+  getAll = _async(function* (this: TypeService, where?: BaseTypeWhere) {
```
webberwang, line 197: I see offset is removed, but what if want offset and limit to be passed here?

libs/frontend/domain/type/src/use-cases/types/get-types/useTypesTableData.ts
```diff
@@ -2,13 +2,25 @@ import type { ITypeService } from '@codelab/frontend/abstract/core'
 import { useAsyncFn } from 'react-use'
 
 export const useTypesTableData = (typeService: ITypeService) => {
-  const [{ loading: isLoadingAllTypes }, getAllTypes] = useAsyncFn(
-    typeService.getAll.bind(typeService),
-    [],
+  typeService.getBaseTypes.bind(typeService)
+  typeService.getAll.bind(typeService)
```
webberwang, line 6: shouldn't get all already get base types?

webberwang (changes requested):

libs/frontend/domain/type/src/store/type.service.ts
jethrocabaluna: The sort by name on the cypher query works well until we get the types of the interfaceType fields. From what I understand from the `class TypeService`, it has an objectMap `types` property and we populate it whenever we get types. So in our pagination:
1. The types in the first page are saved on the `this.types`
2. If we get the types of the fields of an interfaceType, those types will also be saved in the `this.types`
3. When navigated on the second page, the field types that were saved appears before the new set of base types
4. We get the `this.types` with `[...this.types.values()]`, somehow this doesn't return them alphabetically  so I needed to add an extra sort
scrnli_1_5_2023_5-10-38 PM
scrnli_1_5_2023_5-11-51 PM

libs/frontend/domain/type/src/store/type.service.ts
jethrocabaluna: The offset and limit were only used for the `this.getBaseTypes`. I removed that block of code inside the `getAll` so theres no usage of the offset and limit anymore in the function. The `getAllTypes` only accepts a single argument which is an array of string which should be the ids.

The reason I removed the usage of `this.getBaseTypes` is because it can be unnecessary when `getAll` is used but doesn't really need the baseTypes. For example is the `getOne`, which calls `getAll` with a specific id if the type doesn't exist yet in the `this.types`. The `this.getBaseTypes` doesn't need to run in the `getOne`

libs/frontend/domain/type/src/use-cases/types/get-types/useTypesTableData.ts
jethrocabaluna: In this PR, I removed the use of `getBaseTypes` inside the `getAll` because I think it could run unnecessarily and we may use `getAll` but dont need the baseTypes like in the method [getOne](https://github.com/codelab-app/builder/blob/4916af3705dc046340c5095cdbe6593f4a589576/libs/frontend/domain/type/src/store/type.service.ts#L218).

Since they are separated now, I had to use `getBaseTypes` first to get the ids in the current page and use `getAll` with those ids to populate the necessary fields for the baseTypes (e.g. `fields` property of an interfaceType)

jethrocabaluna:
In this video, we can see that the field types are only loaded upon expanding.
@webberwang let me know if should I load the sub field types of interfaceTypes in the background instead of in-demand.
[interface-types-pagination.webm](https://user-images.githubusercontent.com/27695022/210751222-d9d5622b-c3ac-42c2-96af-2ba47ac5db1d.mp4)

jethrocabaluna:
[interface-types-pagination.webm](https://user-images.githubusercontent.com/27695022/210752106-490b16e9-c9e5-4b0d-ab71-e8a63386ec09.mp4)

jethrocabaluna:
Updated the PR so it doesn't only get the descendants of `interfaceTypes`. I moved that logic upon expanding the type.

[type-descendants.webm](https://user-images.githubusercontent.com/27695022/211058039-1ff106a7-ce55-447e-bb59-67ed89905768.webm)

libs/frontend/domain/type/src/store/type.service.ts
webberwang: okay this makes sense, I think prev dev added unnecessary logic

webberwang (changes requested):

PR #2134, opened 2023-01-05
fix: fix script error when expression field is nested in schema

vladyslav-polishchuk:
## Related Issue(s)

Fixes #


webberwang:
Closed in favor of #2137

PR #2135, opened 2023-01-05
url driven pagination

### LINKED ISSUE #2066: Make pages with paginated tables URL-driven
Pages `tags`, `atoms`, `types` (maybe some other pages as well) display tables with data that support pagination.
It is a good practice to make such pages URL-driven, so that users may refresh/bookmark/share page similar to this:
`http://127.0.0.1:3000/types?page=3&pageSize=25` and always get the same result without a need to setup pagination parameters again

mohasarc:
## Related Issue

Fixes #2066


webberwang:
Could you add a video next time showing the new feature?

PR #2138, opened 2023-01-06
feat(apps-builder): implement editing of type API in the builder page

### LINKED ISSUE #2136: Edit type API in the builder page

Instead of the link routing us to the types table just to be able to edit the props, what we could do is open up a modal where we can edit the API.

The benefits we can get doing so are:
1. UX wise, i think its better to stay in the builder page
2. We wont need to fire extra unnecessary network calls
3. Doesn't need to figure out the correct page where the type belongs to
4. [GetTypesTable](https://github.com/codelab-app/builder/blob/master/libs/frontend/domain/type/src/use-cases/types/get-types/GetTypesTable.tsx) will be simpler as we can remove the logic of figuring out the page with the given `typeId` 

The current implementation may not be ideal in the long run because the way it works now is all the types are queried firsthand (with [GetRenderedPageAndCommonAppData](https://github.com/codelab-app/builder/blob/286eb8bdcb0b85c3b2b25edbb996733e39273584/libs/frontend/domain/page/src/graphql/page.endpoints.graphql.gen.ts#L134)) when the builder page is open, and these types are saved in the cache so when the user clicks on the link to edit the type, the pagination will work since all the types are already loaded.
But what if we improve the initial load of the page so that it only requests the necessary data? Then the pagination will not work anymore since all of the types are not in the cache anymore.

[types-edit-implementation.webm](https://user-images.githubusercontent.com/27695022/210969048-6d770916-db51-48cc-b8f8-74aa323ac1ff.webm)

jethrocabaluna:
## Description (Optional)

- can now add, edit, or delete a prop on the builder page

[edit-type-on-builder.webm](https://user-images.githubusercontent.com/27695022/211051824-09d1bb73-9a9c-4365-8f80-f71f5f1b13e6.webm)

## Related Issue(s)

Fixes #2136

### DIFF
--- a/apps/builder/pages/apps/[appId]/pages/[pageId]/builder.tsx
+++ b/apps/builder/pages/apps/[appId]/pages/[pageId]/builder.tsx
@@ -138,17 +138,10 @@ PageBuilder.Layout = observer((page) => {
     () =>
       observer(() => (
         <ConfigPane
-          actionService={actionService}
-          atomService={atomService}
-          builderService={builderService}
-          componentService={componentService}
-          elementService={elementService}
           // The element tree changes depending on whether a page or a component is selected
           elementTree={activeElementTree}
           key={pageBuilderRenderer?.pageTree?.current.root?.id}
           renderService={pageBuilderRenderer}
-          typeService={typeService}
-          userService={userService}
         />
       )),
     [pageBuilderRenderer, builderService],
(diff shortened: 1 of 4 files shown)


jethrocabaluna:
My first approach was trying to add an edit and delete icon on each prop but it gets complicated since these are rendered inside the antd form and could have more overhead.
I went with the simplest solution I can think of where there are three button on the bottom of the props panel for add/edit/delete.

PR #2157, opened 2023-01-10
refactor(libs-frontend-domain-renderer): dont wrap draggables with div

### LINKED ISSUE #2125: DraggableElement div conflict with  styling
## Current Behavior
`DraggableElement` uses a `div` to wrap content which causes a problem with styling where some elements must be direct children in order for styles to work. an example for that is flexbox
right now we can't produce this simple example 
```
<div class="flex-container">
   <div>Navlink 01</div>
   <div>Navlink 02</div>
   <div style={{flex:'1 1 auto'}}> This should be a spacer between links and logout button </>
   <div>Logout button</div>
</div>

```
because it gets render as follow 

```
<div style={{ position: 'relative' }}>  
  <div class="flex-container">
      <div style={{ position: 'relative' }}>  
         <div>Navlink 01</div>
       </div>

      <div style={{ position: 'relative' }}>  
         <div>Navlink 02</div>
       </div>

      <div style={{ position: 'relative' }}>  
         <div style={{flex:'1 1 auto'}}> This should be a spacer between links and logout button </>
       </div>
    
      <div style={{ position: 'relative' }}>  
          <div>Logout button</div>
       </div>

  </div>
</div>

```

jethrocabaluna:
## Description

This PR removes the wrapping of elements to be draggable/droppable with `<div>` since this can affect styling we set for the element itself.
- make the element itself to be the nodeRef used for draggable and droppable
- make the rendered children able to receive additional props by rendering it later in the `DraggableElement` component so that the attributes and listeners can be passed as props to the children
- refactor the indicator styling so that only box shadow is used

**NOTE:** I actually dont like the use of `setTimeout` to resolve the ref. Maybe theres a better way to resolve it to be used for dnd

## Video or Image

https://user-images.githubusercontent.com/27695022/211614218-bfbcad94-291a-4f76-88ca-d394c6c966d4.mp4

## Related Issue(s)

Fixes #2125


libs/frontend/domain/renderer/src/element/DraggableElement.tsx
```diff
@@ -39,14 +44,27 @@ export const DraggableElement = ({
     over,
   } = useDroppable({ id: element.id })
 
+  useEffect(() => {
+    // Passing the setDraggableAndDroppableNodeRef as ref to makeRenderedElements
+    // doesn't work properly for some components like AntDesignImage.
+    // This works for now but could be improved later on
+    setTimeout(() => {
```
webberwang, line 51: Can't we have some custom onRenderComplete event that fires after rendering is complete to trigger this?

webberwang (changes requested):

PR #2160, opened 2023-01-11
fix: cannot dnd atom to the builder if it is the first element

### LINKED ISSUE #2106: Cannot dnd a component to the builder if it's the first element
## Current Behavior
Drag-dropping a component from the right panel doesn't work because I think it needs an element to attach to as next sbling

## Video Showing Bug

https://user-images.githubusercontent.com/51242349/207295130-3386768b-ca7f-4994-9b47-bef2a8c84f65.mp4

jethrocabaluna:
## Description

The root element needs to be droppable so that we can drop a draggable element into it.

The logic for creating a first child should also be fixed.

## Video or Image

https://user-images.githubusercontent.com/27695022/212038489-1736bbae-554e-4d5a-921d-d1e03921a881.mp4

https://user-images.githubusercontent.com/27695022/211850769-23c0e841-638d-410c-9079-4f18001878a3.mp4

## Related Issue(s)

Fixes #2106

### DIFF
--- a/libs/frontend/domain/builder/src/sections/content/Builder.tsx
+++ b/libs/frontend/domain/builder/src/sections/content/Builder.tsx
@@ -7,9 +7,14 @@ import type {
 import {
   BUILDER_CONTAINER_ID,
   DATA_ELEMENT_ID,
+  DragPosition,
 } from '@codelab/frontend/abstract/core'
 import type { RendererRoot } from '@codelab/frontend/domain/renderer'
-import { Renderer } from '@codelab/frontend/domain/renderer'
+import {
+  makeDropIndicatorStyle,
+  Renderer,
+} from '@codelab/frontend/domain/renderer'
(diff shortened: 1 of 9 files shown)


libs/frontend/domain/builder/src/sections/content/Builder.tsx
```diff
@@ -62,6 +67,13 @@ export const Builder = observer<BuilderProps>(
       setCurrentBuilderWidth,
     })
 
+    // by our current IElement, root can be undefined
+    // useDroppable does not accept an undefined id
+    // since this is the root element, id should be defined
+    const { setNodeRef, isOver, over } = useDroppable({
+      id: elementTree.root?.id as string,
```
webberwang, line 74: should throw error before if undefined?

webberwang (changes requested):

PR #2171, opened 2023-01-13
fix: use counter instead of id for component instance name

### LINKED ISSUE #1753: Drag dropping a component into the children of another component
## Current Behavior
When a component is drag droped into another component's children an error is thrown; basically because the element tree is expecting an element not a component. 

### Three possibilities exist:

1. Prevent a component from being moved into another component's children.
2. when the component is moved, create an element as an instance of the componenet being moved.
3. Delete the component and make its root element be the one placed in the recieving tree.

## Steps to Reproduce

https://user-images.githubusercontent.com/51242349/186850078-c32aaa7e-027b-471e-ac58-5a069be61aae.mp4

vladyslav-polishchuk:
## Description
use component instance counter instead of id for component instance name to provide better user experience

## Video or Image

Before            |  After
Screenshot 2023-01-13 at 15 09 58|Screenshot 2023-01-13 at 16 14 24

## Related Issue(s)

Fixes #1753

### DIFF
--- a/libs/frontend/domain/element/src/store/element.service.ts
+++ b/libs/frontend/domain/element/src/store/element.service.ts
@@ -589,9 +589,20 @@ element is new parentElement's first child
         let element = this.element(elementId)
 
         if (!element) {
+          const elementTree = Element.getElementTree(targetElement)
+
+          const existingInstances = elementTree?.elementsList.filter(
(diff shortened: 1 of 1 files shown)


webberwang:
nice quick fix

PR #2176, opened 2023-01-15
feat: update _app page with ability to render children at specific place

### LINKED ISSUE #2081: _app abstract improvement
## Description

_app abstraction improvement: allow to specify WHERE child page should be rendered. Currently, it will place inside the first leaf element, which is not always the case. For example, this approach will not work for a landing page of our application

Possibly related to 
- https://github.com/codelab-app/builder/issues/1767

vladyslav-polishchuk:
## Description

- [x] Add React.Fragment atom to seeder. This may be quite a powerful atom for templating.
- [x] Add the ability to render property with the name "children" as children of the element 
- [x] Add page props form tab with the ability to specify where to render content

## Video or Image

https://user-images.githubusercontent.com/74900868/212762802-3f6fd023-42e8-417d-8352-682d383f8616.mov

## Related Issue(s)

Fixes #2081


libs/shared/data/src/atom/html-atom.data.ts
```diff
@@ -0,0 +1,13 @@
+import { IAtomType } from '@codelab/shared/abstract/core'
+import { HtmlTag } from '../tag/html-tags.data'
+import type { AtomSeedData } from './antd-atom.data'
+
+/**
+ * Assign all data that is related to the atom here
+ */
+export const htmlAtomData: Partial<Record<IAtomType, AtomSeedData>> = {
+  [IAtomType.ReactFragment]: {
```
webberwang, line 9: React fragment isn't technically HTML but React right?

webberwang (changes requested):

libs/backend/infra/adapter/neo4j/src/schema/model/page.schema.ts
```diff
@@ -11,6 +11,7 @@ export const pageSchema = gql`
     app: App! @relationship(type: "PAGES", direction: IN)
     getServerSideProps: String
     isProvider: Boolean! @default(value: false)
+    pageContainerElementId: String
```
webberwang, line 14: could you comment the idea behind this in the code?

webberwang (changes requested):

libs/backend/infra/adapter/neo4j/src/schema/model/page.schema.ts
webberwang: this points to the root element id of _app.tsx?

vladyslav-polishchuk: this initially points to the Body element, but later it can be changed to any element of the _app tree and child pages will render as the last child of the specified element

vladyslav-polishchuk: Also, I have attached a demo video to the PR

webberwang: shouldn’t we use Element instead of string

vladyslav-polishchuk: Will update to use Element

PR #2181, opened 2023-01-17
fix(libs-frontend-domain-renderer): component should be draggable only as a whole

### LINKED ISSUE #2158: Components should only be draggable as a whole
## Current Behavior
Currently, if we use a component that has atoms in it, and put it in the main body, those atoms are draggable and can be moved outside of its component.

The component should only be draggable as a whole, and its atoms inside cannot be re-ordered or moved outside of the component

## Video Showing Bug

https://user-images.githubusercontent.com/27695022/211831725-2ac95e30-882c-4978-ba0f-e90b4ab49f79.mp4

jethrocabaluna:
## Description

- the `data-component-id` prop is stored on the global props context in the whole component
- this is used to determine if the element to be rendered is inside a component
- drag and drop feature is only added to component's root element or elements not in a component
- wrap the whole component with `<span>` instead of fragment to be able to apply drag and drop to the whole component (will also work if first children is multiple siblings)
- <span> should have less effect than `<div>` to the layout
- fix creating and moving an element inside an element + before first child
- add `zIndex` to the indicatorStyle so that it shows properly when theres a sibling element and there is no margin between

## Video or Image

https://user-images.githubusercontent.com/27695022/212916296-522aff46-6ab3-43a1-b097-1bdca1f56628.mp4

https://user-images.githubusercontent.com/27695022/212916353-800dc14a-a71d-4ed1-a0f6-1a65156301f5.mp4

## Related Issue(s)

Fixes #2158

### DIFF
--- a/libs/frontend/domain/builder/src/dnd/useDndDropHandlers.ts
+++ b/libs/frontend/domain/builder/src/dnd/useDndDropHandlers.ts
@@ -37,7 +37,15 @@ export const useDndDropHandler = (
 
     const targetElement = elementService.element(targetElementId)
 
-    if (!targetElement || !dragPosition) {
+    if (!targetElement) {
+      console.error('Target element not found')
+
+      return
+    }
+
+    if (!dragPosition) {
+      console.error('Drag position is required')
+
       return
     }
 
@@ -48,6 +56,7 @@ export const useDndDropHandler = (
 
     let newElement: Nullable<IElement> = null
 
+    // create the new element after the target element
     if (dragPosition === DragPosition.After) {
       createElementDto.prevSiblingId = targetElement.id
       newElement = await elementService.createElementAsNextSibling(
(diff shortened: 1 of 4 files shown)


libs/frontend/domain/builder/src/dnd/useDndDropHandlers.ts
```diff
@@ -57,15 +53,26 @@ export const useDndDropHandler = (
...
+
+    if (dragPosition === DragPosition.Before) {
+      if (targetElement.prevSibling) {
+        createElementDto.prevSiblingId = targetElement.prevSibling.id
+        newElement = await elementService.createElementAsNextSibling(
+          createElementDto,
+        )
+      }
+
+      if (!targetElement.prevSibling && targetElement.parentElement?.id) {
```
webberwang, line 66: For each of those block, can you add comment and summarize the logic?

webberwang (changes requested):

PR #2184, opened 2023-01-17
feat: slug enhancements

### LINKED ISSUE #2142: Slug form enhancement
## Description

- [x] App slug

When typing the name, the slug should be automatically synced (but through all lower cased)

https://user-images.githubusercontent.com/26264925/211187220-2475b8ad-0bba-4a16-9a97-f6ac7ae4bd22.mov

- [x] Page slug

Same for page slug

image

- [x] Element slug

- Shouldn't atom field be required?
- I think move atom field up, so the slug can be computed using that?

- [ ] https://github.com/codelab-app/platform/issues/2233

mohasarc:
## Description

- [x] This PR enhances the slug by automatically calculating the slug based on the name and disallowing the user from changing the slug.
- [x] We'll make sure there is 1 to 1 mapping between the names and slugs by forcing names to be alphanumeric because slug calculation usually ignores special charachters like : ; ? etc.
- [x] unrelated to slug enhancement but done in this PR: element name is calculated based on the selected atom or component while allowing the user to modify it.
- [x] #2200
- [x] fix e2e

## Video or Image

https://user-images.githubusercontent.com/51242349/214833714-7436e9d0-0a01-4daf-b540-045510718956.mp4

now allow capital letters in slug
prevent atom field from flashing id

https://user-images.githubusercontent.com/51242349/215148611-5cef19ee-e35f-45ec-a859-eec298dd5e34.mp4

## Related Issue(s)

Fixes #2142


webberwang (changes requested):
- [ ] If we enter capital letters, it shouldn't show up in slug
- [ ] The atom shouldn't flash ID or any other data

mohasarc:
Regarding the capital letters showing up in the slug, that's the default behaviour of the library we're using; however, that wouldn't be registered because names are not allowed to have capital letters.

libs/frontend/domain/app/src/use-cases/create-app/CreateAppModal.tsx
```diff
@@ -27,8 +32,13 @@ export const CreateAppModal = observer<{
       open={appService.createModal.isOpen}
     >
       <ModalForm.Form
-        model={{
-          auth0Id: userService.user?.auth0Id,
+        model={model}
+        onChange={(k, v) => {
```
webberwang, line 36: use `key`, `value` https://wiki.codelab.app/development/fullstack/code-guidelines#avoid-abbreviations

libs/frontend/domain/app/src/use-cases/update-app/updateAppSchema.ts
```diff
@@ -9,9 +9,12 @@ export const updateAppSchema: JSONSchemaType<IUpdateAppDTO> = {
     name: {
       autoFocus: true,
       ...nonEmptyString,
+      pattern: '^[a-z0-9 ]+$',
     },
     slug: {
       ...nonEmptyString,
+      disabled: true,
+      type: 'string',
```
webberwang, line 17: is this required?

libs/frontend/domain/element/src/use-cases/element/create-element/CreateElementModal.tsx
```diff
@@ -99,11 +125,20 @@ export const CreateElementModal = observer<CreateElementModalProps>(
       >
         <ModalForm.Form<ICreateElementDTO>
           model={model}
+          onChange={(k, v) => {
```
webberwang, line 128: don't use abbreviations

libs/frontend/domain/element/src/use-cases/element/create-element/createElementSchema.ts
```diff
@@ -14,12 +14,14 @@ export const createElementSchema: JSONSchemaType<ICreateElementDTO> = {
...
     name: {
-      autoFocus: true,
       type: 'string',
       nullable: true,
+      pattern: '^[a-z0-9 ]+$',
     },
     slug: {
       ...nonEmptyString,
+      disabled: true,
+      type: 'string',
```
webberwang, line 24: not required here

libs/frontend/domain/element/src/use-cases/element/update-element/UpdateElementForm.tsx
```diff
@@ -77,6 +83,16 @@ export const UpdateElementForm = observer<UpdateElementFormProps>(
         `}
         key={element.id}
         model={model}
+        onChange={(k, v) => {
+          setModel({
+            ...model,
+            slug: k === 'name' ? slugify(v) : model.slug,
```
webberwang, line 89: same

libs/frontend/domain/page/src/use-cases/update-page/UpdatePageModal.tsx
```diff
@@ -39,6 +46,13 @@ export const UpdatePageModal = observer<{ pageService: IPageService }>(
       >
         <ModalForm.Form<Omit<IUpdatePageDTO, 'pageContainerElementId'>>
           model={model}
+          onChange={(k, v) => {
+            setModel({
```
webberwang, line 50: this logic has been repeated many times, you can extract to util function

webberwang (changes requested):

libs/frontend/domain/app/src/use-cases/update-app/updateAppSchema.ts
mohasarc: oh yeah, after the addition of nonEmptyString it deosn't seem to be required anymore.

libs/frontend/domain/page/src/use-cases/update-page/UpdatePageModal.tsx
mohasarc: I couldn't find a good way to extract this part into a util function, do you have any suggestions?

libs/frontend/domain/element/src/components/auto-computed-element-name/AutoComputedElementNameField.tsx
```diff
@@ -0,0 +1,54 @@
...
+/**
+ * This component helps providing the computeElementNameService with
+ * the user input as well as connect the parent form with the final
+ * result from the computeElementNameService.
+ */
+const AutoComputedElementName = observer<AutoComputedElementNameProps>(
+  (props) => {
+    const { name, onChange, computeElementNameService, defaultValue } = props
+
+    useEffect(() => {
```
webberwang, line 27: couldn't we use a `@computed` method with mobx to derive the name?

webberwang (changes requested):

libs/frontend/domain/element/src/components/auto-computed-element-name/AutoComputedElementNameField.tsx
mohasarc: We're doing so in the mobx store. here, we're only passing the user input to the `computeElementNameService`.

PR #2190, opened 2023-01-19
fix: hot reload loading issue

### LINKED ISSUE #2064: Builder could not resolve an object ElementTreeRef

assimhabeek:
## Description

- solve `ogm.init()` error triggered by the server. we must call `OGM.init()` for each request
- refactor `loadRenderedComponentTree` to `loadRenderedComponentsTree` which accepts an array instead of a single object
- disallow `getServerSideProps` code inserted by user from running in builder
- refactor `useRenderedPage` and related hooks inside `builder` and `index` pages the error indicated by the issue is caused by invalid cache on hot reload. to solve the issue we read directly from mobx store and not `useAsync` output
- Remove `forwaredRef` deprecation warning from `Rendered`
- add `load` method  to `resourceService` which writes cache for an array of resources
- add `loadTypesByChunks` to `typeService` to group the logic spread inside `useRenderedPage`      
- replace optional operator with `Maybe` because elementTree is required but could be undefined 
- fix and refactor e2e it wasn't creating provider page which is now required to load builder
### before

https://user-images.githubusercontent.com/33001357/213865931-5492fd6a-b66a-4cc1-bfae-0d6b501873d0.mp4

### after

https://user-images.githubusercontent.com/33001357/213866060-c0ad2d7d-3b98-4208-b531-8edbb0168498.mp4

## Related Issue(s)

Fixes #2064


webberwang:
`they were triggered with inconsistent values. the inner state of useAsync  had problem. so instead of connecting hooks with returned values. read from mobx state instead`

PR #2212, opened 2023-01-27
refactor: combine the render type for create / update element form

### LINKED ISSUE #2195: Refactor ICreateElementDTO and IUpdateElementDTO to group render type
Right now, the `ICreateElementDTO` and `IUpdateElementDTO` can accept `renderComponentTypeId` and `atomId` but it actually just needs one. We can group it to make it clearer that the input will only either create/update a component or an atom element.
e.g. we can make it
```
renderType: {
 id: string
 model: 'Atom' | 'Component'
}
```

### LINKED ISSUE #2196: CreateElementModal accepts a form with having both component and atom, or none
## Current Behavior
Right now you can create an element after selecting an atom and a component, or none at all. We should prevent this in the form level (can also prevent on the backend for additional safety)

## Video Showing Bug

https://user-images.githubusercontent.com/27695022/214014276-27ee3dc7-2aa2-4f48-a63f-ca8a0ba2b86b.mp4

jethrocabaluna:
## Description

- combined the render type for atom and component so user can only select one and the form will only accept an atom or a component, not both
- refactor the interface for the `ICreateElementDTO` and `IUpdateElementDTO` for combining the render type (if atom or component)
- will still allow creating an element with no atom or component (for containers)
- fix affected e2e test for creating elements
- fix updating the element with another render type or another atom/component
- the `props` of the element will be updated when render type changes based on the interface types default values for the new render type

## Video or Image

In this video, I tested:
- creating an empty element
- creating an element with "Atom" render type
- creating an element with "Component" render type
- updating an "Atom" element to "Component" render type
- updating a "Component" element to "Atom" render type
- updating an "Atom" element with another "Atom"
- the "props" of the element will be updated as well based on the interface types default values of the new render type

https://user-images.githubusercontent.com/27695022/215537727-d26fa7e8-f28f-4021-aef9-b52ac836defd.mp4

## Related Issue(s)

Fixes #2196 
Fixes #2195


webberwang:
good work looks great at first glance, I'll likely take a more detailed look when I refactor the frontend

PR #2217, opened 2023-01-29
wip: add logs for production debugging

vladyslav-polishchuk:
## Related Issue(s)

Fixes #

### DIFF
--- a/apps/builder/middleware.ts
+++ b/apps/builder/middleware.ts
@@ -14,6 +14,9 @@ import { redirectExternalDomain } from './src/middleware/redirectExternalDomain'
  */
 export default async function middleware(req: NextRequest) {
   const hostname = req.headers.get('host')
+
+  console.log('req', req)
+
   /**
    * Check if `hostname` contains `builder-egs3r8s85-codelabai.vercel.app`, if so we don't redirect.
    */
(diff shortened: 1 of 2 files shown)


libs/backend/graphql/src/resolvers/domain/mutation/create-domains.resolver.ts
```diff
@@ -41,12 +37,9 @@ export const createDomains: IFieldResolver<
           ${domainSelection}
         }
       }`,
-    input: [
-      {
-        name,
-        id: v4(),
-        app: connectNode(appId),
```
webberwang, line 48: what's wrong with passing in app id and connecting it?

webberwang (changes requested):

libs/backend/graphql/src/resolvers/domain/mutation/create-domains.resolver.ts
vladyslav-polishchuk: There is no `appId` parameter passed from the client to the resolver, only `app` which is already connected by the id of the app. And if to try passing `appId` there is runtime error about unrecognized parameter.
I fixed the TS error though, so must look better now

Issue #2163, opened 2023-01-12
Misc refactor for Renderer


webberwang:
## Description

- [x] Avoid using spread since we have used pick already, how come eslint not working here?

![image](https://user-images.githubusercontent.com/26264925/212011274-0085d141-4458-4572-9cab-1ed23cb6329a.png)

- [x] Remove `stop` and use `RenderIfExpression` in `ElementWrapper`. Create separate issue and link here

- [x] Move `renderComponentMeta` to `Component` model

- [ ] ESLint rule to restrict to arrow function

![image](https://user-images.githubusercontent.com/26264925/212012716-975b6e58-5540-4520-979f-b4de8f457391.png)

- [ ] Replace `isBuilder` and `isComponentBuilder` with `builderType`

PR #2219, opened 2023-01-30
feat: replacing pmb with experssion

### LINKED ISSUE #2144: Replacing PMB with Expressions

Replace Prop Map Binding (PMB) with Expressions. The prop map binding concept can be replaced with expressions for mapping props between elements.

```
A (Component)
  \ B
    \ C 
```

Component A interface:

```
{
  "c_children": "ReactNode"
}
```

Say our goal is to map the prop from Component A's `c_children` property to the `children` of `C`. We can use `C.children` expression field, and choose from Component A's `c_children` property

---

Some brainstorming...

The expression field should have 2 fields. The first field will allow us to select the source data. Maybe there is a local/global toggle, where local only shows the component API & global shows app state.

---

Would require below 

- [ ] #2145 
- [x] #2146

assimhabeek:
## Description

- No need to pass `components`,`renderComponentMeta` to `render.model` anymore  in order to count component render index. we clone component and its tree. partially indicated by #2163
- No need for `COMPONENT_INSTANCE_ID` component now has `instanceElement` property which returns the source element.
- Add `rootElement` to `element.model` which traverse the elements tree and reach the root node. so to access component props from any element we do `rootElement.parentComponent.props`
- Add setters to `IElement` so we are able to clone element model correctly.
- Add `clonedElements` to `elementService` as separation between persisted and runtime elements. Using `elementService.element()` search in both maps. so it is recommended to use in most cases instead of `elementService.elements.get()`
- Same applies for `componentService` and `clonedComponents`
- For now `replaceStateInProps` will accept context `global` which is state+instance props+component props + api default values  or `component`:  which is component props and api default values
- Remove `stop` from renderOutput indicated by #2163
- Remove `unnecessary` indicated by #2163
- Refactor `renderChildren` in `render.model` 
- Fix `css.cy` tests
## Video or Image

https://user-images.githubusercontent.com/33001357/217214095-3e69a444-0353-4d7d-9c7a-776cad36963f.mp4

## Related Issue(s)

Fixes #2144


libs/frontend/abstract/core/src/domain/builder/render.service.interface.ts
```diff
@@ -9,9 +8,9 @@ import type { IRenderer } from './renderer.model.interface'
 export interface RendererProps {
   pageTree: IElementTree
   appStore: IStore
-  components: Array<IComponent>
   appTree?: Nullable<IElementTree>
   isBuilder?: boolean
```
webberwang, line 12: instead of having 2 mutually exclusive booleans, we could have a `builderType` with enum

libs/frontend/domain/element/src/store/element.service.ts
```diff
@@ -1120,4 +1123,27 @@ element is new parentElement's first child
 
     return propMapBinding
   })
+
+  @modelAction
+  writeClonesCache(elementFragment: IElementDTO) {
+    return [...this.clonedElements.values()]
+      .filter((c) => c.sourceElementId === elementFragment.id)
```
webberwang, line 1130: don't use abbreviation

libs/frontend/domain/element/src/store/element.service.ts
```diff
@@ -1120,4 +1123,27 @@ element is new parentElement's first child
...
+      .map((e) =>
```
webberwang, line 1131: don't use abbreviation

libs/frontend/domain/element/src/store/element.service.ts
```diff
@@ -1120,4 +1123,27 @@ element is new parentElement's first child
...
+        e.writeCache({
+          ...elementFragment,
+          parentComponent: e.parentComponent?.current
+            ? ({
+                id: e.parentComponent.current.id,
+              } as IElementDTO['parentComponent'])
```
webberwang, line 1137: why need for casting?

webberwang (changes requested):

libs/frontend/abstract/core/src/domain/builder/render.service.interface.ts
assimhabeek: correct. we should add that as a refactor

libs/frontend/domain/element/src/store/element.service.ts
assimhabeek: it requires a whole fragment, but we only need  id.

libs/frontend/domain/element/src/store/element.service.ts
assimhabeek: that is a bad habit, I need to get rid-off

PR #2226, opened 2023-02-01
feat: props inspector on component props

### LINKED ISSUE #2201: Allow to view and edit component props via prop inspector
## Description
Currently, any element inside of the page builder has props inspector tab where actual rendered props can be viewed as well as edited.
Screenshot 2023-01-23 at 21 25 30

But this is not available for components builder at the moment. Add support for props inspector for components in a similar way to elements

jethrocabaluna:
## Description

This PR is for allowing to view and edit the component props in component builder.

Currently, we only show the props inspector if the node type is "Element". Since components can also have props in the component builder, we can update the props inspector to also show on "Component" type.

I updated the `usePropsInspector` so that it can support both node type and user can also update the props in the component builder.

## Video or Image

In this video, I tested in component props inspector:
- editing the props and it updated the api values in the props tab
- updated the api in the Props tab and it updated the values in the props inspector
- updated the props in the code mirror modal
- clearing the props resets the props to default values set in the api

https://user-images.githubusercontent.com/27695022/216574244-0b462615-d5b0-4753-a690-6016d3229946.mp4

## Related Issue(s)

Fixes #2201

### DIFF
--- a/libs/frontend/abstract/core/src/domain/atom/atom.model.interface.ts
+++ b/libs/frontend/abstract/core/src/domain/atom/atom.model.interface.ts
@@ -3,7 +3,7 @@ import type { IEntity, Nullish } from '@codelab/shared/abstract/types'
 import type { Ref } from 'mobx-keystone'
 import type { ICacheService } from '../../service'
 import type { ITag } from '../tag'
-import type { IAnyType } from '../type'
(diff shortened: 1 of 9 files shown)


webberwang:
nice work

PR #2227, opened 2023-02-02
refactor: change name of production page route

### LINKED ISSUE #1617: Enable custom domains for apps

vladyslav-polishchuk:
## Description
Before this PR the dynamic route name was `_sites/user/[user]/[app]/[page]`.
And in the middleware, it was required to compute [user] and [app], which was impossible without querying the DB.
DB query case 2 problems:
1. It will be executed each time user navigates to ANY page
For example, the user has deployed the app on `user-app.com`; whenever anyone wants to access any page for that app, middleware has to be called to understand the name of the static page to redirect to, which requires to do DB request.
2. Users not authenticated to our builder website (`admin.codelab.app`) will not be able to access `user-app.com` because they do not have a valid session and the middleware will not be able to query DB and perform redirect.

As a solution - static route name is renamed to `_sites/[domain]/[page]`. This provides the same level of uniqueness (no 2 same domains may exist) while no need to do any requests to redirect to this route, all data is already available in middleware.

## Related Issue(s)

Fixes #1617


webberwang:
good optimization thanks

Issue #2224, opened 2023-02-01
Personal portfolio website


webberwang:
## Description

We'll be copying other websites to test out our own platform.

Re-create https://leerob.io. 

- [x] Create all the pages, link each page to domain whitelist (your own domain). Render the title on each page only
- [ ] Test out menu link navigation
- [ ] Responsive menu, tablet mobile view should go on top
- [x] Add all simple content with antd grid layout when needed for all pages
- [x] Add github sign in
- [x] Blog page link to supabase

vladyslav-polishchuk:
Not fully implemented. Only a tiny part of the app was created which most probably will not be compatible after the last major frontend refactoring

PR #2234, opened 2023-02-05
fix: fixes layout shift and duplicated requests

vladyslav-polishchuk:
## Description
During the demo app creation, some issues were detected and fixed:
- spinner appeared before the tabs component and it caused a layout shift (components below the spinner jumped down and up)
- requests for atoms were repeated a lot of times, use atom service with caching instead of manual request. Those requests are slowing down the production builder site a lot:

https://user-images.githubusercontent.com/74900868/217049644-9601e326-208e-416f-a06d-49d5bf6c787b.mov

- fix instability in e2e test because of - https://github.com/codelab-app/platform/issues/2235

## Related Issue(s)

#2224 

@vladyslav-polishchuk can remove fixes keyword so it’s linked at least

### DIFF
--- a/libs/frontend/domain/atom/src/store/atom.service.ts
+++ b/libs/frontend/domain/atom/src/store/atom.service.ts
@@ -32,6 +32,7 @@ import { AtomModalService, AtomsModalService } from './atom-modal.service'
 export class AtomService
   extends Model({
     id: idProp,
+    allAtomsLoaded: prop(() => false),
     atoms: prop(() => objectMap<IAtom>()),
     count: prop(() => 1),
     createModal: prop(() => new ModalService({})),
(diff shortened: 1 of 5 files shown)


libs/frontend/domain/atom/src/store/atom.service.ts
```diff
@@ -32,6 +32,7 @@ import { AtomModalService, AtomsModalService } from './atom-modal.service'
 export class AtomService
   extends Model({
     id: idProp,
+    allAtomsLoaded: prop(() => false),
```
webberwang, line 35: I think this relates to a larger issue with caching. 

With other clients, caching is done automatically like Apollo.

Instead of saying whether all items are loaded, we assume cache is updated. This is the cache then http strategy or something like that 

Whenever another operation affects the items, we will mark the cache as stale, and the next time an operation requires the data, we will fetch it.

This is similar to how redux toolkit, where we specify some query key to see which models to bust the cache 

This seems like a lot, are there other general solutions for the caching?

webberwang (changes requested):

PR #2241, opened 2023-02-07
fix: infinite render when creating state variable

### LINKED ISSUE #2216: Too many rerenders error on new state variable cration
## Current Behavior
This is probably because of duplicated keys exist for store actions and variables. But could be a different reason as well

https://user-images.githubusercontent.com/74900868/215322331-46bab486-98c0-4c54-a933-f07a3bf84939.mov

jethrocabaluna:
## Description

Upon investigation, it seems this [part](https://github.com/codelab-app/platform/blob/master/libs/frontend/domain/renderer/src/renderer.model.ts#L327) causes the current `props` to be recursively added to the element in this [part](https://github.com/codelab-app/platform/blob/master/libs/frontend/domain/renderer/src/renderer.model.ts#L334). But what causes the infinite renders is the use of  [this.elementPropMap.set](https://github.com/codelab-app/platform/blob/master/libs/frontend/domain/renderer/src/ExtraElementProps.ts#L35) which always sets a new object to the element even if there is no change in the props.

image

This PR:
- spreads the `propsForCurrentElement` into the new `props` instead of putting it inside another `props`
- used [Ramda's whereEq](https://ramdajs.com/docs/#whereEq) on [addForElement](https://github.com/codelab-app/platform/blob/master/libs/frontend/domain/renderer/src/ExtraElementProps.ts#L34) so that if the new `props` to be added already exists in the current `props`, it will just skip

## Video or Image

https://user-images.githubusercontent.com/27695022/217235111-377eeb00-307f-4922-9d56-d6e1f0d84759.mp4

## Related Issue(s)

Fixes #2216

### DIFF
--- a/libs/frontend/domain/renderer/src/ExtraElementProps.ts
+++ b/libs/frontend/domain/renderer/src/ExtraElementProps.ts
@@ -14,6 +14,7 @@ import {
   objectMap,
   prop,
 } from 'mobx-keystone'
+import { whereEq } from 'ramda'
 
 @model('@codelab/ExtraElementProps')
 export class ExtraElementProps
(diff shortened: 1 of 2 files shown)


webberwang:
does lodash isEqual do the same thing?

PR #2250, opened 2023-02-08
feat: add validations to props inspector

### LINKED ISSUE #2229: No validations when updating the props via props inspector
## Current Behavior
When updating the api in the props tab, thats fine because there are validations because of the form used. However when updating the props in the props inspector, it does not run any validations and we can save a wrong prop type.

## Video Showing Bug

https://user-images.githubusercontent.com/27695022/216588921-07ecb943-fdce-4434-ae4c-e3db148e50e6.mp4

jethrocabaluna:
## Description

This PR is for validating the edited props in the props inspector with the Element's atom/component interfaceType. The validator is created when props inspector is mounted and will use the selected element's interfaceType to generate the schema to be used for the validation against the editable props.

Also moved [this](https://github.com/codelab-app/platform/pull/2250/files#diff-5dbea52d1684a4f6c0b3927d9494c9298debec9fdedf981a37957d7e4570fa09L44) out of `ref` so that when it is updated, it automatically re-renders the props form with the new values

This PR also includes updating the `defaultValues` field so that if the `nullable` for the prop is not true, it will require the `defaultValues` and will display an error if not provided. The `nullable` field is also hidden when the type is Boolean and it will be default to `false`. The default values for the primitives are also removed so that if the field is nullable and currently has a null value, there will be no value displayed in the field.

Also updated the Default cell on the fields table so that it shows the default values (for primitives only)

## Video or Image

In this video, I tested validation against atom/component props

https://user-images.githubusercontent.com/27695022/217552289-38ddf1a2-4b83-43e9-81a4-2050e62856d5.mp4

<br />
In this video, I tested the nullable defaultValues for the primitive props. The old behaviour is that there will always be a default value even if the prop is nullable

https://user-images.githubusercontent.com/27695022/218710649-71d686f9-2eaf-42e7-87fe-ee4c8c55de18.mp4

<br />

Showing defaults on the fields table
image

## Related Issue(s)

Fixes #2229

### DIFF
--- a/libs/frontend/domain/type/src/interface-form/type-schema.factory.ts
+++ b/libs/frontend/domain/type/src/interface-form/type-schema.factory.ts
@@ -44,13 +44,6 @@ export const primitives = {
   [PrimitiveTypeKind.Boolean]: 'boolean' as const,
 }
 
-const primitivesDefaults = {
-  [PrimitiveTypeKind.Boolean]: false,
-  [PrimitiveTypeKind.Number]: 0.0,
-  [PrimitiveTypeKind.Integer]: 0,
-  [PrimitiveTypeKind.String]: '',
-}
-
 export class TypeSchemaFactory {
   constructor(private readonly options?: TransformTypeOptions) {}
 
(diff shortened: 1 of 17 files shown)


libs/frontend/domain/type/src/interface-form/type-schema.factory.ts
```diff
@@ -95,12 +95,16 @@ export class TypeSchemaFactory {
...
+  fromInterfaceType(
+    type: IInterfaceType,
+    context?: UiPropertiesContext,
+  ): JsonSchema {
     const makeFieldSchema = (field: IField) => ({
       label: field.name || compoundCaseToTitleCase(field.key),
       ...this.transform(field.type.current, {
         validationRules: field.validationRules ?? undefined,
         fieldName: field.name,
+        useDefaults: context?.useDefaults,
```
webberwang, line 107: Can't we just disable the ajv use defaults like this https://ajv.js.org/guide/modifying-data.html#assigning-defaults

webberwang (changes requested):

webberwang:
looks well done, I suppose we could require users to create a default value if it is not nullable.

so either nullable or give a default value

PR #2255, opened 2023-02-09
fix: static page regeneration issue

### LINKED ISSUE #1617: Enable custom domains for apps

vladyslav-polishchuk:
## Description
Manual static page regeneration of user pages works only when executed on the user domain. So for example, if we want to regenerate some pages for `some-user-app.com` domain - we should specify the header with the domain name, even though they share the same codebase.

## Video or Image

https://user-images.githubusercontent.com/74900868/218331515-313809a0-cd03-45d8-9256-ebcfaf6b5bfa.mov

## Related Issue(s)

Fixes #1617

### DIFF
--- a/apps/builder/pages/api/regenerate.ts
+++ b/apps/builder/pages/api/regenerate.ts
@@ -22,7 +22,10 @@ const getAllPagesToRevalidate = async () => {
       for (const page of app.pages) {
         if (!page.isProvider) {
           const pageSlug = extractSlug(page.slug)
-          pathsToRevalidate.push(`/_sites/${domain.name}/${pageSlug}`)
+          pathsToRevalidate.push({
+            domain,
+            path: `/_sites/${domain.name}/${pageSlug}`,
+          })
         }
       }
     }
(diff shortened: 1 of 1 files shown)


apps/builder/pages/api/regenerate.ts
```diff
@@ -91,15 +100,22 @@ const regenerate: NextApiHandler = async (req, res) => {
     const revalidatedPages: Array<string> = []
     const failedPages: Array<string> = []
 
-    const revalidationPromises = pathsToRevalidate.map(async (path) => {
-      try {
-        await res.revalidate(path)
+    const revalidationPromises = pathsToRevalidate.map(
+      async ({ path, domain }) => {
```
webberwang, line 104: I would pass domain/page to this function, and compute the path here. better to keep entity as a whole for as long as possible (down the function chain)

webberwang (changes requested):

PR #2264, opened 2023-02-13
refactor: make slugs computed from names

### LINKED ISSUE #2142: Slug form enhancement

### LINKED ISSUE #2233: Allow capital letters for app/page/elements/etc names
## Current Behavior
Screenshot 2023-02-05 at 13 41 13
Screenshot 2023-02-05 at 13 41 38
Screenshot 2023-02-05 at 13 42 06

### LINKED ISSUE #2246: Remove page/element slug
Remove `slug` from page/element and force the following rules for page/elements names:

- Should start with a letter and consist of letters, numbers and space.
- Make sure it is unique,  just like we did with slug.
- To use in object keys/urls replace space with either underscore or dash

mohasarc:
## Description

This PR aims at removing slugs creation, update, and storage and make them computed from the name in the backend. Cosequently, the name should be unique and restricted in terms of charachters to have 1 to 1 mapping to the computed slug.

TODO:
- [x] make slugs computed
- [x] enforce uniqueness on names
- [x] revise the slugs logic in element.service/page.service to make sure everything is working fine (element/page slugs used to have parent id to make them unique only inside the parent but names don't)
- [x] remove app slugs
- [x] find solution for element.repo and app.repo where selectionSet includes slug but ogm fails to provide slug
- [x] ~transition calculating slugs from graphql to mobx~
- [x] fix e2e

## Video or Image

Fixes #2246
Fixes #2233 
Fixes #2142


libs/frontend/abstract/core/src/domain/app/app.fragment.graphql
```diff
@@ -1,7 +1,6 @@
 fragment AppPreview on App {
   id
   name
-  slug
```
webberwang, line 4: needs to be added back?

libs/backend/graphql/src/resolvers/app/app.resolver.ts
```diff
@@ -0,0 +1 @@
+export const appSlug = ({ name }: { name: string }) => name.replace(/\s/g, '-')
```
webberwang, line 1: why don't we use voca here?

libs/backend/graphql/src/resolvers/element/element.resolver.ts
```diff
@@ -1,23 +1,2 @@
...
- * }
- *
- * Used `ITxnResolver<{ id: string }>` before, but new signature is more flexible
- */
-export const elementDescendants =
-  (parent: { id: string }) => (txn: Transaction) => {
-    return elementRepository.getDescendants(txn, parent.id)
-  }
+export const elementSlug = ({ name }: { name: string }) =>
+  name.replace(/\s/g, '-')
```
webberwang, line 2: voca

libs/backend/graphql/tsconfig.lib.json
```diff
@@ -7,5 +7,9 @@
     "types": ["node"]
   },
   "exclude": ["jest.config.ts", "**/*.spec.ts", "**/*.test.ts"],
-  "include": ["**/*.ts"]
+  "include": [
+    "**/*.ts",
+    "../application/repo-resolvers/resolvers/domain/domain.error.ts",
```
webberwang, line 12: why is this included?

webberwang (changes requested):

libs/backend/graphql/tsconfig.lib.json
mohasarc: this must have happened by mistake, I'll remove it.

PR #2282, opened 2023-02-18
fix: get rid of cy.go which does page reload

### LINKED ISSUE #2280: Provider page e2e fails in headless mode
## Description

During pre-push this spec keeps failing, but during `yarn e2e` it works. CI also fails

image

The builder input does get disabled, but the assertion fails

https://user-images.githubusercontent.com/26264925/219786866-163dae45-eb1b-4c30-a3a5-aaa7e3fbf249.mp4

---

I've commented the spec out for now

vladyslav-polishchuk:
## Description

Full page reload initiated by `cy.go('back')` sometimes was preventing the request from being finished. And the settings on ConfigProvider were not always saved. Get rid of browser navigation and use the header menu to navigate between pages, which fixes instability and is also faster

## Related Issue(s)

Fixes #2280


webberwang:
ohh good catch

PR #2287, opened 2023-02-20
fix: type table search not working

### LINKED ISSUE #2276: Type table search does not work
## Description

Currently the Ant Design table searches using data from the frontend, but with pagination we don't have all the data on the front.

We'll need to call backend API with `limit` and `offset` to get the current types

---

A dev mentioned that search was working before this PR, 

- https://github.com/codelab-app/platform/pull/2148

You can use that as reference

---

https://user-images.githubusercontent.com/26264925/219482956-7eec37d4-900b-4d4d-a1eb-678539f34fef.mov

rexdev0211:
## Description

Fixed Closes #2276 
GetTypesTable component loads only pagesize count of types, so result is not good when user search types.
If Ant Design Table doesn't contain full types, it is difficult to search types.
Modified section loads BaseTypes in GetTypesTable component.

## Video or Image

https://user-images.githubusercontent.com/87414033/220055785-8a0c9d12-aba1-48d4-8a5c-3c171193303b.mp4

## Related Issue(s)

Fixes #2276

### DIFF
--- a/libs/frontend/domain/type/src/use-cases/types/get-types/GetTypesTable.tsx
+++ b/libs/frontend/domain/type/src/use-cases/types/get-types/GetTypesTable.tsx
@@ -52,16 +52,20 @@ export const GetTypesTable = observer<{
     [router],
   )
 
-  const loadAllBaseTypes = async () => {
-    await getBaseTypes({
-      offset: 0,
-      limit: curPageSize,
-    })
-    await getBaseTypes({
-      offset: 0,
-      limit: typeService.count,
-    })
-  }
+  const loadAllBaseTypes = useCallback(
+    async (pagesize: number) => {
+      let callCounter = 0
+
+      while (callCounter < 2) {
+        await getBaseTypes({
+          offset: 0,
+          limit: callCounter === 0 ? pagesize : typeService.count,
+        })
+        callCounter++
+      }
+    },
+    [curPageSize],
+  )
 
   /**
    * Change page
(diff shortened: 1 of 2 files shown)


libs/frontend/domain/type/src/use-cases/types/get-types/GetTypesTable.tsx
```diff
@@ -53,6 +52,17 @@ export const GetTypesTable = observer<{
     [router],
   )
 
+  const loadAllBaseTypes = async () => {
+    await getBaseTypes({
+      offset: 0,
```
webberwang, line 57: how come we fetch twice?

webberwang (changes requested):

libs/frontend/domain/type/src/use-cases/types/get-types/GetTypesTable.tsx
rexdev0211: Get total count of types from first fetch.
Get total types from second fetch.

libs/frontend/domain/type/src/use-cases/types/get-types/GetTypesTable.tsx
```diff
@@ -63,24 +73,13 @@ export const GetTypesTable = observer<{
   }, [curPage, pageSize, pagination])
 
   useEffect(() => {
-    const offset = (curPage - 1) * curPageSize
-    void getBaseTypes({
-      offset,
-      limit: curPageSize,
-    })
-  }, [curPage, curPageSize, getBaseTypes])
```
webberwang, line 71: we removed all params, shouldn't it be based on some params

rexdev0211: Yes. It is difficult to search types in table because table has only pagesize count of types, so table should have full types and I removed some params and modified.

PR #2289, opened 2023-02-20
refactor: cleanup and fix websites project

### LINKED ISSUE #2232: Extract functionality of _sites to separate project
## Description
Currently, builder and production user pages share the same project and this causes a couple of problems:
1. middleware.ts is always executed for requests to the builder and to _sites, but if split into two separate projects - this middleware is not needed at all.
2. _app.tsx file adds a lot of code that is used mostly for builder functionality and is not needed for user pages. So splitting may reduce bundle size for user apps

vladyslav-polishchuk:
## Description
- Optimize static page regeneration process: pass page slugs instead of page ids and then no need to send requests to DB from the regenerate endpoint. Also, get rid of the code that was not used, no need to maintain scenarios that are not in use. If needed we can always use git history.
- Cleanup terraform copy-pasted code
- Cleanup _app page

Perfomance results:

Before             |  After
Before Lighthouse|Stage 1 Lighthouse

## Related Issue(s)
Fixes #2232


webberwang:
that's a decent improvement in performance

PR #2295, opened 2023-02-22
refactor: use code mirror on "render if" field

### LINKED ISSUE #2252: Make "Render If" field a CodeMirror one
## Description
Currently "Render If" is a plain text field, though it is intended to hold an expression. 
It is complicated to write expressions without syntax highlighting and suggestions.
So the suggestion is to use CodeMirror instead of text input

Video to demonstrate an issue:

https://user-images.githubusercontent.com/74900868/220135457-f1b7d09b-7a5f-4b3f-a5fd-b47b6a613718.mov

jethrocabaluna:
## Description

This is for replacing the normal string field of **Render If** into a code mirror.

Refactored the logic of providing the props options and removed the unused [usePropsCompletion](https://github.com/codelab-app/platform/blob/3244eb84af4b70be0dbb890c32dab749c633b6d5/libs/frontend/domain/builder/src/hooks/usePropCompletion.ts). The filtering of props while typing will still work and are handled internally on both code mirror and antd's AutoComplete component

Also refactored the logic around evaluating the `renderIfExpressions`. I moved [these](https://github.com/codelab-app/platform/pull/2295/files#diff-a7cb14661aefe4266b00f0f20f03c7cf011cec67313d2901c2c329b86f0942aaL132) out and they are now just a [normal utility functions](https://github.com/codelab-app/platform/pull/2295/files#diff-e901077e368a354f625db234971d97a86b2bc34a5651d6c096dec2f939543f36R55) since it makes sense that the conditional rendering should use both state and props

### Extra
I added the condition [here](https://github.com/codelab-app/platform/pull/2295/files#diff-fd9129e27b7bbdce155ea18b8a8f07101b9165b6afb1e4bd939da3528ce4846aR64) so that it only applies when using autosave

## Video or Image

https://user-images.githubusercontent.com/27695022/220899912-1396028c-16cb-44f8-a011-f0cc33f4cb8a.mp4

## Related Issue(s)

Fixes #2252


webberwang:
the refactor looks very good, much less code

PR #2305, opened 2023-02-23
fix: not authenticated error on page regeneration

### LINKED ISSUE #2232: Extract functionality of _sites to separate project

vladyslav-polishchuk:
## Description
- fixed auth0 403 error when requested to regenerate app pages.
Now the flow is the following:
user clicks the "rebuild" button on UI -> request to the builder backend is sent (admin.codelab.app) -> builder backend redirects the request to the appropriate domain of the websites project (for example to codelab-demo.online/api/regenerate). Each backend endpoint is protected by auth0 so an unauthenticated user will not be able to trigger the regeneration of any page.

- fixed website page routing after automatic slugs were implemented
after automatic slugs were implemented it became hard to navigate to user pages since they started to look like this: `https://codelab-demo.online/3-fe-3-e-3-b-3-0852-4056-9625-d-7-ba-490-a-125-a-home`. Used slugified page name to navigate to pages instead

## Related Issue(s)

Fixes #2232

### DIFF
--- a/apps/websites/pages/api/regenerate.ts
+++ b/apps/websites/pages/api/regenerate.ts
@@ -1,3 +1,5 @@
+// eslint-disable-next-line @nrwl/nx/enforce-module-boundaries
+import { pageSlug } from '@codelab/backend/graphql'
 import { auth0Instance } from '@codelab/shared/adapter/auth0'
 import type { NextApiHandler } from 'next'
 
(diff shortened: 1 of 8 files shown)


apps/websites/pages/api/regenerate.ts
```diff
@@ -23,7 +24,8 @@ const regenerate: NextApiHandler = async (req, res) => {
     const revalidatedPages: Array<string> = []
     const failedPages: Array<string> = []
 
-    const revalidationPromises = pages.map(async (pageSlug) => {
+    const revalidationPromises = pages.map(async (pageName) => {
+      const pageSlug = slugify(pageName)
```
webberwang, line 28: We actually have a computed slug from the graphql api we could use

webberwang: Same for all other models

webberwang (changes requested):

PR #2320, opened 2023-02-28
fix: forbidden errors in e2e tests

### LINKED ISSUE #2319: Forbidden errors in the e2e tests
## Current Behavior
It seems all the e2e tests in the current PRs are failing due to a Forbidden error which causes some graphql mutations to not succeed that may be used for seeding some data used for the test

image

jethrocabaluna:
## Description

It seems that the user data needs to be upserted first before calling the mutations for seeding the data for testing.

I updated the [loginSession](https://github.com/codelab-app/platform/pull/2320/files#diff-36371f76c76798c489abdf2ec5b8891c261a5ab11d4ca2f756bc6e685d2e62a2R17) so that it visits the page where the [upsert of user data will happen](https://github.com/codelab-app/platform/blob/master/apps/builder/pages/apps/index.tsx#L62) before proceeding with the mutations.

## Related Issue(s)

Fixes #2319

### DIFF
--- a/apps/builder-e2e/src/e2e/admin/admin.cy.ts
+++ b/apps/builder-e2e/src/e2e/admin/admin.cy.ts
@@ -5,9 +5,6 @@ describe('Admin', () => {
   before(() => {
     cy.resetDatabase()
     loginSession()
-    // Visit so we can trigger upsert user
-    cy.visit('/apps')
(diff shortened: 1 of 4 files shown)


webberwang:
nice!

PR #2337, opened 2023-03-06
fix: remove atoms custom resolver

mohasarc:
## Description

this PR removes custom atoms resolver as it was rendered unnecessary after the latest neo4j update.

## Related Issue(s)

Fixes #


libs/frontend/domain/atom/src/use-cases/get-atoms/GetAtomsTable.tsx
```diff
@@ -84,7 +85,9 @@ export const GetAtomsTable = observer<GetAtomsTableProps>(
         id: atom.id,
         library: getAtomLibrary(atom.type),
         name: atom.name,
-        tags: atom.tags.map((tag) => tag.current),
+        tags: atom.tags
+          .map((tag) => tag.maybeCurrent)
+          .filter(Boolean) as Array<ITag>,
```
mohasarc, line 90: I'm adding this because for some reason tags are not created when atoms are created, we only have the ref.

webberwang: shouldn't be a need to cast if we filter correctly

webberwang (changes requested):

PR #2354, opened 2023-03-09
fix: elements swapping issues

jethrocabaluna:
## Description

- removed the `make*inputs` methods
- this follows the optimistic update we are trying to implement where we update the cache first then do network calls in a transaction

The issue was that the elements are not resolved correctly when there are multiple elements with the same `parent` because it is being treated as the same as `firstChild` relationship in [here](https://github.com/codelab-app/platform/blob/0d0ff6c1057095cd9232c0eb5a1547854ae06119/libs/backend/infra/adapter/neo4j/src/schema/model/element.schema.ts#L13).

This PR makes sure the connected nodes are correctly updated in the cache so that the [input](https://github.com/codelab-app/platform/pull/2354/files#diff-b94c779f8caff32713bbc7c19a7ac77b8ca9ef6e419712f7b16b4d45e235626cR403) is correctly generated for disconnecting/connecting nodes.

Also added [updateNodes](https://github.com/codelab-app/platform/pull/2354/files#diff-6a88f5a3eede7ca9edcd9091bdfe6c9521e4d605bfdb27fa04d0a6f4b4f7ec6aR41) on the element repo since it seems to be faster when there are less fields so that elements swapping will be faster in the network calls.

The implementation has changed so that we don't need to generate the [inputs](https://github.com/codelab-app/platform/pull/2354/files#diff-b94c779f8caff32713bbc7c19a7ac77b8ca9ef6e419712f7b16b4d45e235626cL558) for each operation for updating, and just base on the current data of an element

## Video or Image

https://user-images.githubusercontent.com/27695022/224760629-6597bec0-3c6e-43ee-aa55-c2ba6060dcf4.mp4

https://user-images.githubusercontent.com/27695022/224760735-a944e60a-a146-4d68-90db-76caaf56a66d.mp4

## Related Issue(s)

Fixes #

### DIFF
--- a/libs/frontend/domain/element/src/store/element.model.ts
+++ b/libs/frontend/domain/element/src/store/element.model.ts
@@ -38,16 +38,8 @@ import { compoundCaseToTitleCase, mergeProps } from '@codelab/shared/utils'
 import attempt from 'lodash/attempt'
 import isError from 'lodash/isError'
 import { computed } from 'mobx'
-import {
-  clone,
-  idProp,
-  Model,
-  model,
-  modelAction,
-  prop,
-  Ref,
-} from 'mobx-keystone'
-import { makeUpdateElementInput } from './api.utils'
+import type { Ref } from 'mobx-keystone'
+import { clone, idProp, Model, model, modelAction, prop } from 'mobx-keystone'
 import { getElementTree } from './element-tree/element-tree.util'
 import { getRenderType } from './utils'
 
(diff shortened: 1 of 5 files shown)


libs/frontend/domain/element/src/store/element.model.ts
```diff
@@ -501,34 +502,29 @@ export class Element
...
+     * Connect nextSibling to the parent
+     */
+    if (this.parent.current.firstChild?.id === this.id) {
+      // Connect parent to nextSibling
+      this.parent.current.firstChild = this.nextSibling?.isValid
+        ? elementRef(this.nextSibling.current)
+        : null
+
+      // Connect nextSibling to parent
+      // this.nextSibling?.maybeCurrent?.setParent(elementRef(this.parent.id))
```
webberwang, line 519: this is still needed right?

libs/frontend/domain/element/src/store/element.model.ts
```diff
@@ -410,12 +402,18 @@ export class Element
...
       name: this.name,
+      nextSibling: reconnectNodeId(this.nextSibling?.id),
+      parent: reconnectNodeId(this.parent?.id),
       postRenderAction: reconnectNodeId(this.postRenderAction?.id),
       preRenderAction: reconnectNodeId(this.preRenderAction?.id),
+      prevSibling: reconnectNodeId(this.prevSibling?.id),
       props: {
-        update: { node: { data: JSON.stringify(this.props.current.data) } },
+        update: {
+          node: { data: JSON.stringify(this.props.maybeCurrent?.data ?? {}) },
```
webberwang, line 415: we should use `current` so it throws error for us to catch the bug?

PR #2356, opened 2023-03-09
refactor: type repository

mohasarc:
## Description

TODO:

- [x] map create type data to type dto
- [x] implement type repository
- [ ] Figure out why getBaseTypes is different from getAll
- [ ] merge getAllWithDescendants with getAll by adding an optional `fetchDescendants? = false` param to getAll?

## Related Issue(s)

Fixes #


libs/frontend/domain/type/src/use-cases/types/create-type/create-type.factory.ts
```diff
@@ -36,5 +34,5 @@ export const createTypeFactory = (
       type.kind === ITypeKind.UnionType
         ? makeTypesOfUnionTypeCreateInput(type)
         : undefined,
-  }))
+  } as ICreateTypeInput
```
webberwang, line 37: why need casting here?

mohasarc: Idk a weird behavior by typescript, it shows an error without the casting. Anyways this code is deleted now.

PR #2359, opened 2023-03-11
[WIP] fix: ts build error, e2e tests, non hidden system fields

vladyslav-polishchuk:
## Description
- Fixed when in production not all fields were hidden on create/update forms of App/Page/Atom/etc
- Fixed build error because of typescript
- Fixed some e2e errors

## Related Issue(s)

Fixes #


apps/builder-e2e/src/e2e/component.cy.ts
```diff
@@ -34,14 +34,14 @@ describe('Component CRUD', () => {
     cy.resetDatabase()
     loginSession()
     cy.getCurrentUserId()
```
webberwang, line 36: should rename to `getCurrentOwner()`, then we can use `connectAuth0Owner(owner)`

webberwang (changes requested):

PR #2373, opened 2023-03-16
fix: element creation shows error and resets fields while processing

jethrocabaluna:
## Description

This fixes the issue where when creating an element, the form fields get reset while still processing and this is due to updating the cache first before the actual save and is causing the CreateElementModal to re-render because the [element options](https://github.com/codelab-app/platform/pull/2373/files#diff-1e051f74cebcc472394998ff540f6d5e0788a178f1b5d90e79abff5581e44887L67) got updated.

This PR is for supplying the create modal with the element options upon opening so that the options does not change while the modal is open.

### Before

https://user-images.githubusercontent.com/27695022/225651739-0cce78e0-323b-4128-bee9-fa536d91ef3d.mp4

<br />

### After

https://user-images.githubusercontent.com/27695022/225651930-8c792def-717b-4d13-bab2-0ffc8d2cee7d.mp4

## Related Issue(s)

Fixes #

### DIFF
--- a/libs/frontend/abstract/core/src/domain/element/element.service.interface.ts
+++ b/libs/frontend/abstract/core/src/domain/element/element.service.interface.ts
@@ -20,6 +20,11 @@ import type { IElementTree } from './element-tree.interface.model'
  * Used for modal input
  */
 export interface CreateElementData {
+  elementOptions: Array<{
+    childrenIds?: Array<string>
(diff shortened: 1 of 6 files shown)


webberwang (changes requested):
Are we able to copy the options so it isn't observable anymore?

PR #2375, opened 2023-03-17
refactor: field repo

mohasarc:
## Related Issue(s)

Fixes #


libs/frontend/domain/type/src/store/field.service.ts
```diff
@@ -59,98 +59,40 @@ export class FieldService
...
-      interfaceTypeId,
-    ) as IInterfaceType
-
-    interfaceType.writeFieldCache([field!])
-
-    return this.add(field!)
+    const field = this.add(FieldService.mapDataToDTO(createFieldData))
+    const interfaceType = this.typeService.type(field.api.id) as IInterfaceType
+
+    interfaceType.writeFieldCache([field])
```
webberwang, line 67: should be `fieldModel.writeCache()`

webberwang (changes requested):

PR #2380, opened 2023-03-18
fix: user not upserted, apps not loaded

vladyslav-polishchuk:
## Description

- fixes but when user not upserted on visiting apps, since `user.sub` always exists event when user does not exist in neo4j db
- fix issue when apps not loaded on /apps page
- removed `components-tab.cy` test since other tests already test component functionality
- fixes build error

## Related Issue(s)

Fixes #


apps/builder/pages/apps/index.tsx
```diff
@@ -59,16 +59,17 @@ const AppsPage: CodelabPage<DashboardTemplateProps> = (props) => {
   const { appService, domainService, userService } = useStore()
   const { user } = useUser()
 
-  const [{ error, result: app, status }, loadApp] = useAsync(
-    (owner: IAuth0Owner) => appService.loadAppWithNestedPreviews({ owner }),
+  const [{ result: app, status }, loadApp] = useAsync((owner: IAuth0Owner) =>
```
webberwang, line 62: isn't this the same as before? how does it fix?

PR #2383, opened 2023-03-19
fix: only one apps is loaded

melhamin:
## Related Issue(s)

Fixes #


libs/frontend/domain/app/src/store/app.service.ts
```diff
@@ -188,19 +188,21 @@ export class AppService
...
   @modelFlow
-  loadAppWithNestedPreviews = _async(function* (
+  loadAppsWithNestedPreviews = _async(function* (
     this: AppService,
     where: AppWhere,
   ) {
-    const appData = (yield* _await(this.appRepository.find(where)))[0]!
-    const app = this.add(appData)
+    const appsData = yield* _await(this.appRepository.find(where))
+    const apps = appsData.map((appData) => this.add(appData))
```
webberwang, line 196: why not combine into single map

libs/frontend/domain/app/src/use-cases/get-apps/GetAppsItem.tsx
```diff
@@ -17,20 +18,22 @@ export interface GetAppsItemProps {
...
-    const appDomains = domains?.filter((domain) => domain.app.id === app.id)
-    const Title = <Link href={href}>{app.name}</Link>
+export const GetAppsItem = observer<GetAppsItemProps>(({ app, appService }) => {
+  const href = { pathname: PageType.PageList, query: { appId: app.id } }
 
-    const Dropdown = (
-      <ItemDropdown app={app} appService={appService} domains={appDomains} />
-    )
+  const appDomains = app.domains
+    .map((domain) => domain.maybeCurrent)
```
webberwang, line 25: use current here

webberwang (changes requested):

PR #2387, opened 2023-03-20
fix: component issues, landing project build

vladyslav-polishchuk:
## Description

Fixed a couple of component-related issues:
- issues with cloning (some of the component data was not cloned)
- issue with updating component props on form
- issue when ref id was used instead of actual element/component id

## Related Issue(s)

Fixes #

### DIFF
--- a/libs/frontend/domain/component/src/store/component.model.ts
+++ b/libs/frontend/domain/component/src/store/component.model.ts
@@ -119,7 +119,7 @@ export class Component
         clonedElement.setRenderType(componentRef(componentClone.id))
       }
 
-      if (element.id === clonedComponent.childrenContainerElement.id) {
+      if (element.id === this.childrenContainerElement.maybeCurrent?.id) {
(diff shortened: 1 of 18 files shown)


libs/frontend/domain/component/src/store/component.model.ts
```diff
@@ -135,13 +135,32 @@ export class Component
       ? elementMap.get(this.elementTree.root.id)
       : null
 
+    elements.forEach((element) => {
+      const { firstChild, nextSibling, parent, prevSibling } = element
+      const parentId = parent && elementMap.get(parent.current.id)
+      const firstChildId = firstChild && elementMap.get(firstChild.current.id)
```
webberwang, line 141: I think more clear if we do

```typescript
if (parentId) {
  const parentElement = elementMap.get(parent.current.id)
  element.setParent(elementRef(parentElement))
}
```

for each of the 4

webberwang (changes requested):

PR #2406, opened 2023-03-23
fix: script error on new page creation

vladyslav-polishchuk:
## Description
Root element was missing reference to parent page which caused script error on attempt to open newly created page:

Screenshot 2023-03-23 at 17 53 18

## Related Issue(s)

Fixes #


webberwang:
odd that I hadn't seen this

PR #2426, opened 2023-03-30
fix: unit tests

### LINKED ISSUE #2419: Fix unit specs renderer & others
## Description

Possibly can use store setup as our main app, meaning only services

create-root-store_ts_—_platform

This way we can re-use some create methods

component_service_ts_—_platform

We just have to stub out the repositories since we're doing a unit specs without database connection

https://jestjs.io/docs/jest-object#jestspyonobject-methodname

- [ ] Revert DTO changes

image

We shouldn't have to modify the dto to make the specs work

vladyslav-polishchuk:
## Description
Fix unit tests. Refactor setup setup-tests.ts to use services to create test data

## Related Issue(s)

Fixes #2419

### DIFF
--- a/libs/frontend/domain/renderer/src/test/setup/setup-test.ts
+++ b/libs/frontend/domain/renderer/src/test/setup/setup-test.ts
@@ -2,24 +2,20 @@
 
 import type { IType } from '@codelab/frontend/abstract/core'
 import {
-  componentRef,
   CUSTOM_TEXT_PROP_KEY,
   elementRef,
   elementTreeRef,
+  IRenderTypeKind,
   RendererType,
   ROOT_ELEMENT_NAME,
 } from '@codelab/frontend/abstract/core'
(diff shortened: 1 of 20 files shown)


libs/frontend/domain/renderer/src/test/setup/setup-test.ts
```diff
@@ -30,219 +26,202 @@ import {
...
       name: ROOT_ELEMENT_NAME,
-      props: propRef(data.elementToRenderProps.id),
-      propTransformationJs: `
-    // Write a transformer function, you get the input props as parameter
-    // All returned props will get merged with the original ones
-    function transform(props) {
-      return Object.keys(props)
+    })
+    data.elementToRender.setPage(pageRef(pageId))
+    data.elementToRender.setProps(propRef(elementToRenderProps))
```
webberwang, line 139: I don't think we need those set methods, we can just pass in prop id to elementService.create

webberwang (changes requested):

Issue #2215, opened 2023-01-29
Poor performance on production because of duplicated requests


vladyslav-polishchuk:
## What Feature Can Be Enhanced?
Reduce amount of duplicated requests.

Example 1 

https://user-images.githubusercontent.com/74900868/215321999-31f63524-9bf9-41c6-a3f2-088d16e08b34.mov

Example 2 

https://user-images.githubusercontent.com/74900868/215322089-5856c5f0-262a-423f-b263-1a7264454be0.mov

PR #2449, opened 2023-04-05
fix: use interface from store if exists

vladyslav-polishchuk:
## Description

- Do not request types over and over again if they were already loaded
- Fix linting errors on exported files with data (no new line in the end)
- Reduce the amount of parallel unit tests runs in CI. This is an experimental fix, since unit tests currently fail on CI almost in 100% cases with strange errors like `A jest worker process (pid=11453) was terminated by another process: signal=SIGKILL, exitCode=null.` This may be because of memory issues.

## Video or Image

Before

https://user-images.githubusercontent.com/74900868/230017315-71274696-b682-4bd2-9f81-8bffa25f73e1.mov

After

https://user-images.githubusercontent.com/74900868/230017349-758461a8-af7a-4cc8-8b4c-6fa6f5c2c48c.mov

## Related Issue(s)

One of the fixes under #2215

### DIFF
--- a/apps/cli/src/commands/tasks/tasks.command.ts
+++ b/apps/cli/src/commands/tasks/tasks.command.ts
@@ -64,7 +64,7 @@ export const tasksCommand: CommandModule<unknown, unknown> = {
 
           if (stage === Stage.CI) {
             execCommand(
-              'npx nx affected --target=test --testPathPattern="[^i].spec.ts" --color --parallel=3 --verbose',
+              'npx nx affected --target=test --testPathPattern="[^i].spec.ts" --color --verbose',
             )
           }
         },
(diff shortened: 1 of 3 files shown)


apps/cli/src/commands/tasks/tasks.command.ts
```diff
@@ -64,7 +64,7 @@ export const tasksCommand: CommandModule<unknown, unknown> = {
 
           if (stage === Stage.CI) {
             execCommand(
-              'npx nx affected --target=test --testPathPattern="[^i].spec.ts" --color --parallel=3 --verbose',
```
webberwang, line 67: so this removes parallel?

vladyslav-polishchuk: Very often random unit tests fail with some strange "A jest worker process (pid=11453) was terminated by another process: signal=SIGKILL, exitCode=null" error. This is an experimental fix to check if the root cause of the issue is concurrency. So if unit tests will stop failing, looks like it is better to run then sequentially. Otherwise - we can revert and try something else

PR #2451, opened 2023-04-06
fix: default values field validation schema

### LINKED ISSUE #2208: Custom Field for defaultValues on the create/update field modal for array types
We have a custom field made for the `defaultValues` if it is an array in the create/update field form but the built-in validations from uniforms does not work properly.
In this [PR](https://github.com/codelab-app/builder/pull/2199), theres a custom schema made for the defaultValues but the errors doesn't show on the uniforms fields for the `defaultValues`.

When a new item is added, the initial value shows the default for that primitive type, but in the actual model it is still undefined, hence, it will have an error on the validations upon submitting. The workaround is to manually enter a value so that its value changes in the model.

https://user-images.githubusercontent.com/27695022/214823921-02eb5ccf-c485-4977-b021-98df4c33c24c.mp4

A solution could be to transform the undefined values to the default values of whatever is the item type for the array, but there may be too much overhead, and may have to create some kind of recursive function if it is an array of array (or more array).

### LINKED ISSUE #2423: UpdateFieldModal issues
## Current Behavior

1. if default value field is empty, for most fields in the UpdateFieldModel like `name`, 'description', and other validation fields, the update button does not sumbit even if the values for these fields are changed.
2. Cannot reset default value.

## Video Showing Bug

https://user-images.githubusercontent.com/51242349/228643437-3b4cf803-d64d-4e74-8cc0-0256701abdf2.mp4

https://user-images.githubusercontent.com/51242349/230371449-d5b1482e-2621-42de-b935-d88856e8dcb1.mp4

mohasarc:
## Description

This PR does some changes to the validation schema of `defaultValues` field of the field schema. Additionally, it fixes the issue of not being able to reset the value of the default value.

The main issue was that because `oneOf` is used in `fieldDefaultValues` definition and all the options provided are `nullable: true`, when the value is null, multiple intities are matched; hence the value is not valid because we want only one of them to match.

one question that comes to mind would be that there is a conditon that removes `defaultValues` feild from the required list if `validationRules.general.nullable == true`. However, the field not being required does not mean that it's value will not be checked against the schema. hence, the if consition here practically does nothing.

To fix that:
- The `nullable: true` is removed from `fieldDefaultValues`
- The schema for `defaultValues` doesn't use `fieldDefaultValues` definition immedietly, but a union type of the form:
```
{
   anyOf: [
    { $ref: 'customTypes#/definitions/fieldDefaultValues' },
    { 
        anyOf: [
           { $ref: 'customTypes#/definitions/fieldDefaultValues' }
           { type: 'null' },
        ]
     }  
   ]
}
```
the reason for this is that to be able to use the if condition correctly, we need to specify all posibilities as a union type (using `anyOf`). here the first posibility is `fieldDefaultValues` definition that is not nullable. The other posibility is either fieldDefaultValues` definition or a null type.
- finally, we need to specify which of the posibilities to check on `then` if `validationRules.general.nullable == false`
```
then: {
    properties: {
      defaultValues: {
            $ref: 'customTypes#/definitions/fieldDefaultValues',
      },
    },
  },
```
which means don't accept null. jva infers that the `else` part would be the other posibility (the one that accepts null type)

## Video or Image

https://user-images.githubusercontent.com/51242349/230355660-78584211-5e17-4d4e-b718-b8bf70f940a6.mp4

https://user-images.githubusercontent.com/51242349/230355762-f35b585c-a146-4dac-a855-c4c9890640e2.mp4

https://user-images.githubusercontent.com/51242349/230373210-480fd7f7-56f6-4f15-afc5-0375c613ae33.mp4

## Related Issue(s)

Fixes #2423
Fixes #2208

### DIFF
--- a/libs/frontend/view/components/form/hooks/uniformUtils.ts
+++ b/libs/frontend/view/components/form/hooks/uniformUtils.ts
@@ -30,29 +30,47 @@ ajv.addSchema({
   definitions: {
     // this allows validation on array or object type that references itself
     fieldDefaultValues: {
-      oneOf: [
+      anyOf: [
         {
-          nullable: true,
           type: 'string',
         },
         {
           type: 'boolean',
         },
         {
-          nullable: true,
           type: ['number', 'integer'],
         },
         {
           items: { $ref: '#/definitions/fieldDefaultValues' },
-          nullable: true,
           type: 'array',
         },
         {
-          nullable: true,
           patternProperties: {
             '^.*$': { $ref: '#/definitions/fieldDefaultValues' },
           },
           type: 'object',
+        },
+      ],
+    },
+    // Adding these definitions here to avoid type errors because
+    // JSONSchemaType does not support unions although json schema does
+    fieldDefaultValuesOrNullableFieldDefaultValues: {
+      anyOf: [
+        {
+          $ref: '#/definitions/fieldDefaultValues',
+        },
+        {
+          $ref: '#/definitions/nullableFieldDefaultValues',
+        },
+      ],
+    },
+    nullableFieldDefaultValues: {
+      anyOf: [
+        {
+          $ref: '#/definitions/fieldDefaultValues',
+        },
+        {
+          type: 'null',
         },
       ],
     },
--- a/libs/frontend/domain/type/src/store/models/field.model.ts
+++ b/libs/frontend/domain/type/src/store/models/field.model.ts
@@ -94,9 +94,7 @@ export class Field
     this.validationRules = validationRules
       ? JSON.parse(validationRules || '{}')
       : this.validationRules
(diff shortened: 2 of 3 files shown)


webberwang:
we should use oneOf instead of anyOf?

webberwang (changes requested):

libs/frontend/view/components/form/hooks/uniformUtils.ts
```diff
@@ -30,32 +30,50 @@ ajv.addSchema({
   definitions: {
     // this allows validation on array or object type that references itself
```
webberwang, line 31: do you mean by using `anyOf` instead of `oneOf`?

PR #2455, opened 2023-04-09
fix: can't create second app

### LINKED ISSUE #2454: Can't create second app
## Video Showing Bug

https://user-images.githubusercontent.com/74900868/230735084-aadf006d-79e8-40b8-b891-b21649b09886.mov

vladyslav-polishchuk:
## Description

In an attempt to create a second app an error about 'constraint validation failed' was thrown (see the video in issue). This is because each page has its own store and store API, where the `name` was computed as `{pageName} Store API`. Since API name has to be unique - a collision happens when pages with the same names exist in different apps: in this case system pages (_app, 404, 500).
In order to fix this - store api is created using the pattern `{appName}({userName}) {pageName} Store API`. This will allow pages with the same name

Screenshot 2023-04-12 at 10 37 11

## Related Issue(s)

Fixes #2454


webberwang:
This won't work across users though

vladyslav-polishchuk:
> This won't work across users though

Updated to use user id as well

libs/frontend/domain/page/src/store/page.service.ts
```diff
@@ -153,10 +153,14 @@ export class PageService
...
     })
 
+    const appModel = this.appService.apps.get(app.id)
+
     const interfaceType = this.typeService.addInterface({
       id: v4(),
       kind: ITypeKind.InterfaceType,
-      name: InterfaceType.createName(`${name} Store`),
+      name: InterfaceType.createName(
+        `${appModel?.name}(${owner.auth0Id}) ${name} Store`,
```
webberwang, line 162: Username is unique we could use that instead?

libs/frontend/abstract/core/src/domain/user/user.interface.ts
```diff
@@ -1,9 +1,9 @@
 import type { IRole } from '@codelab/shared/abstract/core'
-import type { Ref } from 'mobx-keystone'
-import type { IApp } from '../app'
+// import type { Ref } from 'mobx-keystone'
+// import type { IApp } from '../app'
 
 export interface IUser {
-  apps: Array<Ref<IApp>>
+  // apps: Array<Ref<IApp>>
```
vladyslav-polishchuk, line 6: Commented out since it is not used and it caused circular reference after adding `userService` dependency to `app.service`

apps/cli/src/commands/tasks/tasks.command.ts
```diff
@@ -64,7 +64,7 @@ export const tasksCommand: CommandModule<unknown, unknown> = {
 
           if (stage === Stage.CI) {
             execCommand(
-              'npx nx affected --target=test --testPathPattern="[^i].spec.ts" --color --verbose',
+              'npx nx affected --target=test --testPathPattern="[^i].spec.ts" --color --parallel=3 --verbose',
```
vladyslav-polishchuk, line 67: Reverted back to dun tests in parallel. I though this might fix issue with random unit tests failures on CI, but it did not work

PR #2463, opened 2023-04-11
refactor: simplify app query to use less resources

### LINKED ISSUE #2441: Fix Neo4j db crash when running with minimal resources
## Current Behavior
Neo4j DB crashes on an attempt to get all apps with dependencies (simply visiting the /apps page would crash) when the DB runs with few resources.
This reproduces in 100% of cases on the production where the DB has 1GB of memory and 1 Core:

https://user-images.githubusercontent.com/74900868/229530838-16c70c3d-5dd0-4cfc-9305-721268403796.mov

And the cloud reports out-of-memory errors:
Screenshot 2023-04-03 at 15 56 14

But this also can be reproduced locally. If to allocate a minimal amount of resources to the Docker image - the same behavior happens

https://user-images.githubusercontent.com/74900868/229531456-4f09c546-baed-4cea-80cb-f0dfc2150852.mov

If to increase resources - everything starts to work properly:

https://user-images.githubusercontent.com/74900868/229531616-b2529ec2-0b19-4aed-8331-3afddfa61037.mov

The query to request the apps is quiet complex. As a possible solution we can refactor apps data loading to not perform such complex query.

vladyslav-polishchuk:
## Description

Query to get app and all related resources (pages, elements, components, stores, etc) cause the image that has only 1GB RAM to crash with an Out Of Memory error.
Reduce the complexity of the query by replacing full components information loading with only ID.
Not the query successes on 1GB ram image

Before

https://user-images.githubusercontent.com/74900868/231205525-69cfdbd9-a1c8-4a89-8d5c-ad5ca0825612.mov

## Related Issue(s)

Fixes #2441

### DIFF
--- a/libs/frontend/abstract/core/src/domain/element/element.fragment.graphql
+++ b/libs/frontend/abstract/core/src/domain/element/element.fragment.graphql
@@ -8,7 +8,7 @@ fragment Element on Element {
     id
   }
   renderComponentType {
-    ...Component
+    id
   }
   renderAtomType {
     ...Atom
@@ -25,7 +25,7 @@ fragment Element on Element {
     id
   }
   parentComponent {
-    ...Component
+    id
   }
   parent {
     id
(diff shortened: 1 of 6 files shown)


webberwang:
do we need to fetch the component elsewhere?

vladyslav-polishchuk:
> do we need to fetch the component elsewhere?

It is already fetched as a part of the GetRenderedPageAndCommonAppData query. So all the components should already be in store

PR #2471, opened 2023-04-13
fix: move element to another tree crashes the builder

### LINKED ISSUE #2465: Moving element between components issues
## Current Behavior

1. When an element is moved from one component into another, an error is thrown and the builder crashes. however, everything works fine after refresh.
2. Circular reference is possible because it's not checked when drag/dropping a component to another component.

## Video Showing Bug

1.
https://user-images.githubusercontent.com/51242349/231375305-043e6325-4c46-4ec2-bf6f-a7e0efeba601.mp4

2. 

https://user-images.githubusercontent.com/51242349/231728069-d83d6aac-0c95-4f24-b964-e566e6c28ccd.mp4

mohasarc:
## Description

The issue stems from pushing partial updates to the backend. more specifically, we used to detach the elements from their current tree, push updates, then attach to the new tree. and it seems like mobx tried to recalculate the value of `closestContainerNode` while we're pushing the updates, which causes the crash.

Instead we can perform all the actions on the local cache: detaching from current tree and attaching to the new tree, then push the changes to the backend.

On a separate note, the logic of `moveElementToAnotherTree` seemed quite complicated because it tries to perform one of two things: either move an element to another tree, or create a new element in the other tree as an instance of the component. hence, I renamed it to `moveNodeToAnotherTree` and made it call another method based on what will be performed: `moveElementToAnotherTree` or `moveComponentToAnotherTree`

TODO: 

- [x] fix builder crashes when moving element to another tree
- [x] prevent circular reference when drag/dropping a component to another.

## Video or Image

https://user-images.githubusercontent.com/51242349/231723971-d424af48-4e7c-48d9-93d5-100f349b0dab.mp4

https://user-images.githubusercontent.com/51242349/231737427-703750a4-beb0-477d-bd0b-3bdb5f85da8e.mp4

## Related Issue(s)

Fixes #2465

### DIFF
--- a/libs/frontend/abstract/core/src/domain/element/element.service.interface.ts
+++ b/libs/frontend/abstract/core/src/domain/element/element.service.interface.ts
@@ -83,9 +83,11 @@ export interface IElementService
     element: IEntity
     targetElement: IEntity
   }): Promise<void>
+  /**
+   * @param props.object an element or a component
+   */
   moveObjectToAnotherTree(props: {
     dropPosition: number
-    // an element or a component
     object: IEntity
     targetElement: IEntity
   }): Promise<void>
--- a/libs/frontend/abstract/core/src/domain/element/element.service.interface.ts
+++ b/libs/frontend/abstract/core/src/domain/element/element.service.interface.ts
@@ -83,9 +83,10 @@ export interface IElementService
     element: IEntity
     targetElement: IEntity
   }): Promise<void>
-  moveElementToAnotherTree(props: {
+  moveObjectToAnotherTree(props: {
     dropPosition: number
-    element: IEntity
+    // an element or a component
+    object: IEntity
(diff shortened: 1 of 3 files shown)


libs/frontend/abstract/core/src/domain/element/element.service.interface.ts
```diff
@@ -83,9 +83,12 @@ export interface IElementService
     element: IEntity
     targetElement: IEntity
   }): Promise<void>
-  moveElementToAnotherTree(props: {
+  /**
+   * @param props.object an element or a component
+   */
+  moveObjectToAnotherTree(props: {
```
webberwang, line 89: Can you rename to `moveNode`

webberwang (changes requested):

PR #2479, opened 2023-04-14
refactor: atoms and types dropdown

### LINKED ISSUE #2476: App freezes while atoms and types items for the dropdown are being fetched and stored
## Current Behavior

It seems the process of fetching and storing all atoms for the first time freezes the app for a while.

Some options we could do:
 - make it only trigger upon opening the atoms dropdown (but will still freeze)
 - not store all atoms to cache because maybe its too heavy when theres like a hundred of items to store.
 - make the dropdowns for types and atoms select field paginated by scroll

We could do the same for the Types select field too.

## Video Showing Bug

In this video, im trying to click on the other tabs but nothing is happening, while the atoms are getting resolved on the select atoms field

https://user-images.githubusercontent.com/27695022/231942224-2ff53ab1-1f78-476c-9622-db8120a7bf15.mp4

jethrocabaluna:
## Description

- created new queries for [atoms](https://github.com/codelab-app/platform/pull/2479/files#diff-36c113451e50e0945bc589072123548132618c3473c3b2dde8574d552162b0bfR29) and [types](https://github.com/codelab-app/platform/pull/2479/files#diff-b876f80c92ce861ebb6a7992b0928b077561c15751f7669eaa2e84710c73742aR157) to be specifically used for fetching all the atoms and types for the dropdown
- new queries are a lot smaller compared to the older query being used where lots of extra unused fields are being fetched
- created new methods for the atoms and types service (`getOptions`) that will use the new queries and will not do extra stuffs like storing to cache or loading underlying types that potentially freezes the app
- since this reduces the size of the responses and the extra processes like storing and loading stuffs, this is a lot quicker now and doesn't have to do pagination on the dropdown which just adds complex codes

## Video or Image

Loading of types options on the dropdown is very quick now and doesn't freeze the app

https://user-images.githubusercontent.com/27695022/232540964-cd3022d0-71a1-4a99-93f0-834688fc63e3.mp4

<br />

Loading of atoms options on the dropdown is very quick now and doesn't freeze the app

https://user-images.githubusercontent.com/27695022/232541213-c80a1258-c81a-4438-81dc-95eb2c662c7c.mp4

## Related Issue(s)

Fixes #2476

### DIFF
--- a/libs/frontend/domain/type/src/graphql/get-type.endpoints.graphql
+++ b/libs/frontend/domain/type/src/graphql/get-type.endpoints.graphql
@@ -153,3 +153,15 @@ query GetCodeMirrorTypes(
     ...Type
   }
 }
+
+query GetTypeOptions {
+  # needs hardcoded limit here since the cypher query for base types requires a limit
+  # and it is set to 10 by default in the backend and is safer to just hardcode here
+  baseTypes(options: { limit: 99999 }) {
+    items {
+      id
+      name
+      kind
+    }
+  }
+}
--- a/libs/frontend/domain/type/src/interface-form/fields/select-atom/SelectAtom.tsx
+++ b/libs/frontend/domain/type/src/interface-form/fields/select-atom/SelectAtom.tsx
@@ -1,9 +1,11 @@
 import type { IAtom } from '@codelab/frontend/abstract/core'
-import { filterNotHookType } from '@codelab/frontend/abstract/core'
 import { useStore } from '@codelab/frontend/presenter/container'
 import type { UniformSelectFieldProps } from '@codelab/shared/abstract/types'
-import { useAsync, useMountEffect } from '@react-hookz/web'
+import { useAsync } from '@react-hookz/web'
+import compact from 'lodash/compact'
+import uniqBy from 'lodash/uniqBy'
 import React from 'react'
+import { useField } from 'uniforms'
 import { SelectField } from 'uniforms-antd'
 
 export type SelectAtomProps = Pick<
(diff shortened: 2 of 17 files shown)


libs/frontend/domain/type/src/graphql/get-type.endpoints.graphql
```diff
@@ -153,3 +153,15 @@ query GetCodeMirrorTypes(
     ...Type
   }
 }
+
+query GetTypeOptions {
+  # needs hardcoded limit here since the cypher query for base types requires a limit
+  # and it is set to 10 by default in the backend and is safer to just hardcode here
+  baseTypes(options: { limit: 99999 }) {
```
webberwang, line 160: could we make the cypher query such that if no limit is passed we don't do any limit?

libs/frontend/domain/type/src/interface-form/fields/select-atom/SelectAtom.tsx
```diff
@@ -16,51 +18,42 @@ export type SelectAtomProps = Pick<
...
   return (
     <SelectField
       error={error || queryError}
       getPopupContainer={(triggerNode) => triggerNode.parentElement}
       label={label}
-      loading={loading}
+      loading={status === 'loading'}
       name={name}
+      onDropdownVisibleChange={async (open) => {
+        if (open && status === 'not-executed') {
```
webberwang, line 51: I think we need a cohesive way to handle caching with ttl

webberwang (changes requested):

Issue #2495, opened 2023-04-18
Make sure props form uses field order


webberwang:
## Description

In the builder view make sure the fields are generated in the same fields order

You can create an element with an atom (choose button for testing). Then you can click the props form at the right

PR #2526, opened 2023-04-23
fix: make props form use field order

amrmuhamedd:
## Description
make props field follow the fields order 

## Video o

https://user-images.githubusercontent.com/47860740/233812381-eee3672e-eed8-4615-9c90-93fafd3e42ed.mp4

r Image

## Related Issue(s)
#2495 Make sure props form uses field order

Fixes #
#2495  Make sure props form uses field order

### DIFF
--- a/libs/frontend/domain/type/src/shared/sortFields.ts
+++ b/libs/frontend/domain/type/src/shared/sortFields.ts
@@ -1,26 +1,19 @@
 import { type IField } from '@codelab/frontend/abstract/core'
+import sortBy from 'lodash/sortBy'
 
 export const sortFieldsArray = (fields: Array<IField>) => {
-  const compareNodes = (field1: IField, field2: IField) => {
-    let node1 = field1
-    let node2 = field2
-    const pathA = [field1.id]
-    const pathB = [field2.id]
+  const nodes = new Map(fields.map((field) => [field.id, field]))
 
-    while (node1.prevSibling) {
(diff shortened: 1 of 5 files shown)


libs/frontend/domain/type/src/shared/sortFields.ts
```diff
@@ -0,0 +1,26 @@
+import { type IField } from '@codelab/frontend/abstract/core'
```
webberwang, line 1: could use lodash `sortBy`

webberwang (changes requested):

webberwang:
I asked chatgpt4 to refactor with lodash sort

Mouse_Highlight_Overlay_and_Refactor_array_sort_lodash

amrmuhamedd:
chatGpt solution didn't work correctly I try to refactor my code and get :

```
import { type IField } from '@codelab/frontend/abstract/core'
import { sortBy } from 'lodash'

export const sortFieldsArray = (fields: Array<IField>) => {
  const nodes = new Map(fields.map((field) => [field.id, field]))
  const getPath = (field: IField) => {
    const path = [field.id]
    while (field.prevSibling) {
      field = nodes.get(field.prevSibling.id) as IField
      path.unshift(field.id)
    }
    return path.join()
  }

  return sortBy(fields, [(field) => getPath(field), 'id'])
}
```

PR #2528, opened 2023-04-23
fix: apply global css fix

### LINKED ISSUE #2523: Separate scroll for Inspector panel
Currently, the inspector panel shares the same scroll with the full builder page, which is sometimes not what the user may expect.
For example, if the element props form has a lot of props - user needs to scroll the page down, and then the user can't see the effect after changing some element prop without scrolling back to the top:

https://user-images.githubusercontent.com/74900868/233800436-dfbc29df-731a-4780-a42f-988a1d03b7ba.mov

vladyslav-polishchuk:
## Description

Apply `antd/dist/reset.css` to fix common CSS problems.
Also stabilize e2e test that failed on this PR.

This fixes the scrolling problem described in the current issue:

https://user-images.githubusercontent.com/74900868/233830592-35c64a8f-b1df-4b1e-8e46-3dc70b7e57c3.mov

And also some other issues with sizing, padding, margins:

Before             |  After
Screenshot 2023-04-23 at 11 06 09|Screenshot 2023-04-23 at 11 03 05

Before             |  After
Screenshot 2023-04-23 at 11 05 46|Screenshot 2023-04-23 at 11 03 44

Before             |  After
Screenshot 2023-04-23 at 11 05 10|Screenshot 2023-04-23 at 11 04 26

## Related Issue(s)

Fixes #2523

### DIFF
--- a/apps/platform/pages/_app.tsx
+++ b/apps/platform/pages/_app.tsx
@@ -5,6 +5,11 @@ import 'react-quill/dist/quill.snow.css'
 import '../src/styles/quill.snow.override.css'
 import 'react-grid-layout/css/styles.css'
 import 'react-resizable/css/styles.css'
+// apply fix for common css problems:
+// - remove default padding/margin from html and body
+// - set 100% width and height for html and body
+// - set box-sizing, remove outlines, etc
+import 'antd/dist/reset.css'
 // https://www.elvisduru.com/blog/how-to-customize-ant-design-theme-in-nextjs
 import { UserProvider } from '@auth0/nextjs-auth0'
 import type { IAppProps, IPageProps } from '@codelab/frontend/abstract/core'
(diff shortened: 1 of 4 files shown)


apps/platform/pages/_app.tsx
```diff
@@ -5,6 +5,11 @@ import 'react-quill/dist/quill.snow.css'
 import '../src/styles/quill.snow.override.css'
 import 'react-grid-layout/css/styles.css'
 import 'react-resizable/css/styles.css'
+// apply fix for common css problems:
+// - remove default padding/margin from html and body
+// - set 100% width and height for html and body
+// - set box-sizing, remove outlines, etc
+import 'antd/dist/reset.css'
```
webberwang, line 12: Would we need to add to to `landing` and `websites`?

webberwang (changes requested):

apps/platform/pages/_app.tsx
vladyslav-polishchuk: Yes, that makes sense

PR #2531, opened 2023-04-24
fix: state expression not working in ReactNodeType

### LINKED ISSUE #2285: ReactNodeType and RenderPropsType do not have access to component props
## Current Behavior
Currently when creating an expression for ReactNodeType or RenderPropsType field inside the component - it is impossible to get values of component props.
So for example on a link component, it is possible to access component props for expression on "Href" string prop, but impossible for "title" which is of ReactNodeType

## Video Showing Bug

https://user-images.githubusercontent.com/74900868/219970669-1b789a81-b7e7-47a0-8d6d-9fc6d6d71370.mov

mohasarc:
## Description

The issue was that when transforming the `ReactNodeType` in `ReactNodeTypedValueTransformer` we aren't passing the render props as part of the context, hence the created component has no access to that data.

## Video or Image

https://user-images.githubusercontent.com/51242349/234057965-5b3ce540-85ca-42a3-8cc3-4185d31aac5f.mp4

## Related Issue(s)

Fixes #2285

### DIFF
--- a/libs/frontend/domain/renderer/src/abstract/ITypedValueTransformer.ts
+++ b/libs/frontend/domain/renderer/src/abstract/ITypedValueTransformer.ts
@@ -1,7 +1,9 @@
+/* eslint-disable @typescript-eslint/no-explicit-any */
 import type {
   IBaseRenderPipe,
   IComponentService,
   IElementService,
+  IPropData,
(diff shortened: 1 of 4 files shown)


webberwang:
good fix!

PR #2546, opened 2023-04-25
refactor: builder renderer

### LINKED ISSUE #2537: Refactor renderer prop passing

Seems like the renederer is passed many levels deep, could be a code smell.

Mouse_Highlight_Overlay_and_Renderer_tsx_—_platform

Builder_tsx_—_platform

If we move to context, we could memoize at the context level, and keep it a single object without having to break it up halfway passing the prop down.

I think better to keep object as whole throughout the props passing

jethrocabaluna:
## Description

This refactors the implementations around using services and props passing in the builder renderer. This also includes some UI enhancements around loading the data.

Also checked if there is a lot of re-renders happening again due to refactor, seems its fine now. I haven't notice any slowdown or performance issues so far with the refactor.

## Video or Image

Before:
image

After:
image

<br />

Testing the builder

https://user-images.githubusercontent.com/27695022/234649200-260de3ea-4873-437a-bd4f-e3fdb1c72bc8.mp4

## Related Issue(s)

Fixes #2537


apps/platform/pages/apps/[appId]/pages/[pageId]/builder.tsx
```diff
@@ -63,46 +35,50 @@ PageBuilder.Layout = observer(({ children }) => {
...
   const contentStyles = useMemo(() => ({ paddingTop: '0rem' }), [])
 
-  const ConfigPaneComponent = useCallback(
-    () => (
-      <Spinner isLoading={isLoading}>
-        <ConfigPane renderService={renderer} />
-      </Spinner>
-    ),
-    [isLoading, renderer],
+  const ConfigPaneComponent = () => (
```
Comment (vladyslav-polishchuk) [inline apps/platform/pages/apps/[appId]/pages/[pageId]/builder.tsx:40]: Could you check if removing`useCallback` would not cause re-renders, please? You can use `wdyr.ts` for that.

Comment (vladyslav-polishchuk) [reply apps/platform/pages/apps/[appId]/pages/[pageId]/builder.tsx:40]: I added these `useCallback`s recently because they were causing a lot of content to be re-rendered and this resulted to instability in e2e tests when they tried to click/type into nodes that no longer exist (new nodes created after re-rendering). But this may be no longer the issue after your refactoring, so it would be good to check once you finish

Comment (webberwang) [reply apps/platform/pages/apps/[appId]/pages/[pageId]/builder.tsx:40]: Actually wdyr doesn't work with mobx https://github.com/welldone-software/why-did-you-render/issues/162

Comment (vladyslav-polishchuk) [reply apps/platform/pages/apps/[appId]/pages/[pageId]/builder.tsx]: @webberwang, @jethrocabaluna, Just uncommenting import of `wdyr.ts` in `_app.tsx` does not work, but I am using it like this (directly pasting the code into _app.tsx)
Screenshot 2023-04-27 at 09 19 01

Then it works just fine
Screenshot 2023-04-27 at 09 20 35

webberwang:
Looks great! the UI enhancements are nice

Issue #2508, opened 2023-04-19
Element disappears after changing props


melhamin:
## Current Behavior
Issue occurs in some atoms (`AntDesign Select` here), not all atoms.

## Video Showing Bug

https://user-images.githubusercontent.com/48331678/233182711-f3a40e20-8e3e-4a77-a546-75b98003f5ee.mov

vladyslav-polishchuk:
Issue fixed

PR #2548, opened 2023-04-25
fix: don't apply "false" by default for bool props

vladyslav-polishchuk:
## Description

Once an element is created - no props are passed to it and it renders using default values only. But when the first property is specified on the Element Props Form - the form schema is validated and default values are applied. Previously the form schema was set up to set all boolean values to "false" values, which is not what we want.

Before:

https://user-images.githubusercontent.com/74900868/234360828-c2026965-1b92-4f27-b80e-4c76173124ed.mov

After:

https://user-images.githubusercontent.com/74900868/234360160-3cf532e0-5331-4802-b57b-4020dd0b3e2b.mov

## Related Issue(s)

Fixes

- #2508

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -13,4 +13,6 @@
 
 <!-- Please link the issue being fixed so it gets closed when this is merged -->
 
-Fixes #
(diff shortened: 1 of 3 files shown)


webberwang:
👍

PR #2553, opened 2023-04-27
fix: error when element deleted after css update

### LINKED ISSUE #2530: Error deleting element
## Current Behavior
Error when deleting an element after modifying its CSS. The element is deleted though. 

## Video Showing Bug

https://user-images.githubusercontent.com/48331678/234017269-d89479da-2946-42ae-ac5e-431b64041b38.mov

vladyslav-polishchuk:
## Description

The error was thrown when an element was deleted and the `ElementCssEditor` component tried to send an update request for the same element on unmount. Also, many redundant requests were sent each time `ElementCssEditor` is opened.
This PR fixes redundant requests, some re-rendering issues and fixes css.cy test, and removes it from ignore list.

Before:

https://user-images.githubusercontent.com/74900868/234806203-37b5c24c-347a-43b6-b6a5-a52b43c65df3.mov

After:

https://user-images.githubusercontent.com/74900868/234805772-3b5aa703-08be-44b3-b402-c166a10c4fd3.mov

## Related Issue(s)

Fixes #2530


webberwang (changes requested):
Instead of using useState and ref to check for change in object, we could consider something more native.

https://immutable-js.com/docs/v4.3.0/ is a great library, and it creates a new object instead of altering the old object when updating (all objects are immutable).

This way we can use https://immutable-js.com/docs/v4.3.0/is()/ check to see if something has changed

Issue #2558, opened 2023-04-28
GuiCSS is reset after page is refreshed


melhamin:
## Video Showing Bug

https://user-images.githubusercontent.com/48331678/235131345-95d64611-c01a-4f3d-b4ac-c6803e86a6de.mov

PR #2563, opened 2023-04-28
refactor: reduce hooks usage

### LINKED ISSUE #2530: Error deleting element

vladyslav-polishchuk:
## Description

Reduce the number of hooks used in `ElementCssEditor` from 18 to just 5.

A bit different approach was used: all the changes were immediately applied from the editor to the `element` model, instead of saving it to the component state. And then either by debouncer or on component unmount these changes are saved to DB (depending on what happens earlier)

## Related Issue(s)

Fixes #2530

### DIFF
--- a/libs/frontend/abstract/core/src/domain/element/element.model.interface.ts
+++ b/libs/frontend/abstract/core/src/domain/element/element.model.interface.ts
@@ -106,6 +106,7 @@ export interface IElement
   detachAsFirstChild(): void
   detachFromParent(): void
   executePropTransformJs(props: IPropData): IPropData
+  setCustomCss(css: string): void
   setFirstChild(firstChild: Ref<IElement>): void
   setName(name: string): void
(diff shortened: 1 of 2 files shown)


melhamin:
It seems that issue #2558 is not fixed here. Could you fix it here? so we can close the issue with this pr?

https://user-images.githubusercontent.com/48331678/235206394-c28860f0-0fba-4e66-93bb-5397abb9b34f.mov

vladyslav-polishchuk:
> It seems that issue #2558 is not fixed here. Could you fix it here? so we can close the issue (and its pr #2559) with this pr?
> 
>  Screen.Recording.2023-04-28.at.19.43.13.mov

Will have a look, thanks!

PR #2574, opened 2023-05-01
fix: missing labels on array types of interface types

### LINKED ISSUE #2540: Array type of InterfaceType field does not show labels
## Current Behavior

For the AntDesignMenu, when you try to expand the items field in the props form, there are no labels for the items.

Based from its labels from the types table, there should be a field name for each
image

But no field names are displayed in the props form
image

jethrocabaluna:
## Description

This PR fixes the labels not showing on the interface types when they are inside an array type prop

| Before | After |
|-----------|---------|
| image |  image |
| image | image |

## Related Issue(s)

Fixes #2540

### DIFF
--- a/libs/frontend/presentation/view/components/form/fields/WrappedListField.tsx
+++ b/libs/frontend/presentation/view/components/form/fields/WrappedListField.tsx
@@ -4,8 +4,15 @@ import { connectField } from 'uniforms'
 import type { ListFieldProps } from 'uniforms-antd'
 import { ListField, wrapField } from 'uniforms-antd'
 
+// `label` here needs to be null so that if the array's items has nested fields (e.g. interface type)
+// their labels will be displayed. This is based on the uniforms documentation for the `label` prop
+// https://uniforms.tools/docs/api-fields/#common-props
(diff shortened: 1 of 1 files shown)


libs/frontend/presentation/view/components/form/fields/WrappedListField.tsx
```diff
@@ -4,8 +4,15 @@ import { connectField } from 'uniforms'
 import type { ListFieldProps } from 'uniforms-antd'
 import { ListField, wrapField } from 'uniforms-antd'
 
+// `label` here needs to be null so that if the array's items has nested fields (e.g. interface type)
+// their labels will be displayed. This is based on the uniforms documentation for the `label` prop
+// https://uniforms.tools/docs/api-fields/#common-props
+// We have a custom component that will render the label for the array field
 const WrappedListFieldInternal = (props: ListFieldProps) =>
```
webberwang, line 11: Can't we do `(props: Omit<ListFieldProps, 'onReset'>)`?

webberwang (changes requested):

PR #2672, opened 2023-05-24
fix: create atom schema issue with type unique validation

volodymyryankovskyi:
Close #2672


webberwang:
The atom type should be unique

webberwang (changes requested):
The atom type should be unique, the goal is to give the user a more user friendly validation error message

PR #2677, opened 2023-05-25
feat: loading external packages dynamically

### LINKED ISSUE #2662: Creating a component from external resources
## Description

The idea is for a user to be able to import a component from external npm resources.

So instead of importing it ourselves by writing the jsx for it then making it re-usable to everyone, we can make the users to import re-usable react components from an external npm packages.

For example is if a user wants to add a component from the [google-map-react](https://github.com/google-map-react/google-map-react), the user should be able to and it will be rendered in the builder when added as a new element.

Will have to do more research on how this can be implemented

jethrocabaluna:
## Description

This is just an initial implementation, more like a POC. We can improve this more later and test more commonly used external react components. 

The external components are loaded dynamically via injecting a script tag to import from a cdn url used for the custom atom.
- added new fields for the atom model
  - externalCssSource
  - externalJsSource
  - externalSourceType
    - this is for `type` that is "ExternalComponent", and should be unique
- the logic for dynamically injecting the external css and js is on the `add` method in atomService

Good things about this implementation is
- we don't have to install all of the external react components that all our users want to use.
- we don't have to limit users regarding how many external react components they want to use
  - but we can still limit if it affects performance due to having too many requests for all the cdns

Some things to consider further with this implementation
- right now we can use the cdn from https://www.jsdelivr.com/, but what if it becomes offline, or cdn doens't work?
  - should we download the content of the external cdn and store it, for example in s3?
- how can we make it easier for the user to build the API for the external component?
  - add a field on the creation of atom that accepts a json with the component's props and their types then automatically create the API based on it?
  - ask chatGPT to make a json of the component's props and their types then automatically create the API based on it?
- how do we validate the cdn url that the user will enter? or how do we make sure it exports a default react component?
  - scan the downloaded content? (might be hard as the content is minimized or compressed)
- maybe instead of users copy-pasting a cdn link, we can have a custom field that searches for external react components
  - might be hard to implement to filter for react components only
  - also what if a user has their own compiled react component
- how do we make sure the user won't be able to import a malicious content?
  - this should be safer with this implementation though since it will only be on the client side, our backend will only save the cdn links

references:
- https://www.ngdevelop.tech/loading-external-libraries-from-cdn-in-angular-application/
- https://itnext.io/dynamically-import-esm-modules-from-a-cdn-5a6f741e2a1c

## Video or Image

Trying [react-calendar](https://github.com/wojtekmaj/react-calendar) and [awesome-react-calculator](https://github.com/avinashsivaraman/awesome-react-calculator)

https://github.com/codelab-app/platform/assets/27695022/d95d2ad9-5577-43e7-bbb6-9f6f3e0665b7

<br />

Trying [google-map-react](https://github.com/google-map-react/google-map-react)

https://github.com/codelab-app/platform/assets/27695022/dbd505de-b215-4998-9cc9-448b230f5805

<br />

Trying [react-calendar-timeline](https://github.com/namespace-ee/react-calendar-timeline/)

https://github.com/codelab-app/platform/assets/27695022/ffde4701-0b59-4146-b823-99a1db508678

## Related Issue(s)

Fixes #2662


webberwang:
very nice feature

PR #2682, opened 2023-05-27
fix: styling after migration to antd5

### LINKED ISSUE #2521: UI issues after migration to AntD 5 version
## Video Showing Bug

https://user-images.githubusercontent.com/74900868/233794711-757e040a-7a2c-4c33-9d88-0bbc9cc4b37e.mov

vladyslav-polishchuk:
## Description
- Fix sizing and positioning of builder tabs
- Apply temporal fix for form fields positioning, since the uniforms-antd package is not updated yet to support ant5 and it is unknown when they will do updates so that form renders properly with antd5

Before             |  After
Screenshot 2023-05-27 at 23 38 44|Screenshot 2023-05-27 at 23 37 54
Screenshot 2023-05-27 at 23 39 34|Screenshot 2023-05-27 at 23 36 36
Screenshot 2023-05-27 at 23 40 30|Screenshot 2023-05-27 at 23 36 12

## Related Issue(s)

Fixes #2521

### DIFF
--- a/apps/platform/pages/_app.tsx
+++ b/apps/platform/pages/_app.tsx
@@ -3,6 +3,7 @@
 import 'react-quill/dist/quill.snow.css'
 // This stylesheet is used to override some of the default Quill editor's styles.
 import '../src/styles/quill.snow.override.css'
+import '../src/styles/antd.global.fix.css'
 import 'react-grid-layout/css/styles.css'
 import 'react-resizable/css/styles.css'
 // apply fix for common css problems:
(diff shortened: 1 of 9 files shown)


webberwang (changes requested):
Shouldn't we apply these at the global css level as opposed to individual components?

could imagine some file like `antd-v5-fix.css`

vladyslav-polishchuk:
> Shouldn't we apply these at the global css level as opposed to individual components?
> 
> could imagine some file like `antd-v5-fix.css`

Makes sense, will try this approach

PR #2691, opened 2023-05-31
refactor: use item names in browser URL instead of IDs

### LINKED ISSUE #2684: Use item names instead of IDs in browser URL bar
To make URLs more compact and meaningful - try to use item names in URL.
For example instead of 
`http://127.0.0.1:3000/apps/3ac439ef-49bd-47d0-b78e-865ac82f2717/pages/4d1ff961-ce76-4aee-a591-1da2743cd618`
try to have
`http://127.0.0.1:3000/apps/MyWebsite/pages/Home/builder`

vladyslav-polishchuk:
## Description

App pages are now accessible at 
`http://127.0.0.1:3000/apps/gm.vpolishchuk/App/pages` 
instead of 
`http://127.0.0.1:3000/apps/3ac439ef-49bd-47d0-b78e-865ac82f2717/pages`. 
Builder is accessible at 
`http://127.0.0.1:3000/apps/gm.vpolishchuk/App/pages/Home/builder?explorerPaneKey=explorer`
instead of
`http://127.0.0.1:3000/apps/3ac439ef-49bd-47d0-b78e-865ac82f2717/pages/4d1ff961-ce76-4aee-a591-1da2743cd618/builder?explorerPaneKey=explorer`

Fixes #2684


webberwang (changes requested):
should use slug instead of name?

vladyslav-polishchuk:
> should use slug instead of name?

In the current schema, this does not look possible. Since `slug` is not a DB field and it is just a custom resolver on apps and pages - I can't filter apps/pages by it.
For example, having URL like this `/apps/some-app-slug/pages`, I can extract app slug `some-app-slug` but I can't then filter apps by that value, I can filter only by `_compoundName`, but slugs are irreversible (meaning I can't reliably get origin app name from the slug) so I can't compute `_compoundName` and filter by it.

To make it possible to get apps/pages by slug I see a couple of options (but not actually like any of them):
- make slug a DB field instead of the resolver. Then we can filter by slugs, but this way we store in the DB redundant value
- create a resolver for apps, so that we can filter on the backend by the slug without the need to add it as a field to DB, this may require a lot of work for this feature which may not be worth it
- get all the apps for the current user and on the client apply filtering by slug, this way the performance will be reduced if the user has many apps

@webberwang  Any other ideas, or should we stick to IDs in the URL and not do anything?

webberwang:
> 

`App._compoundName` is created by `auth0ID-appName`, the `appName` is present in URL and `auth0ID` is available to use, so in frontend we can re-constructor and filter by `_compoundName`

This doesn't require any resolvers, we could search using `_compoundName`

This could be at the Next.js middleware level

vladyslav-polishchuk:
> @vladyslav-polishchuk
> 
> `App._compoundName` is created by `auth0ID-appName`, the `appName` is present in URL and `auth0ID` is available to use, so in frontend we can re-construct and filter by `_compoundName`
> 
> This doesn't require any resolvers, we could search using `_compoundName`
> 
> This could be at the Next.js middleware level

@webberwang 
Currently, it works like you described. I take `appName` from the URL, take user `auth0Id`, then construct `_compoundName` and filter by it. But you suggested using slugs instead of names in the URL, so I will have only app/page `slug` and `auth0Id`, based on this I can't construct `_compoundName` (sluggified name is not reliably reversible, meaning I can get slug from the name, but can't get the name from slug). Or maybe I misunderstood your idea, could you clarify a little bit more

webberwang:
@vladyslav-polishchuk oh I got a bit mixed up, how come slug can't be reliably converted to name?

webberwang:
nice! url looks much cleaner

PR #2713, opened 2023-06-12
feat: use cui tree for store

### LINKED ISSUE #2707: Use CuiTree for State and Actions in the Explorer Sidebar

migrate the current implementation of State and Actions to a new version using CuiTree matching the mockups.

The new version should support

- Vieweing the type of the field as well as the name as primary and secondary titles.
- Showing the subfield of `InterfaceType` fields.

#### Current

image

#### Target

image

mohasarc:
## Related Issue(s)

Fixes #2707


libs/frontend/domain/type/src/store/type.service.ts
```diff
@@ -320,4 +322,50 @@ export class TypeService
 
     return type
   })
+
+  /**
+   * Decides whether loading the type is necessary or not.
+   * loading is required if the type of any of its descendant
+   * types are not available in the cache.
+   */
+  shouldLoadType(typeId: string) {
```
webberwang, line 331: could we perhaps use our custom `@TTLCache()`

webberwang (changes requested):

PR #2725, opened 2023-06-17
fix: cannot use UnionType inside ArrayType

### LINKED ISSUE #2663: Cannot use UnionType
## Current Behavior
Cannot use `UnionType` inside `ArrayType`.
Here `Test Font` type is a union type of two other `InterfaceType`s.

## Video Showing Bug

https://github.com/codelab-app/platform/assets/48331678/99d35641-008e-4ca2-9e1a-88979368a823

melhamin:
## Related Issue(s)

Fixes #2663

### DIFF
--- a/libs/frontend/domain/type/src/interface-form/fields/select-union-type-value/SelectUnionTypeValue.tsx
+++ b/libs/frontend/domain/type/src/interface-form/fields/select-union-type-value/SelectUnionTypeValue.tsx
@@ -47,8 +47,11 @@ export const SelectUnionTypeValue = (props: SelectUnionTypeValueProps) => {
   const { name } = props
   const [fieldProps, context] = useField(name, props)
   const oneOf = fieldProps.field.oneOf
-  const typeFieldName = `${name}.type`
-  const valueFieldName = `${name}.value`
+  // For the nested fields i.e. UnionType inside ArrayType, the name seems to be empty
+  // this causes the field not be validated correctly against the schema
+  // hence `type` would is valid not "".type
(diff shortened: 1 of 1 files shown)


libs/frontend/domain/type/src/interface-form/fields/select-union-type-value/SelectUnionTypeValue.tsx
```diff
@@ -47,8 +47,11 @@ export const SelectUnionTypeValue = (props: SelectUnionTypeValueProps) => {
   const { name } = props
   const [fieldProps, context] = useField(name, props)
   const oneOf = fieldProps.field.oneOf
-  const typeFieldName = `${name}.type`
-  const valueFieldName = `${name}.value`
+  // For the nested fields i.e. UnionType inside ArrayType, the name seems to be empty
+  // this causes the field not be validated correctly against the schema
+  // hence `type` would is valid not "".type
+  const typeFieldName = name ? `${name}.type` : 'type'
```
webberwang, line 53: is there a way to add the name for nested? feels solution should be closer to where the UnionType creates the field type name

melhamin: I don't like this solution either tbh, the problem is that we don't pass the field type name manually, it's generated internally by uniforms from the JsoneSchema created for the type.

I inspected the uniform's source code but couldn't figure out what could be the reason for this behavior.

If you think it's too hacky I will keep looking more into it.

PR #2791, opened 2023-07-09
fix: caching decorators after TS version change

### LINKED ISSUE #2638: Refactor atom repo to use cache ttl
## Description

This is an example of cache ttl implemented in the backend, we could do something similar for the frontend

atom_service_ts_—_platform

vladyslav-polishchuk:
## Description
Fix caching decorators to match the newer TS version, as a result, they became more compact and a bit simpler.
Also removed the `modelFlow` decorator from API methods, since they do not alter store.

## Related Issue(s)

Fixes #2638


webberwang:
this looks great! let's follow this pattern

PR #2794, opened 2023-07-10
fix: load descendant page elements on app load

### LINKED ISSUE #2782: Delete app error
## Description of Bug

When deleting app, it says elementRef is not defined. But refreshing seemed to fix it, so the cache was off.

## How to Reproduce

Difficult to reproduce

vladyslav-polishchuk:
## Description

The issue was that `loadAppsWithNestedPreviews` did not load descendant elements to the store and when the app was deleted - an error was thrown on an attempt to iterate over all page descendant elements because it could not resolve refs.
As a solution - `loadAppsWithNestedPreviews` is updated to load all elements to store, to keep the state valid and avoid script errors about unresolved refs

## Video or Image

Stable steps to reproduce issue

https://github.com/codelab-app/platform/assets/74900868/6873fc5a-d7ff-49ad-9b53-1a1534e8367e

After the fix

https://github.com/codelab-app/platform/assets/74900868/3cde88d2-8cb9-4f92-82bc-040c56dc4e5a

## Related Issue(s)

Fixes #2782

### DIFF
--- a/libs/frontend/domain/app/src/store/app.service.ts
+++ b/libs/frontend/domain/app/src/store/app.service.ts
@@ -179,7 +179,12 @@ export class AppService
        * Pages
        */
       appData.pages.forEach((pageData) => {
-        this.elementService.add(pageData.rootElement)
+        const pageElements = [
+          pageData.rootElement,
+          ...pageData.rootElement.descendantElements,
+        ]
+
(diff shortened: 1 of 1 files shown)


vladyslav-polishchuk:
For some reason, videos are not uploaded to GitHub at the moment. I will try one more time a bit later

PR #2798, opened 2023-07-12
2789 pages page is not fully functional

### LINKED ISSUE #2789: "Pages" page is not fully functional
## How to Reproduce

1. Navigate to "pages" page (http://127.0.0.1:3000/apps/{username}/{app_name}/pages) 
2. See that there is no top toolbar exists

We can either add a missing toolbar for that page, or alternatively - remove the page completely, since it is not accessible from the UI, and is accessible by manually typing the URL only.

Screenshot 2023-07-08 at 12 35 31

amrmuhamedd:
## Description
remove index form pages to make URL only from ui

## Related Issue(s)

Fixes #2789


webberwang (changes requested):
Solution should be to redirect to _app page

webberwang (changes requested):
on your system the yarn is incorrect

try `nx build tools-workspace && yarn`

amrmuhamedd:
OK I will fix this thanks

PR #2825, opened 2023-07-19
feat: state sharing between pages

### LINKED ISSUE #2811: Implement state sharing between _app and regular pages
## Description

Currently, it is not possible to create a state variable or action that is shared between pages and can be reused. This is because each page now has an independent store.
Need to implement ability to define common state fields and actions on _app pages and share this actions with other pages

As a result, the user should be able to define the state field/action on _app page and be able to access/execute it on any other page

jethrocabaluna:
## Description

State from the provider page can be accessed via `rootState.[propName]`. The state from the provider page will be included on evaluating the runtime props of an element.

The actions from the provider page are also included in the options to choose from in the SelectActions fields in a regular page.

Also updated the inspector tab in the explorer to show the state values from the local store, component store, and provider store so users can see all the state values it can use

Added e2e tests for testing that the state created from the provider page is accessible in a regular page.

## Video or Image

Using the `products` state from the provider as the child mapper prop in a regular page.

https://github.com/codelab-app/platform/assets/27695022/5332e43f-bbc1-4073-ade2-b1ba14125764

<br />

Updated inspector to show states from the provider and component as well

https://github.com/codelab-app/platform/assets/27695022/a27c5ef5-2dba-4c0e-ad65-b4d6dc161aaa

<br />

Using an action from the provider as an onclick handler in a regular page

https://github.com/codelab-app/platform/assets/27695022/bedfe417-4a14-4761-8515-bf6bf8988b51

## Related Issue(s)

Fixes #2811

### DIFF
--- a/apps/platform-e2e/src/e2e/builder.cy.ts
+++ b/apps/platform-e2e/src/e2e/builder.cy.ts
@@ -364,7 +364,7 @@ describe('Element Child Mapper', () => {
   })
 })
 
-describe.only('State variables sharing between pages', () => {
+describe('State variables sharing between pages', () => {
   before(() => {
     cy.resetDatabase()
     loginSession()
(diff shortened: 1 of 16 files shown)


webberwang:
nice work!

PR #2843, opened 2023-07-25
feat: in-app routing with next/link

### LINKED ISSUE #2839: in-app routing functionality
## Description
If the user has a link element in his app, and the link's href is something like `/test-page`, the user will expect to go to the `/test-page` of his app. But whats happening right now is it will route to our actual platform so the user will not be able to navigate around his app in the builder.

https://github.com/codelab-app/platform/assets/27695022/2276db90-0c63-45f9-8421-060d17e00482

As of now, to be able to do routing in the app is via using a regular link and providing the complete url that navigates within the app.

But the issue with that is it will do a complete reload of the page. It could be slow, and the current state values will be wiped.

We need something like in [this mock e-commerce website](https://demo.vercel.store/search) where if a product is clicked, it will navigate to another page without needing to reload.

This should work on the deployed app, and while on the builder as well.

jethrocabaluna:
## Description

This is for enabling routing within the builder and adding the NextLink component that can be used for routing
- [x] routing with static urls
- [x] e2e tests with static routes
- [x] routing with dynamic urls
- [x] e2e tests with dynamic routes

The dynamic url segments is also stored in the renderer and it can be used as additional props that can be used for evaluating the expressions. This allows fetching with dynamic url _e.g. if the url of the page is `/products/:productId`, and routed with `/products/t-shirt`, the productId can be accessed as {{url.productId}}_  -  check the video about dynamic routing on how the `url.productId` is used in the rest api action

## Video or Image

Routing in builder

https://github.com/codelab-app/platform/assets/27695022/bfc8772b-af75-47fb-a432-e0d0175a7f5d

<br />

Dynamic routes in builder mode

https://github.com/codelab-app/platform/assets/27695022/4b7ab263-b800-468a-b048-e93214e7ec75

<br />

Dynamic routes in preview mode

https://github.com/codelab-app/platform/assets/27695022/9c35121c-01d3-4c95-bef1-fbc2373f5735

## Related Issue(s)

Fixes #2839


webberwang (changes requested):
Is it possible to use https://nextjs.org/docs/app/building-your-application/routing/middleware?

PR #2848, opened 2023-07-26
fix: controlled components

### LINKED ISSUE #2834: Can't create controlled input component
## Description of Bug

Sometimes there is a need to create a controlled input element. It is when we have some state variable, we use this state variable as a `value` for the input element, and we also assign `onChange` handler to that input that updates the state variable.

A use case for this could be the following: I am having input element and button, and when the user presses the button the value of input should be sent to the server (it is like posting a message in messenger). I need a controlled input in this case in order to be able to access input value without going to DOM, and in order to clear input value once the message was sent.
Here is my use case example:
Screenshot 2023-07-23 at 09 16 05

## How to Reproduce

1. Create an input element
2. Create string state variable
3. Create an action that updates the string variable
4. Assign state variable to input `value` property
5. Assign action to input `onChange` property
6. Try to edit input

As a result - state recieves updated, but the input loses focus after entering 1 symbol to it (probable it gets re-rendered or something).

https://github.com/codelab-app/platform/assets/74900868/a8d62465-3bfc-4d6c-8e38-200c5dcb6e31

vladyslav-polishchuk:
## Description

The entire user component tree was producing new nodes on each re-render. This happened because we dynamically wrapped all elements with `styled()`. This is not how the styled-components were intended to use, because `styled()` produces each time new component (impossible to use `useMemo` for it because then error is thrown about changed rendered hooks count). This was the reason why it was impossible to use controlled components (the input loses focus on each new key pressed because new input is created on each state update). Now the styled component is created only once, and using `as` property we pass to it which component we want to render and by `css` prop we control set of css rules to be applied.
Also, to avoid script errors and allow React to effectively patch user elements - added `key` attribute for each `ElementWrapper` component.

## Video or Image

https://github.com/codelab-app/platform/assets/74900868/f5976e96-34a5-4c37-b44c-e9039a7a5a0a

A practical example of using the controlled component:

https://github.com/codelab-app/platform/assets/74900868/b3c7661a-29bb-4e2d-bfa8-20cd87c3b5f5

## Related Issue(s)

Fixes #2834

### DIFF
--- a/libs/frontend/domain/renderer/src/element/element-wrapper.ts
+++ b/libs/frontend/domain/renderer/src/element/element-wrapper.ts
@@ -16,7 +16,7 @@ import React, { useEffect } from 'react'
 import { ErrorBoundary } from 'react-error-boundary'
 import { shouldRenderElement } from '../utils'
 import { mapOutput } from '../utils/render-output-utils'
-import { getStyledComponent } from './get-styled-components'
+import { renderComponentWithStyles } from './get-styled-components'
 import { extractValidProps, getReactComponent } from './wrapper.utils'
 
 export interface ElementWrapperProps {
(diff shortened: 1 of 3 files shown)


webberwang:
oh thanks for that I didn't realize it would cause re-render, this would be great for performance as well

Issue #1871, opened 2022-09-25
Atom parent/child validation


webberwang:
## Description

Some components can only have certain components to work. For example, `Form.Item` goes under `Form`

## Use Case

Would allow component filtering to make it easier to select valid components.

## Suggested Implementation

We've created a field called allowedChildren to add the relationship here

ky213:
@webberwang 
in the case of an atom where all kinds of child atoms are allowed like for instance the "Body" the user will also be able to select for example "AnchorLink" atom which will throw an error immediately.
This why I suggest that we follow the approach of "atom.requiredParent" rather than "atom.allowedChildren". In this case, if the user selects an atom with a missing required parent in the builder we just show him an alert.
What do you think?

melhamin:
@webberwang 
We already have `requiredParents` and we do validations for it as well. Why do we still need `allowedChildren` field? I belive we can achieve the desired behavior with `requiredParents` field. 

https://github.com/codelab-app/platform/assets/48331678/fa1f2cba-8a87-4aa5-9a7d-d9161556b3c1

Also, the potential issue mentioned here won't be fixed with `allowedChildren`.


> @webberwang in the case of an atom where all kinds of child atoms are allowed like for instance the "Body" the user will also be able to select for example "AnchorLink" atom which will throw an error immediately. This why I suggest that we follow the approach of "atom.requiredParent" rather than "atom.allowedChildren". In this case, if the user selects an atom with a missing required parent in the builder we just show him an alert. What do you think?

nettohadi:
We ended up using the `requiredParents` to achieve the desired behavior. Thanks everyone for the inputs.

PR #2857, opened 2023-07-28
feat: add atom parent and child validation

nettohadi:
## Description

Some atoms can only under certain atoms. For example, Form.Item goes under Form. We are using the `requiredParents` field to validate if a child atom can go under a parent atom. Below are four cases that we're handling:

1. Only display the atoms if their required parents include the parent 

[LOOM DEMO](https://www.loom.com/share/c13139bd66a64fa1b143a77b6f5d2ed2)

2. Only display the atoms if the required parents of the parent doesn't include the atom 

[LOOM DEMO](https://www.loom.com/share/c7e5c8e42dcf4a69984ab2ce1f17a09f)

3. Only display the atoms without required parents 

[LOOM DEMO](https://www.loom.com/share/47b820a95cc34396aae0c8d4d929daaa)

4. Only display the required parents of the atoms in the parent element dropdown 

[LOOM DEMO](https://www.loom.com/share/ed345a1f236041c4b3d312a7d9a0175b)

## Related Issue(s)

Fixes [#1871 Atom parent and child validation](https://github.com/codelab-app/platform/issues/1871)

### DIFF
--- a/.yarnrc.yml
+++ b/.yarnrc.yml
@@ -10,6 +10,6 @@ plugins:
   - path: .yarn/plugins/@yarnpkg/plugin-after-install.cjs
     spec: 'https://raw.githubusercontent.com/mhassan1/yarn-plugin-after-install/v0.4.0/bundles/@yarnpkg/plugin-after-install.js'
 
-afterInstall: ./scripts/yarn/after-install.sh
+# afterInstall: ./scripts/yarn/after-install.sh
 
 yarnPath: .yarn/releases/yarn-3.5.1.cjs
--- a/dist/libs/tools/workspace/package.json
+++ b/dist/libs/tools/workspace/package.json
@@ -1,5 +1,5 @@
 {
-  "name": "tools-workspace",
+  "name": "@codelab/tools-workspace",
   "version": "0.0.1",
   "type": "commonjs",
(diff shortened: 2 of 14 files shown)


webberwang:
Case #2 was the HtmlAtom added as required parent as well?

nettohadi:
> Case #2 was the HtmlAtom added as required parent as well?

No, it wasn't.

webberwang:
should it show up there?

nettohadi:
> should it show up there?

Do we have HtmlAtom? If yes, I couldn't find it.

nettohadi:
Did you mean `HtmlForm`, @webberwang? No, I didn't add the `HtmlForm` as the required parent of the `AntDesignFormItem`. That was why the `HtmlForm` showed up in the dropdown.

webberwang:
> Did you mean `HtmlForm`, @webberwang? No, I didn't add the `HtmlForm` as the required parent of the `AntDesignFormItem`. That was why the `HtmlForm` showed up in the dropdown.

Oh right it's showing the AntDesignForm isn't a children, I suppose FormItem here doesn't have allowedChildren added which is why the rest is showing up. But that is not related to the point here

PR #2858, opened 2023-07-28
2831 define app pages order when displayed list of pages

### LINKED ISSUE #2831: Define app pages order when displayed list of pages

Define some order in which pages will be displayed. Random order looks strange.

Maybe we can use system pages on top, and then regular pages sorted alphabetically under the system ones?

Currently it looks like this
Screenshot 2023-07-22 at 12 27 51

The suggested list for this case could be
_app
404
500
About
Blog
Guestbook
Home

@webberwang what do you think? Does this make sense?

CreativeDev0809:
## Description

Fixed the sort pages of apps with this order: _app, normal pages, 404, 500

## Related Issue(s)

Fixes #2831

### DIFF
--- a/libs/frontend/domain/app/src/store/app.service.ts
+++ b/libs/frontend/domain/app/src/store/app.service.ts
@@ -336,17 +336,23 @@ export class AppService
      * Sort pages for app. Order is app, custom pages, 404, 500
      */
 
+    enum PageType {
+      App = '_app',
+      NotFound = '404',
+      InternalError = '500',
+    }
+
     const _pages: Array<BuilderPageFragment> = appData.pages
     _pages.sort((a, b) => {
-      if (a.name === '_app') {
+      if (a.name === PageType.App) {
         return -1
       }
 
-      if (a.name === '404') {
+      if (a.name === PageType.NotFound) {
         return 1
       }
 
-      if (a.name === '500') {
+      if (a.name === PageType.InternalError) {
         return 1
       }
 
--- a/libs/frontend/domain/app/src/store/app.service.ts
+++ b/libs/frontend/domain/app/src/store/app.service.ts
@@ -337,14 +337,14 @@ export class AppService
      */
 
     enum PageType {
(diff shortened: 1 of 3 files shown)


libs/frontend/domain/app/src/store/app.service.ts
```diff
@@ -325,10 +326,33 @@ export class AppService
...
       return undefined
     }
 
+    /**
+     * Sort pages for app. Order is app, custom pages, 404, 500
+     */
+
+    const _pages: Array<BuilderPageFragment> = appData.pages
+    _pages.sort((a, b) => {
+      if (a.name === '_app') {
```
webberwang, line 341: we should use enum here

libs/frontend/domain/page/src/use-cases/get-pages/PageList.tsx
```diff
@@ -9,6 +9,8 @@ interface PageListProps {
 }
 
 export const PageList = observer(({ app }: PageListProps) => {
+  console.log(app.pages)
```
webberwang, line 12: remove the log

webberwang (changes requested):

libs/frontend/domain/app/src/store/app.service.ts
```diff
@@ -325,10 +326,39 @@ export class AppService
...
+
     if (!appData) {
       return undefined
     }
 
+    /**
+     * Sort pages for app. Order is app, custom pages, 404, 500
+     */
+
+    enum PageType {
```
webberwang, line 339: the enum already exists in our project, it's inside shared abstract folder I believe

webberwang (changes requested):

libs/frontend/domain/app/src/store/app.service.ts
CreativeDev0809: Fixed using compare with page kind(used enum IPageKind in abstract/core). Please check it.

PR #2860, opened 2023-07-28
fix: remove page index

### LINKED ISSUE #2789: "Pages" page is not fully functional

amrmuhamedd:
## Description
remove index form pages to make URL only from ui

## Related Issue(s)

Fixes #2789


webberwang:
@mohasarc does it make sense to remove this page?

mohasarc:
Yeah, I can't think of any usage for it.

PR #2879, opened 2023-08-04
feat: implement CuiPopover and use it for create/update forms

### LINKED ISSUE #2870: Cui Popover Design
## Description

Currently, we have 2 main ways to show forms, either an inline form or a modal. The issue with the inline form is that it obscures the primary sidebar views (usually tree views). on the other hand, the issue with a modal is that it is very detached from the overall UI.  The popover, a tab that opens to the right of the primary sidebar, would match the sidebar in terms of overall style and would feel like a continuation of it.

We can basically transition all the forms to use the popover except for delete confirmation modals.

### Design

Group 226

### Mockups

Group 221

Group 222

Group 223

Group 224

Related #2643

mohasarc:
## Description

This PR replaces all the inline forms and create/update modals with popover forms.

current implementation kept the use of the formService to mainly set the data to be used by each form. in addition cui context is introduced.

the whole app is wrapped with `CuiProvider` and `useCui` would provide some useful methods to control cui components. for example, in this PR access to the `popover` store is provided, using which we can open/close the popovers.

The main logic is that each popover has a key, using which we can open/close that particular popover from anywhere in the app (usually adjacent to where we open the forms).

## Video or Image

**Before**

https://github.com/codelab-app/platform/assets/51242349/38b217c3-e056-43be-b832-06a65090e1b3

**After**

https://github.com/codelab-app/platform/assets/51242349/6491e235-c2e2-4022-ba15-3b5c5cf893de

## Related Issue(s)

Fixes #2870


webberwang:
nice work!

PR #2880, opened 2023-08-04
2831 define app pages order when displayed list of pages

CreativeDev0809:
## Description

Fixed issue delay while sorting pages.

## Related Issue(s)

#2831 Define app pages order when displayed list of pages 

Fixes #
Fixed issue delay while sorting pages.


webberwang:
doesn't seem like a good approach to sort in both places

webberwang (changes requested):
doesn't seem like a good approach to sort in both places

vladyslav-polishchuk:
I see that sorting rule is still the same as before fix. But before the order was not as expected.
The desired order is: `_app` page, normal pages then `404` and `500`.

Here is how ChatGPT suggest to do this:

```
let arr = ['hello', 'world', '_app', '404', '500', 'example', 'more', 'strings'];

arr.sort((a, b) => {
    // '_app' should always be first
    if (a === '_app') return -1;
    if (b === '_app') return 1;

    // '404' should always be last but one
    if (a === '404') return b === '500' ? -1 : 1;
    if (b === '404') return a === '500' ? 1 : -1;

    // '500' should always be last
    if (a === '500') return 1;
    if (b === '500') return -1;

    // All other strings should be sorted in their natural order
    return a.localeCompare(b);
});
```

You can try this

CreativeDev0809:
> doesn't seem like a good approach to sort in both places

I also think so.
But as I got, it fetch app data(including page data) and shows page list using that, and then it fetches page data again and shows it.
So there was an issue seems like delay in sort.
Please let me know your thought about this.
Thanks.

vladyslav-polishchuk:
> > doesn't seem like a good approach to sort in both places
> 
> I also think so. But as I got, it fetch app data(including page data) and shows page list using that, and then it fetches page data again and shows it. So there was an issue seems like delay in sort. Please let me know your thought about this. Thanks.

There are actually several options: creating backend resolver that would return pages sorted, or we can define a sorting function in one place and reuse it, so that we at least do not duplicate the code. Alternatively, maybe we can define getter for pages property that would return them in sorted order. @webberwang, what do you think is better?

Issue #2597, opened 2023-05-05
Filter not used metadata for production websites


vladyslav-polishchuk:
## What Feature Can Be Enhanced?
Currently, production user websites load information about the app/page and ALL metadata: interfaces, fields, atoms,etc without checking if this information is required for specific page.
This approach has two negative effects on production websites performance:
- big page size, even for simple pages. next.js also shows warning about this in console:
`Warning: data for page "/[domain]/[...page]" (path "/codelab-demo.online/page") is 341 kB which exceeds the threshold of 128 kB, this amount of data can reduce performance.`
This leads to delay of displaying page to the user since need to download large file with data first.
- all these interfaces are very expensive to load into the store. Lighthouse shows low performance score for a very simple page with single text element, and the interfaces and fields take 2 seconds of blocking time to just be loaded.
<img width="1728" alt="Screenshot 2023-05-05 at 18 29 50" src="https://user-images.githubusercontent.com/74900868/236515049-4fab005b-686c-4b84-b0de-cc8bb38f9df2.png">

webberwang:
Maybe interfaces isn't need for runtime just build time, similar to how typescript works.

vladyslav-polishchuk:
> @vladyslav-polishchuk how to reproduce this?

Discussed in DM

vladyslav-polishchuk:
> Maybe interfaces isn't need for runtime just build time, similar to how typescript works.

If this is possible - this may be a significant performance improvement for user pages

nettohadi:
# How to Run the Websites Locally

### Assign a Domain to Your App
If you haven't already, start by creating an app. Then, follow the steps below to assign a domain to it. In this example, we'll use "codelab.id" as our domain:

1. Find your app's card and click on the three-dot menu icon located at its top-right corner.
2. From the dropdown list, select "Domains."
3. In the top-right corner of the Domains section, click on the "+" icon to add a new domain.
4. Enter your desired domain name in the "Name" field. Ensure it is URL-friendly.
5. Hit the "Create Domain" button.
6. If you encounter an error message stating "Domain misconfigured because it's already assigned to another project," you can safely ignore it.

### Pointing the Newly-Created Domain to Your Local IP Address
After setting up your domain, the subsequent step involves directing it towards your local IP address. This mirrors the DNS resolution process on our production server. Instructions vary based on the operating system in use. Below are the detailed steps for Mac, Windows, and Linux:

#### Mac:
[...]

PR #2889, opened 2023-08-09
feat: filter out unused metadata for production website

nettohadi:
## Description

Currently, production websites fetch extensive details about the app/page, including metadata like interfaces, fields, atoms, etc., without determining if this data is essential for the page in view. Moreover, it retrieves metadata for all pages, even when we're viewing just one. This leads to two significant performance issues:

- **Increased Page Size**: Even simple pages are bloated with unnecessary data. This is flagged by Next.js, which warns that the large data size exceeds the optimal threshold, potentially degrading performance. 

- **Expensive Loading Into the Store**: Loading all these interfaces is resource-intensive. Lighthouse performance metrics reveal that even a simple page with a single text element suffers due to the time taken to load the interfaces and fields. This consumes 2 seconds of blocking time, negatively impacting the overall user experience.

Screenshot 2023-05-05 at 18 29 50

### Changes made in this pull request include:
- Simplify and modify the [middleware](https://github.com/codelab-app/platform/pull/2889/files#diff-93f11f0d6c60cef7ea42c07ac05f9d18e7e83bc16528d4208fc8e184b3de4ddb) to prevent incorrect redirection.

Screen Shot 2023-08-05 at 08 41 49

- Change the websites port to port 80 because port 3000 is already used by the platform.
- Create a new GraphQL query to not include interfaces, owner, and tags for the production page.
- Load only the metadata relevant to the current page and the `_app` page. Since the `_app` page wraps other pages and offers shared data, its metadata is essential.
- Add missing `_compoundName` and `auth0Id` to [this query](https://github.com/codelab-app/platform/pull/2889/files#diff-21a42ed73108ee920c15274463215eeaf072484527ca5b7379fbc0b9507722dc) because they are needed [here](https://github.com/codelab-app/platform/blob/54b6d371c14490bc91d0856e3c370590c162ee84/apps/websites/pages/%5Bdomain%5D/%5B...page%5D.tsx#L70).
-  Since we've removed the interfaces from the production metadata, we can't retrieve the default prop values from them. To address this, we ensure that empty element props are consistently assigned their default values at build time. This is already implemented when creating an element. What remains is ensuring the same behavior when updating the prop values. Here's the [commit](https://github.com/codelab-app/platform/pull/2889/commits/96b15e08ace4c0966f76b497f61830fddf423134).

## Result
By only loading the metadata relevant to the current page and the `_app` page, and by removing the interfaces, we see a decrease in page size. However, the Lighthouse score doesn't show a significant improvement. The following images illustrate the results before and after the change..

### Before:
Screen Shot 2023-08-11 at 15 25 09

Screen Shot 2023-08-09 at 22 03 01

### After:
Screen Shot 2023-08-11 at 15 24 45

#### Load only the metadata relevant to the current page and the `_app` page
Screen Shot 2023-08-09 at 21 58 02

#### Interfaces are removed from the metadata
Screen Shot 2023-08-11 at 15 27 44

## Related Issue(s)

Fixes [#2597](https://github.com/codelab-app/platform/issues/2597)


nettohadi:
I am still trying to remove the interfaces. This may take awhile.

webberwang:
> I am still trying to remove the interfaces. This might take some time.

Yeah sure it is a complex task

apps/websites/middleware.ts
```diff
@@ -1,27 +1,25 @@
...
-    isVercelDomain ||
-    isInternal ||
-    isPublic ||
-    isFavicon ||
-    !hostname
-  ) {
-    return NextResponse.next()
-  }
+const middleware = async (req: NextRequest) => {
+  const hostname = req.headers.get('host')?.replace(/https?:\/\//g, '')
```
webberwang, line 19: could convert to new URL and get the hostname

libs/frontend/abstract/core/src/domain/atom/atom.fragment.graphql
```diff
@@ -27,6 +27,32 @@ fragment Atom on Atom {
   externalSourceType
 }
 
+fragment AtomLite on Atom {
```
webberwang, line 30: I used `Preview` for some other queries, but I think both Preview and Lite could be more descriptive. The vocab should relate to where it is used. The ProductionPage naming is much better

libs/frontend/abstract/core/src/domain/element/element.fragment.graphql
```diff
@@ -57,3 +57,62 @@ fragment Element on Element {
   }
   propTransformationJs
 }
+
+fragment ElementLite on Element {
```
webberwang, line 61: Same here

libs/frontend/domain/atom/src/store/atom.model.ts
```diff
@@ -42,7 +42,8 @@ const create = ({
   type,
 }: IAtomDTO) => {
   return new Atom({
-    api: typeRef<IInterfaceType>(api.id),
+    // eslint-disable-next-line @typescript-eslint/no-unnecessary-condition
+    api: typeRef<IInterfaceType>(api?.id ?? ''),
```
webberwang, line 46: typeRef to an empty string doesn't really make sense (even though it solves the lint)

libs/frontend/domain/store/src/store/models/store.model.ts
```diff
@@ -40,7 +40,8 @@ const create = ({
 }: IStoreDTO): IStore =>
   new Store({
     actions: actions.map((action) => actionRef(action.id)),
-    api: typeRef(api.id) as Ref<IInterfaceType>,
+    // eslint-disable-next-line @typescript-eslint/no-unnecessary-condition
+    api: typeRef(api?.id ?? '') as Ref<IInterfaceType>,
```
webberwang, line 44: same

apps/websites/middleware.ts
nettohadi: Hmm, I've realized that I don't need to use `replace(/https?:\/\//g, '')` because it's already stripped of `http://`.

libs/frontend/abstract/core/src/domain/atom/atom.fragment.graphql
nettohadi: Which name makes more sense: `AtomWithoutInterfaces` or `AtomProduction`?

libs/frontend/domain/atom/src/store/atom.model.ts
nettohadi: Will fix this.

nettohadi: Fixed [here](https://github.com/codelab-app/platform/pull/2889/commits/60f1dec84339dad763dafce6efe0de40ec5a3891).I ended up modifying many more files to handle the possibility of an undefined `api`.

nettohadi: Looks like we still have errors. I am fixing them now.

nettohadi: Never mind. it was a false alarm.

libs/frontend/domain/store/src/store/models/store.model.ts
nettohadi: Fixed [here](https://github.com/codelab-app/platform/pull/2889/commits/69d471146ed7b1f708004400b1089cded4803625).

nettohadi:
@webberwang - As discussed earlier, we'll ensure that empty element props are always assigned their default values. This eliminates the need to fetch the default values from interfaces in production. In essence, we'll synchronize the props with their respective default values during build time. Here is the [commit](https://github.com/codelab-app/platform/pull/2889/commits/96b15e08ace4c0966f76b497f61830fddf423134).

libs/backend/domain/atom/src/repository/atom.repo.ts
```diff
@@ -48,7 +48,7 @@ export class AtomRepository extends AbstractRepository<
             ...atom
           }) => ({
             ...atom,
-            api: connectNodeId(api.id),
```
webberwang, line 51: connectNodeId can take undefined as arg

libs/frontend/domain/element/src/use-cases/element/update-element-props/UpdateElementPropsForm.tsx
```diff
@@ -69,11 +73,17 @@ export const UpdateElementPropsForm = observer<UpdateElementPropsFormProps>(
     }, [apiId])
 
     const onSubmit = (data: IPropData) => {
```
webberwang, line 75: should move this inside the propService.update, or create a new propService.updateWithDefaultValuesApplied or something if the update function needs the original function elsewhere

webberwang (changes requested):

libs/frontend/domain/element/src/use-cases/element/update-element-props/UpdateElementPropsForm.tsx
nettohadi: Updated [here](https://github.com/codelab-app/platform/pull/2889/commits/169147986a95234669f297dec6b1491a30df1a87).

libs/backend/domain/atom/src/repository/atom.repo.ts
nettohadi: Updated [here](https://github.com/codelab-app/platform/pull/2889/commits/8c40acedee78b81bca7639e61311826459e76dad).

libs/frontend/abstract/core/src/domain/atom/atom.fragment.graphql
nettohadi: Updated [here](https://github.com/codelab-app/platform/pull/2889/commits/be75d3e9d04ec99d7b0084c36f5be40cc09b5ece).

libs/frontend/abstract/core/src/domain/element/element.fragment.graphql
nettohadi: Updated [here](https://github.com/codelab-app/platform/pull/2889/commits/be75d3e9d04ec99d7b0084c36f5be40cc09b5ece).

libs/backend/domain/store/src/model/store.model.ts
```diff
@@ -14,7 +14,7 @@ export class Store implements IStoreDTO {
   name: string
 
   constructor({ api, id, name }: IStoreDTO) {
-    this.api = api
+    this.api = api ?? { id: '' }
```
webberwang, line 17: makes more sense to have `api | undefined` or `api | null`

libs/frontend/domain/atom/src/use-cases/get-atoms/AtomsTreeItem.tsx
```diff
@@ -43,10 +43,15 @@ export const AtomsTreeItem = ({ data }: AtomsTreeItemProps) => {
   }
 
   const onAddField = () => {
-    const interfaceId = node.api.id
-    const interfaceRef = typeRef(interfaceId) as Ref<IInterfaceType>
+    const interfaceId = node.api?.id
 
-    fieldService.createModal.open(interfaceRef)
+    const interfaceRef = interfaceId
+      ? (typeRef(interfaceId) as Ref<IInterfaceType>)
```
webberwang, line 49: I believe typeRef is generic and requires no casting

libs/shared/data/test/src/atoms.data.ts
```diff
@@ -36,7 +36,7 @@ export const createAtomsApiData = (
 ): Array<IInterfaceTypeDTO> =>
   atomsData.map((atom) => ({
     fields: [],
-    id: atom.api.id,
+    id: atom.api?.id ?? '',
```
webberwang, line 39: since is required here, we should throw an error in the map callback if the id doesn't exist

webberwang (changes requested):

libs/frontend/domain/atom/src/use-cases/get-atoms/AtomsTreeItem.tsx
nettohadi: Fixed [here](https://github.com/codelab-app/platform/pull/2889/commits/6db96dca731f159cfed007b1dd3c845ac99e534e).

libs/shared/data/test/src/atoms.data.ts
nettohadi: Updated [here](https://github.com/codelab-app/platform/pull/2889/commits/6922102e6af31c4a2f0c5066c4ebe479ccdc906f).

libs/backend/domain/store/src/model/store.model.ts
nettohadi: Updated [here](https://github.com/codelab-app/platform/pull/2889/commits/49743b00eede912d1716486f769fd7edf0b86217).

libs/frontend/abstract/core/src/domain/atom/atom.fragment.graphql
```diff
@@ -27,6 +27,32 @@ fragment Atom on Atom {
   externalSourceType
 }
 
+fragment ProductionAtom on Atom {
+  icon
+  id
+  name
+  type
+  owner {
```
vladyslav-polishchuk, line 35: Do you think it is possible to omit `owner` and `tags` for production?

nettohadi: Looking into this.

nettohadi: I've tested this, and everything appears to work fine in production website without the `owner` and `tags`. Any concern if we remove them, @webberwang?

webberwang: seems okay without it

webberwang (changes requested):
some merge conflicts after master was updated thanks

nettohadi:
> some merge conflicts after master was updated thanks

Will resolve the conflicts.

PR #2903, opened 2023-08-17
fix: script error on adding renderType to Body

### LINKED ISSUE #2893: Error setting renderType of body
## How to Reproduce

1. Try setting the `renderType` of the body of a page

https://github.com/codelab-app/platform/assets/48331678/9ae3410d-a02b-4438-857a-f19b6cada649

vladyslav-polishchuk:
## Description
Fixed 2 issues in this PR:
1. The problem that the interface and corresponding fields were not loaded for Body at all (for renderType)
2. The problem when renderType interface and fields were not loaded on element *update*: when the user changes renderType to an atom that was not used on the current page before, this results in the same script error because could not resolve interface type for that atom

## Video or Image

https://github.com/codelab-app/platform/assets/74900868/c40fa740-4a39-4bd9-8e12-747a25a5eb08

## Related Issue(s)

Fixes #2893

### DIFF
--- a/libs/frontend/domain/element/src/store/element.service.ts
+++ b/libs/frontend/domain/element/src/store/element.service.ts
@@ -130,7 +130,7 @@ export class ElementService
    */
   @modelFlow
   @transaction
-  loadRenderTypeInterface = _async(function* (
+  private loadRenderTypeInterface = _async(function* (
     this: ElementService,
     data: ICreateElementData,
   ) {
--- a/libs/frontend/domain/element/src/store/element.service.ts
+++ b/libs/frontend/domain/element/src/store/element.service.ts
@@ -124,11 +124,16 @@ export class ElementService
   }
 
   /**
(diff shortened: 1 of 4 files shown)


libs/frontend/domain/element/src/store/element.service.ts
```diff
@@ -124,11 +124,16 @@ export class ElementService
...
   /**
-   * We need a separate create function for element trees
+   * If an element is created or updated with renderType atom or component
+   * that is not used yet anywhere in the current page,
+   * this will load its interface types and its fields
    */
   @modelFlow
   @transaction
-  create = _async(function* (this: ElementService, data: ICreateElementData) {
+  loadRenderTypeInterface = _async(function* (
```
webberwang, line 133: make private here

webberwang (changes requested):

PR #2936, opened 2023-08-24
feat: allow users to use container queries

### LINKED ISSUE #2924: Think of approach to provide better UX for creating mobile/tablet layouts
## Description

Currently, to update styles for specific screen size (to show/hide content, change paddings/margins, etc) media queries can be used, like this:
```
@media (max-width: 600px) {
  display: block;
}
```

But the problem with this approach is that you can't test this in builder. Media queries are executed in context of entire window, not a specific DOM node. So for example if user app has media queries for mobile - they would not be applied in builder:

https://github.com/codelab-app/platform/assets/74900868/e19dcffd-c37f-40f0-a4b0-4a8862b3315a

Need to think of some way to provide user with ability to create layouts for PC/Tablet/Mobile with a better UX than always switching to page preview and resizing entire window to desired size.

vladyslav-polishchuk:
## Description

This is useful when creating pages that target multiple screen sizes.
Now user is able to see how the app will look on the desired screen size in the builder immediately

## Video or Image

https://github.com/codelab-app/platform/assets/74900868/d8302c76-8608-4a9b-9656-72c8d559a2f8

## Related Issue(s)

Fixes #2924


vladyslav-polishchuk:
@webberwang, do you think this is enough, or should we implement some UI for creating styles for a specific screen size?
Since this would be pretty hard to discover that users may use container queries by having only a single textarea field for element CSS.

nettohadi:
> @webberwang, do you think this is enough, or should we implement some UI for creating styles for a specific screen size? Since this would be pretty hard to discover that users may use container queries by having only a single textarea field for element CSS.

In webflow, brizy, or other website builders, the responsive css is automatically generated when we update the css in specific breakpoint.

webberwang:
> 

I do think we need some preset breakpoints within the builder, that is used to resize the builder width, along with the css breakpoints.


The CSS could be tabbed by screen size perhaps, ideas are welcome

PR #2941, opened 2023-08-26
fix: persist store state on NextLink navigation

### LINKED ISSUE #2910: Persist _app page state between navigation using NextLink atom

If there are some state variables exist on _app page (called rootState), this state is lost each time navigation occurs. Even if not use native `a` atom, but also if to use new `NextLink` atom. Full page reload does not occur, but the rootState still loses all values.
Need to find a way to persist _app page state

https://github.com/codelab-app/platform/assets/74900868/8de260d5-abd6-4d37-9536-75bf28035b3f

vladyslav-polishchuk:
## Description

Do not re-create store and store.state when not necessary.

## Video or Image

Uploading Screen Recording 2023-08-26 at 10.11.14.mov…

## Related Issue(s)

Fixes #2910

### DIFF
--- a/libs/frontend/domain/store/src/store/models/store.model.ts
+++ b/libs/frontend/domain/store/src/store/models/store.model.ts
@@ -118,11 +118,17 @@ export class Store
     return getRenderService(this)
   }
 
+  private cachedState: Nullable<object> = null
+
   @computed
   get state() {
+    if (this.cachedState) {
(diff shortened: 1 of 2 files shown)


webberwang:
video was cut off but it's okay

vladyslav-polishchuk:
Uploaded video, to have a history just in case we would need to revise the solution

Issue #2946, opened 2023-08-28
Some CSS Properties are not Applied as Expected


nettohadi:
## Description of Bug
I’ve identified a bug in the way we use Styled Components. While playing with CSS flexbox in the builder, I observed that the CSS wasn’t applied as expected. Upon further inspection, I found that the CSS properties remained in camelCase. I suspect that Styled Components didn’t convert them to kebab-case because the CSS input was already in string format.

## How to Reproduce

1. Select any element
2. Update any CSS properties, especially the multi-word properties like: `border-width`, `border-style`, `align-items`, etc
3. When you open the devtool, you will see the css properties are in camelCase

## Video/Screenshot Showing Bug
![Screen Shot 2023-08-25 at 09 49 09](https://github.com/codelab-app/platform/assets/32300655/37220b27-726c-43e6-a7e2-7864de3ac9f8)

PR #2947, opened 2023-08-28
fix: some css properties are not applied as expected

nettohadi:
## Description

Some CSS properties aren't applied as expected because they remain in camelCase. I suspect that Styled Components didn't convert them to kebab-case due to the CSS input being in string format. This PR converts properties from camelCase to kebab-case before feeding them into Styled Components.

## Video or Image

**Before:**
Screen Shot 2023-08-25 at 09 49 09

**After:**
Screen Shot 2023-08-28 at 11 24 33

## Related Issue(s)

Fixes [#2946](https://github.com/codelab-app/platform/issues/2946)

### DIFF
--- a/libs/frontend/domain/renderer/src/utils/camel-case-to-kebab-case.ts
+++ b/libs/frontend/domain/renderer/src/utils/camel-case-to-kebab-case.ts
@@ -1,0 +1,3 @@
+export const camelCaseToKebabCase = (word: string) => {
+  return word.replace(/([A-Z])/g, '-$1').toLowerCase()
+}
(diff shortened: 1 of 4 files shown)


libs/frontend/domain/renderer/src/utils/camel-case-to-kebab-case.ts
```diff
@@ -0,0 +1,3 @@
+export const camelCaseToKebabCase = (word?: string) => {
```
webberwang, line 1: I would put this in `libs/shared/utils/src/transform/string`

webberwang (changes requested):

libs/frontend/domain/renderer/src/utils/camel-case-to-kebab-case.ts
nettohadi: Moved [here](https://github.com/codelab-app/platform/pull/2947/commits/d370e5eec8235f1982d38bb2a5628a80d9192c7f).

PR #2950, opened 2023-08-28
fix: allow refs for elements with atoms only

### LINKED ISSUE #2949: Restrict refs usage for elements with atoms only
Refs should only be attached to elements with atoms since they are the only elements that has an actual dom that is rendered.

assimhabeek:
## Related Issue(s)

Fixes #2949


assimhabeek:
We will change strategy. to making element name unique

Issue #2955, opened 2023-08-29
Make Elements in the Builder View Draggable and Droppable


nettohadi:
## Description
We will enhance the builder view by making all elements more drag-and-drop friendly. This will include the following functionalities (ordered by priorities):

- [x] Make elements selectable and hoverable
https://github.com/codelab-app/platform/pull/3036

https://github.com/codelab-app/platform/assets/32300655/a88e2264-55ae-4fc2-b93d-f526e47411b5

- [x] Make elements draggable and droppable
https://github.com/codelab-app/platform/pull/3020

https://github.com/codelab-app/platform/assets/32300655/ecc22309-ab35-4471-b942-90369633d84f

- [x] Display margins and paddings
https://github.com/codelab-app/platform/pull/3034

![paddings-margins](https://github.com/codelab-app/platform/assets/32300655/e97853fd-b7d1-4f5a-81c7-37192c0c6e5e)

nettohadi:
For the past two days, I've been grappling with an elusive bug related to element selection. As of now, I'm still not entirely sure about the best way to resolve it.

As most of you are aware, when we select an element in the tree view, a corresponding black square (referred to as `clickOverlay` in the code) appears around the element in the builder view. However, when we modify the CSS properties related to an element's size or margins, the `clickOverlay` doesn't update accordingly. This leads to a situation where the `clickOverlay` no longer aligns correctly with the element. This issue seems to arise because we're not adequately tracking changes to the element's size. But, this is not bug I've been struggling with.


https://github.com/codelab-app/platform/assets/32300655/61d4cc1e-922a-45bb-a220-4d295a990f8e

So, I've made some modifications to the code to track any changes in the CSS, ensuring that the `clickOverlay` updates each time the CSS is updated. Initial tests showed that it worked as expected. However, when I tested it with the `AntDesignButton`, the issue resurfaced. This is the specific bug that has been giving me trouble. You will see in the following video that the bug only happened in `AntDesignButton`.



https://github.com/codelab-app/platform/assets/32300655/046ccba3-6188-4c8d-87f6-9b3fd39b5918



[...]

PR #2969, opened 2023-09-02
fix: ClickOverlay is not aligned with target element when css is updated

nettohadi:
## Description

For the past two days, I've been grappling with an elusive bug related to element selection. As of now, I'm still not entirely sure about the best way to resolve it.

As most of you are aware, when we select an element in the tree view, a corresponding black square (referred to as `clickOverlay` in the code) appears around the element in the builder view. However, when we modify the CSS properties related to an element's size or margins, the `clickOverlay` doesn't update accordingly. This leads to a situation where the `clickOverlay` no longer aligns correctly with the element. This issue seems to arise because we're not adequately tracking changes to the element's size. But, this is not bug I've been struggling with.

https://github.com/codelab-app/platform/assets/32300655/61d4cc1e-922a-45bb-a220-4d295a990f8e

So, I've made some modifications to the code to track any changes in the CSS, ensuring that the `clickOverlay` updates each time the CSS is updated. Initial tests showed that it worked as expected. However, when I tested it with the `AntDesignButton`, the issue resurfaced. This is the specific bug that has been giving me trouble. You will see in the following video that the bug only happened in `AntDesignButton`.

https://github.com/codelab-app/platform/assets/32300655/046ccba3-6188-4c8d-87f6-9b3fd39b5918

So, what exactly is happening here? To properly align the clickOverlay with the element, we obtain the element's DOMRect values using `getBoundingClientRect` and then apply those values to the `clickOverlay`. In the case of `AntDesignButton`, the issue arises because we are fetching these values before the browser has finished rendering the button. As a result, we end up with outdated values. To confirm this, I added a 200ms delay before retrieving the values, and doing so produced the correct values. You can see it the following video.

https://github.com/codelab-app/platform/assets/32300655/90cbcf97-1131-46fc-bab2-fc35b4011718

For now, adding a delay somewhat solves the bug. However, I can't help but feel that this isn't the right solution and that introducing a delay makes things fragile. So far, I've only observed this bug in `AntDesignButton`, but it could potentially occur in more complex Atoms as well. I could use your advice here, @webberwang.

## Related Issue(s)

Fixes [#2955](https://github.com/codelab-app/platform/issues/2955)

### DIFF
--- a/libs/frontend/domain/builder/src/sections/overlay-toolbar/BuilderClickOverlay.tsx
+++ b/libs/frontend/domain/builder/src/sections/overlay-toolbar/BuilderClickOverlay.tsx
@@ -79,6 +79,11 @@ export const BuilderClickOverlay = observer<{
   return createPortal(
     <ClickOverlay
       content={content}
+      dependencies={[
+        selectedNode.current.guiCss,
+        selectedNode.current.customCss,
+        selectedNode.current.props.current.values,
+      ]}
       element={element}
       renderContainer={renderContainerRef.current}
     />,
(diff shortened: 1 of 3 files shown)


webberwang:
nice work on the write up, we can use this solution for the time being

PR #2988, opened 2023-09-07
fix: cannot type into NextLink atom

### LINKED ISSUE #2985: Impossible to set inner text on Link atom
## How to Reproduce

1. Create a NextLink atom on the page
2. Try to set a text label for the link

The user is being redirected to the link target page

https://github.com/codelab-app/platform/assets/74900868/d1c4486a-00e3-40d4-9c13-ff16f152157b

melhamin:
## Related Issue(s)

Fixes #2985


webberwang:
should we disable behavior in builder mode? how does other builder handle this?

melhamin:
> should we disable behavior in builder mode? how does other builder handle this?

In other builders also the links are not navigatable in builder mode. So I assume the default behavior is disabled.

webberwang:
> > should we disable behavior in builder mode? how does other builder handle this?
> 
> In other builders also the links are not navigatable in builder mode. So I assume the default behavior is disabled.

We should follow this strategy, could you come up with a list of features that would be disabled

PR #2992, opened 2023-09-08
fix: css string values converted to kebab-case

### LINKED ISSUE #2991: Some CSS styles are not applied
## Description of Bug
I have noticed the issue with `font-family` not being applied, other stylings might not work too. The issue seems to be with converting camelCase to kebab-case [here](https://github.com/codelab-app/platform/pull/2947). Only keys should be converted. Currently, the values are also converted.
## How to Reproduce

1. Create an element
2. Set the element's font family

Screenshot 2023-09-08 at 08 20 43

melhamin:
## Description
We had to convert camelCase in `css` string to kebab-case to be correctly applied. This conversion was done for both keys and values, which caused some of the styles i.e. font-family to not work if the name is two words. 

So now we only convert the keys to kebab-case. 

## Related Issue(s)

Fixes #2991


nettohadi:
Nice. Thanks for fixing this. This will only target words that are directly followed by a colon.

PR #3017, opened 2023-09-14
fix: update overlay on position/breakpoint change

### LINKED ISSUE #3016: Element selection issues
## How to Reproduce

1. Element selection is not updated on breakpoint change

https://github.com/codelab-app/platform/assets/74900868/4a0f3240-3ace-40ab-b770-82aedb63decd

2. Element Selection is not updated on elements re-ordering

https://github.com/codelab-app/platform/assets/74900868/42eb3efd-1320-4ab4-b88d-eae6ecb42b42

vladyslav-polishchuk:
## Description

Whenever selected element is moved on the page, or whenever breakpoint changes - update the selected element overlay

## Video or Image

https://github.com/codelab-app/platform/assets/74900868/de970b2b-bb46-43af-bca9-471704ffece6

## Related Issue(s)

Fixes #3016

### DIFF
--- a/libs/frontend/domain/builder/src/sections/overlay-toolbar/BuilderClickOverlay.tsx
+++ b/libs/frontend/domain/builder/src/sections/overlay-toolbar/BuilderClickOverlay.tsx
@@ -59,7 +59,7 @@ export const BuilderClickOverlay = observer<{
   }
 
   const content = (
-    <StyledOverlayContainer className="click-overlay-toolbar">
+    <StyledOverlayContainer>
       <StyledSpan>{selectedNode.current.name}</StyledSpan>
       <StyledOverlayButtonGroup>
         <Button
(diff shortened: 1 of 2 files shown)


nettohadi:
Hi @vladyslav-polishchuk, first of all, thanks for fixing this. You also addressed the issue that occurs when the element is moved, which is something I had noticed as well 👍. 

I tested it locally and your changes handled all the cases except for when the builder view is manually resized. Below is a quick video to show that.

https://github.com/codelab-app/platform/assets/32300655/39fa3480-b8cf-4ebd-91c8-a45f982f0e7f

Actually, we do have `useResizeObserver` in `ClickOverlay` that can be used to track size changes. However, it didn't work because we didn't include the `height` & `width` to the dependency list.

Screen Shot 2023-09-18 at 16 39 46

By including the width and height to dependency list, the `ClickOverlay` is properly aligned when breakpoint and size change.


https://github.com/codelab-app/platform/assets/32300655/d6d2a2a3-20fa-479f-8ef3-e17cb0471e2f

Let me know what you think about this. I could be missing something.

nettohadi:
Also, I believe we don't need `useResizeObserver({ ref: element })` anymore because we already track the element props.

vladyslav-polishchuk:
> Hi @vladyslav-polishchuk, first of all, thanks for fixing this. You also addressed the issue that occurs when the element is moved, which is something I had noticed as well 👍.
> 
> I tested it locally and your changes handled all the cases except for when the builder view was manually resized. Below is a quick video to show that.
> 
>  Screen.Recording.2023-09-18.at.16.37.45.mov 
> Actually, we do have `useResizeObserver` in `ClickOverlay` that can be used to track size changes. However, it didn't work because we didn't include the `height` & `width` to the dependency list.
> 
> Screen Shot 2023-09-18 at 16 39 46
> 
> By including the width and height to dependency list, the `ClickOverlay` is properly aligned when breakpoint and size change.
> 
>  Screen.Recording.2023-09-18.at.16.41.10.mov 
> Let me know what you think about this. I could be missing something.

@nettohadi , thanks for your comment! After we finish with nestjs refactoring, I will return to this PR and try to apply your suggestions

PR #3009, opened 2023-09-13
make elements selectable and hoverable within the builder

nettohadi:
## Description

Currently, element selection is done by clicking on the element in the tree view. This is not particularly convenient, as users often prefer to select elements directly in the builder view, where they can see the layout.

This Pull Request aims to enhance user experience by making elements both selectable and hoverable directly within the builder view.

## Video or Image

**Before:**

https://github.com/codelab-app/platform/assets/32300655/661c6e67-d80d-474a-84e4-675e82a45083

**After:**

https://github.com/codelab-app/platform/assets/32300655/a88e2264-55ae-4fc2-b93d-f526e47411b5

## Related Issue(s)

Fixes #2955

PR #3020, opened 2023-09-18
make elements draggable droppable within the builder view

### LINKED ISSUE #2955: Make Elements in the Builder View Draggable and Droppable

nettohadi:
## Description

Currently, dragging and dropping can only be done within the tree view. This is not particularly convenient, as users often prefer to drag and drop elements directly in the builder view, where they can see the layout.

This Pull Request aims to enhance user experience by making elements both draggable and droppable directly within the builder view.

Changes in this PR include:

1. Ability to drag an element and drop it into another element.
2. Ability to drag an element and drop it before or after another element.
3. Displaying a full blue rectangle to indicate that the dragged element will be dropped into the target element.
4. Displaying a partial blue border to indicate that the dragged element will be dropped either before or after the target element.
5. Inline text editing can now only be activated by double-clicking the element. Previously, inline text editing was always active, which made dragging or grabbing an element difficult.
6. The ghost element, which was displayed when dragging an element in the tree view, has now been removed.

## Video or Image

https://github.com/codelab-app/platform/assets/32300655/ecc22309-ab35-4471-b942-90369633d84f

## Related Issue(s)

Fixes #2955


apps/platform/pages/_document.tsx
```diff
@@ -49,6 +49,18 @@ export default class MyDocument extends Document {
...
+              __html: `
+            window.addEventListener('error', event => {
+              if(event.filename.includes('editorjs')) {
+                event.stopImmediatePropagation()
+              }
+            })
+          `,
+            }}
+          ></script>
         </Head>
```
nettohadi, line 64: In case you're wondering why we have this script here. Toggling the `readOnly` in this [line](https://github.com/codelab-app/platform/blob/661bbc7cc4054dc6f20299827ccb7dce871a7d59/libs/frontend/domain/renderer/src/element/text-editor/TextEditor.tsx#L84), caused the editor js to spit some errors that crashed the app. Fortunately, even with these errors, I could still use editor js normally. From reading around the internet, I concluded that these errors were just some `undefined` state not handled correctly. Anyway, we can only wait for editor js team to fix these errors so I figured we could just ignore these errors for now.

libs/frontend/domain/renderer/src/utils/useDragDropHandlers.hook.ts
```diff
@@ -0,0 +1,243 @@
...
+  return true
+  /**  
+   For now, all atoms are droppable.
+   Basically, we will add a new field called "allowedChildren" to the atom schema.
+   If the allowedChildren is empty, then any atoms can be dropped.
+   If the allowedChildren has one or more atoms, then only those atoms can be dropped.
+   We will implement the logic for this function later.
+   Adding a new field to the schema in this PR would make the changes overwhelming
+  */
+}
```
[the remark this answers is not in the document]
nettohadi: But, if you have a suggestion on how to handle this better, do let me know.

apps/platform/pages/_document.tsx
nettohadi: And if you're wondering why the script is placed here, it's because putting it in other locations won't work. Next.js has a built-in script that catches all errors and displays them to users. We aim to intercept these errors before Next.js does.

nettohadi:
This [selection & hover PR](https://github.com/codelab-app/platform/pull/3009) needs to be merged before this PR.

apps/platform/pages/_document.tsx
webberwang: oh i see thanks for the info

libs/frontend/domain/renderer/src/utils/useDragDropHandlers.hook.ts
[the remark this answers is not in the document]
nettohadi: I agree. That is a better idea.

PR #3041, opened 2023-09-27
refactor websites to call platform-api cirectly

### LINKED ISSUE #3040: Refactor websites to call platform-api directly
### The Problem
Currently, in order for the `websites` to run, we need to also run `platform-api` and `platform`. This is not ideal, since `websites` doesn't really need `platform` other than being a proxy to fetch data from `platform-api`. Also, running three node apps plus docker can really slow down development machine. 

### Proposed Solution
Refactor the `websites` app to call `platform-api` directly, bypassing the need for `platform`.

nettohadi:
## Description

Currently, in order for the `websites` to run, we need to also run `platform-api` and `platform`. This is not ideal, since `websites` doesn't really need `platform` other than being a proxy to fetch data from `platform-api`. Also, running three node apps plus docker can really slow down development machine. 

In this PR, we add GraphQL handler (cloned from the `platform`) to `websites`, allowing  direct communication with `platform-api`.

## How to Test

1. Navigate to the root of the `websites` directory and create a new `.env` file.
2. Add the key `NEXT_PUBLIC_PLATFORM_HOST` and assign it the corresponding value from your original `.env` file.
3. Run `platform-api` and `websites` normally
4. Access any page you've created

When you run `websites`, Nx loads the `.env` file located in the `websites` directory. Conversely, when running `platform`, Nx accesses the `.env` file from the project's root directory. This allows us to use the same environment variable with different values depending on the app that is running.

## Related Issue(s)

Fixes #3040


webberwang (changes requested):
Can you add `.env.example`

nettohadi:
> Can you add `.env.example`

Sure, added [here](https://github.com/codelab-app/platform/pull/3041/commits/4f10ab8428289496b342cf1941f3a1f3af4f1ae7).

PR #3042, opened 2023-09-27
feat: restore get server side props

### LINKED ISSUE #3025: Restore getServerSideProps to guard pages
## Description

add `Page.getServerSideSide` which contains a custom code that could be used to protect pages from loading and maybe load data if users wants

assimhabeek:
## Related Issue(s)

Fixes #3025


vladyslav-polishchuk:
@assimhabeek, @webberwang, if I am not mistaken `getServerSideProps` can't be used with static site generation (which is currently how user web apps are rendered in production). So either we would need to change the approach to how we render user apps in production as well (without static generation apps probably going to be slow), or need to find another way to protect pages than using `getServerSideProps`. What do you think?

PR #3064, opened 2023-10-09
chore: splitting renderer lib between application and domain

### LINKED ISSUE #3050: Renderer refactor and fix
## Description

- [ ] Separate to application & domain folder
- [x] Fix specs

jethrocabaluna:
## Domain
- atoms
- render pipes
- typed prop trnasformers
- action-runner model
- runtime props logic
- render service
- renderer model

## Application
- RootRenderer.tsx

## Presentation
- text editor logic and react components
- element-wrapper

## NOTES
- I had to move the dependent React components like element-wrapper into the presentation lib since they are needed in the renderer.model and to not have a cyclic dependency error
- I think renderer.model makes sense to be in domain since it is mainly using mobx cache

## Related Issue(s)

Fixes #3050


webberwang:
Unfortunately couldn't merge this, since I've also made some changes splitting. Will be easier to redo if needed

PR #3074, opened 2023-10-11
fix: component props undefined in props evaluation context

### LINKED ISSUE #3070: componentProps are not available in preview mode
## Description of Bug

issue seems to be caused because we're not adding component's runtime props to the renderer in any mode other than the `ComponentBuilder` one.

image

## How to Reproduce

1. add some component prop with a default value

image

2. use the prop value somewhere in the component (see video bellow)

3. open preview mode

https://github.com/codelab-app/platform/assets/51242349/620593d5-3b59-466a-895d-5a55a536216d

mohasarc:
## Description

Issue used to happen because

1. Component props are not attached to the builder unless it is in a `component-builder` mode
2. no element inside the component other than the root had access to the component props no matter what.

## Video or Image

1. first component testing component props usage in the root element
2. second component testing component props usage in another element

https://github.com/codelab-app/platform/assets/51242349/5db576be-1b2f-4c3e-8cae-2f39f099d9e7

## Related Issue(s)

Fixes #3070

### DIFF
--- a/libs/frontend/application/renderer/src/render.service.ts
+++ b/libs/frontend/application/renderer/src/render.service.ts
@@ -3,11 +3,7 @@ import type {
   IRenderService,
   RendererProps,
 } from '@codelab/frontend/abstract/domain'
-import {
-  componentRef,
-  getRendererId,
-  RendererType,
-} from '@codelab/frontend/abstract/domain'
+import { getRendererId } from '@codelab/frontend/abstract/domain'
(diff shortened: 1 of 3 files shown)


webberwang:
@mohasarc is there a way to unit test this?

mohasarc:
> @mohasarc is there a way to unit test this?

There should be, I'll have a look

PR #3082, opened 2023-10-14
fix: lint, unit and integration jobs on ci

vladyslav-polishchuk:
## Related Issue(s)

Fixes #

### DIFF
--- a/libs/backend/application/admin/src/use-case/seed-framework/import-export.i.spec.ts
+++ b/libs/backend/application/admin/src/use-case/seed-framework/import-export.i.spec.ts
@@ -13,7 +13,6 @@ import fs from 'fs'
 import path from 'path'
 import { AdminController } from '../../admin.application.controller'
 import { AdminApplicationModule } from '../../admin.application.module'
-import { concatenateFileContents } from './compare-directories'
 
 @Module({})
 class Auth0ModuleMock {}
@@ -26,7 +25,7 @@ const currentUser = {
   username: 'Codelab',
 }
 
-jest.setTimeout(300000)
+jest.setTimeout(200000)
 
 const exportPath = path.resolve('./data/export-v2')
 const exportTestPath = path.resolve('./tmp/data/export-v2')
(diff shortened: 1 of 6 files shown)


libs/frontend/application/app/src/use-cases/app-development/app-development.endpoints.graphql
vladyslav-polishchuk, line 1: The `primitiveTypes`, `reactNodeTypes` and `renderPropTypes` already included in the query several lines above (before `resources`)

libs/frontend/application/app/src/use-cases/app-development/app-development.endpoints.graphql
```diff
@@ -21,15 +21,4 @@ query GetAppDevelopment($appCompositeKey: String!, $pageName: String!) {
   resources {
     ...Resource
   }
-  # Load system types
```
vladyslav-polishchuk, line 24: The primitiveTypes, reactNodeTypes and renderPropTypes already included in the query several lines above (before resources)

libs/backend/application/admin/src/use-case/seed-framework/import-export.i.spec.ts
```diff
@@ -26,7 +26,7 @@ const currentUser = {
   username: 'Codelab',
 }
 
-jest.setTimeout(240000)
```
webberwang, line 29: Instead of increasing timeout, we can seed maybe 20 random atoms each time. This means we don't have to seed so much data, and the randomness will likely encompass all atoms over the course of multiple ci runs

webberwang (changes requested):

webberwang (changes requested):
I'm thinking another approach
import-admin-data_command_service_ts_—_codelab__Workspace_

We can use an existing Stage enum, pick test & dev since those are the 2 different scenarios

then similar to `if(baseDataPaths)`, we can set a boolean property on ReadAdminDataService

we then shuffle and limit say some arbitrary hard coded value

vladyslav-polishchuk:
> I'm thinking another approach import-admin-data_command_service_ts_—codelab__Workspace
> 
> We can use an existing Stage enum, pick test & dev since those are the 2 different scenarios
> 
> then similar to `if(baseDataPaths)`, we can set a boolean property on ReadAdminDataService
> 
> we then shuffle and limit say some arbitrary hard coded value

A couple of questions about this approach:

1. There are 4 stages exist: `dev`, `prod`, `test` and `ci`. This logic should be implemented for `ci` only, right? `prod` looks like is used only with terraform commands, `test` is used in commit hooks (probably need to run integration test with all atoms in this case), `dev` is the same, probably better run on all atoms, and the `ci` is the only case we want?

2. Currently the best way to understand the `stage` is CI is 
```
this.readAdminDataService.partiallySeed =
      process.env['NX_TASK_TARGET_CONFIGURATION'] === Stage.CI
```
but this seems a bit odd. Maybe you know better approach?

webberwang:
> > I'm thinking another approach import-admin-data_command_service_ts_—codelab__Workspace
> > We can use an existing Stage enum, pick test & dev since those are the 2 different scenarios
> > then similar to `if(baseDataPaths)`, we can set a boolean property on ReadAdminDataService
> > we then shuffle and limit say some arbitrary hard coded value
> 
> A couple of questions about this approach:
> 
> 1. There are 4 stages exist: `dev`, `prod`, `test` and `ci`. This logic should be implemented for `ci` only, right? `prod` looks like is used only with terraform commands, `test` is used in commit hooks (probably need to run integration test with all atoms in this case), `dev` is the same, probably better run on all atoms, and the `ci` is the only case we want?
> 2. Currently the best way to understand the `stage` is CI is
> 
> ```
> this.readAdminDataService.partiallySeed =
>       process.env['NX_TASK_TARGET_CONFIGURATION'] === Stage.CI
> ```
> 
> but this seems a bit odd. Maybe you know better approach?

Nevermind seems a boolean value of `partialSeed: boolean` or something could work

libs/backend/application/admin/src/use-case/seed-framework/import-export.i.spec.ts
```diff
@@ -66,9 +65,35 @@ describe('Seed, import, & export data', () => {
     await adminController.import({ adminDataPath: exportPath })
     await adminController.export({ adminDataPath: exportTestPath })
 
-    const exportPathContents = concatenateFileContents(exportPath)
-    const exportTestPathContents = concatenateFileContents(exportTestPath)
+    const pathsToCompare = [
```
webberwang, line 68: I think it's more clear if we construct a `Record<ExportFilePath, ImportFilePath>`

Then we transform into `Record<ExportFileContent, ImportFileContent>` to compare them

webberwang (changes requested):

PR #3084, opened 2023-10-15
feat: add test cases for moving element

nettohadi:
## Description

In this PR we add test cases for moving elements.

## Related Issue(s)

Fixes #


libs/frontend/domain/element/src/tests/element.domain.spec.ts
```diff
@@ -253,4 +253,102 @@ describe('Element domain', () => {
...
+      id: v4(),
+      name: 'Second Element',
+      parentElement: secondParentDto,
+    }
+
+    let firstParent: IElementModel
+    let secondParent: IElementModel
+    let firstElement: IElementModel
+    let secondElement: IElementModel
+
```
nettohadi, line 290: Hmm, might need to update the description for better readability.

PR #3095, opened 2023-10-21
3067 refactor the drag and drop within the builder to use dnd-kit

### LINKED ISSUE #3067: Refactor the drag and drop functionality to use dnd-kit
## Description

Currently, the drag and drop functionality within the builder is implemented using vanilla js. We want to [dnd-kit](https://dndkit.com/) instead as an abstraction to reduce the complexity.

nettohadi:
## Description

In this PR, we've refactored the builder's drag and drop functionality to use dnd-kit, moving away from the previous vanilla js implementation. The previous implementation can be found [here](https://github.com/codelab-app/platform/pull/3020).

In various parts of the codebase, I've found traces of dnd-kit. The drag and drop implementation seems nearly complete, just requiring one more step to pull everything together. However, during my review, I discovered some issues with the implementation. Here are the issues I found and the steps I took to address them.

1. **Dragging Interferes with the Click Event**: Dnd-kit uses the onPointerDown event to detect when a user initiates a drag action. Notably, `onPointerDown` triggers before `onClick`, causing a rerender that subsequently cancels out the `onClick` event. This presents a challenge when users simply want to select an element by clicking on it. To address this, I attempted various solutions. Ultimately, I decided to [increase the delay](https://github.com/codelab-app/platform/blob/76699d076ef8599d3426f2e4f244629098f72b4b/libs/frontend/application/builder/src/dnd/useBuilderDnd.hook.ts#L34) that activates dragging.
2. **Box Shadow as a Drop Indicator is Occasionally Invisible**: The previous dnd-kit implementation employed a [box shadow](https://github.com/codelab-app/platform/blob/beddb9e86a991a4aa655d3194d07d9ae7c368473/libs/frontend/application/renderer/src/utils/make-drop-indicator-style.ts#L11) to indicate the intended drop location of the element. While it's an improvement over using a border color, it has its drawbacks. Specifically, the box shadow becomes invisible when the element is aligned with the edge of its container. I experimented with the inset property to cast the shadow inside the element, but this too became obscured when the element had nested children. To resolve this, I reverted to using the BuilderDragDropOverlay, ensuring consistent visibility.
3. **Tracking Pointer Position by Attaching a Mouse Move Event to the Document**: Although this isn't a major concern, it's worth noting. The [earlier approach](https://github.com/codelab-app/platform/blob/1fa0cdd60a41fa23e2152b9960a29e704c301fc6/libs/frontend/application/renderer/src/element/draggable-element.util.ts#L26) involved attaching a mouse move event to the entire document to track and store the pointer's position in a state. While I get the reasoning, I can't help but feel there's a more optimal method. After experimenting with various alternatives (and nearly reverting to the original strategy), I found that we can leverage the [collision detection](https://github.com/codelab-app/platform/blob/76699d076ef8599d3426f2e4f244629098f72b4b/libs/frontend/application/builder/src/dnd/BuilderContext.tsx#L83) to obtain the pointer position, which is then used to [calculate the drag/drop position](https://github.com/codelab-app/platform/blob/76699d076ef8599d3426f2e4f244629098f72b4b/libs/frontend/application/builder/src/dnd/BuilderContext.tsx#L151).

## Related Issue(s)

Fixes #3067


libs/frontend/application/builder/src/dnd/BuilderContext.tsx
```diff
@@ -36,27 +57,173 @@ export const BuilderContext = observer<PropsWithChildren>(({ children }) => {
...
-      <DragOverlay dropAnimation={null}>
-        {draggedElement && draggedElement.data.current?.overlayRenderer()}
+      <DragOverlay
+        className="w-auto"
+        dropAnimation={null}
+        modifiers={[snapCenterToCursor]}
+      >
+        <div className="min-h-[20px] min-w-[70px] max-w-[120px] truncate rounded bg-white p-2 text-center text-[14px]">
+          {sourceAtomName}
+        </div>
```
nettohadi, line 165: You might be wondering why I don't just use `builderService.dragDropData.target.current.atomName` instead of `sourceAtomName` in line 164. Well, I did, and the resulted in infinite loop somehow. Here is related [issue](https://github.com/mobxjs/mobx/issues/3728) in GitHub.

PR #3108, opened 2023-10-29
feat: support ReactNode type for components

### LINKED ISSUE #3104: Support ReactNode type for components
## Description

This prop would be visible in the props view for elements inside a custom component, where any component prop of type `ReactNodeType` can be picked to be rendered as a child of the element.

It would look something like this 

image (6)

mohasarc:
## Description

This PR mainly adds support for binding component props to the children of any of the component elements.

The implementation involves allowing `SelectComponent` field to view the component props of type `ReactNodeType` as part of the options in addition to the default options (the components).

The reason this solution is picked include:

1. This approach utilizes the existing type transformer.
2. Allows for the use of the component props for any atom that accepts `ReactNodeType` for any of it's props (previously limited to components only).

## Video or Image

https://github.com/codelab-app/platform/assets/51242349/e6ac226b-4db2-4e4d-845e-8734842ec4e8

## Related Issue(s)

Fixes #3104

### DIFF
--- a/data/export-v3/admin/atoms/HtmlDiv.json
+++ b/data/export-v3/admin/atoms/HtmlDiv.json
@@ -1,7 +1,24 @@
 {
   "api": {
     "__typename": "InterfaceType",
-    "fields": [],
+    "fields": [
+      {
+        "api": {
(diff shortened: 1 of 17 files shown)


mohasarc:
closing this in favor of #3268

PR #3109, opened 2023-10-30
suppress Twind warning logs

### LINKED ISSUE #3100: Tailwind warning logs
## Description of Bug
There are warning logs in the console when `tailwind` classes is enabled. 
## How to Reproduce

1.
2.

Screenshot 2023-10-24 at 00 30 55

nettohadi:
## Description

In development mode, Twind emits warnings for detected invalid Tailwind class names. This behavior is intentional, serving as a cautionary measure against the use of incorrect class names. However, given that we aren't restricting ourselves solely to Tailwind class names, such warnings in the console are expected.

There is no easy way to turn it off (I wish there was). After reading Twind [codebase](https://github.com/tw-in-js/twind/blob/e6acbea2f48e3c6441e0cf71cd069f48500ca493/packages/twind/src/internal/warn.ts#L18) and [change log](https://github.com/tw-in-js/twind/blob/80affbe27d49ed6176bc49dff4624cc814318da0/packages/core/CHANGELOG.md?plain=1#L99), I decided to add an event listener to suppress it.

## Video or Image

**Before:**
Screen Shot 2023-10-30 at 10 41 59

**After:**
Screen Shot 2023-10-30 at 10 41 26

## Related Issue(s)

Fixes #3100


melhamin:
@nettohadi I had a look into this issue too. I noticed that we are using [`@twind/with-next`](https://twind.style/packages/@twind/with-next). The official [`twind`](twind.dev) website seems to suggest using [@twind/next](https://twind.dev/usage-guides/nextjs.html).

The issue doesn't occur with this package. You may want to have a look into that.  

I am not sure if there's a specific reason for using `@twind/with-next` though.

nettohadi:
Thanks for looking into this, @melhamin. The [twind.dev](https://twind.dev/handbook/introduction.html) is the old website (v0.16). The new official website is [twind.style](https://twind.style/migration#from-twind-v016) and [@twind/with-next](https://twind.style/with-next) is the new way to integrate with Next js. The reason I used `@twind/with-next` is because we're currently using the latest version of Twind. Those warnings was added in v1, so that's why it didn't show up in v0. You can read the change log [here](https://github.com/tw-in-js/twind/blob/80affbe27d49ed6176bc49dff4624cc814318da0/packages/core/CHANGELOG.md?plain=1#L86).

So, you've tried using [@twind/next](https://twind.dev/usage-guides/nextjs.html) with Twind v1 with no issue? We could use the old next plugin, but I think that's not a good idea. What do you think, @melhamin?

melhamin:
@nettohadi I didn't notice that 👀. 

I also don't think using the older version is a good idea.

PR #3117, opened 2023-11-03
test: type transformer specs

### LINKED ISSUE #3103: Cover more cases in type transformer specs
## Description of Bug

- Add one test for each type
- Add react unit testing for type dropdowns (react testing without cypress)

## How to Reproduce

1.
2.

melhamin:
## Description

- [x] Added more specs for the type-transformer
- [x] Use react unit testing, without cypress, for testing components

## Related Issue(s)

Fixes #3103

### DIFF
--- a/libs/backend/abstract/codegen/src/ogm-types.gen.ts
+++ b/libs/backend/abstract/codegen/src/ogm-types.gen.ts
@@ -1846,6 +1846,11 @@ export type AnyType =
   | ReactNodeType
   | EnumType
   | ActionType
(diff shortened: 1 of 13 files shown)


webberwang (changes requested):
looks very good, one thing to note on how we setup the store.

follow this instead https://github.com/codelab-app/platform/blob/da360f07606f0ebef0f758fcf7f0e5253301bb66/libs/frontend/domain/element/src/test/root.test.store.ts#L17

PR #3119, opened 2023-11-04
fix: login into production

vladyslav-polishchuk:
## Description

- fix environment variables for vercel production deployments
- increase `maxDuration` for serverless function on Vercel from 15s to 60s, since requests often rejected by timeout
- fix atoms import: import all atoms without `suggestedChildren` first, and then selectively update atoms that have `suggestedChildren`
- after user logs in - skip `importSystemTypes` and only seed user from request. This speeds up login as well as does not require us to deploy `data/export-v3` folder
- remove redundant request inside of `update` method in tag.repository, since it is not used anymore (previously it was used to get current value of tag `parent` property)

## Related Issue(s)

Fixes #

### DIFF
--- a/apps/platform-api/pages/api/data/[[...index]].ts
+++ b/apps/platform-api/pages/api/data/[[...index]].ts
@@ -39,6 +39,7 @@ export const config = {
      */
     externalResolver: true,
   },
+  maxDuration: 60,
 }
 
 export default handler
--- a/apps/platform-api/pages/api/graphql.ts
+++ b/apps/platform-api/pages/api/graphql.ts
@@ -22,6 +22,7 @@ export const config = {
      */
     externalResolver: true,
   },
+  maxDuration: 60,
 }
 
 export default handler
--- a/apps/platform/pages/api/[[...index]].ts
+++ b/apps/platform/pages/api/[[...index]].ts
@@ -1,3 +1,7 @@
 import { proxyMiddleware } from '@codelab/backend/infra/adapter/graphql'
 
 export default proxyMiddleware
+
+export const config = {
+  maxDuration: 60,
+}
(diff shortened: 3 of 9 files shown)


scripts/vercel/platform-api/install.sh
```diff
@@ -8,3 +8,7 @@ set -x
 cd ../..
 echo $PWD
 yarn install --immutable
+
+# include seed data in vercel deployment
+# so that the build can use it it from admin dashboard
+cp -r data ./apps/platform-api
```
webberwang, line 14: no need to copy right? if we are only running migration locally

webberwang (changes requested):

scripts/vercel/platform-api/install.sh
vladyslav-polishchuk: Currently, `setup-dev` is called after login to populate system data and it fails without `data/export-v3/system/types/system-types.json`. So if we do not need to run migration directly from production UI - we may still need at least this file to be copied so that login command successes.

Or alternatively, we can call `setup-dev` only from local environment, so it is not getting called in production each time any user logs in. What do you think?

vladyslav-polishchuk: Also, If we plan to run seed for production from local environment, do I need to continue work with bull queues? Since from local machine there is no timeout of how long the seed request may be running

webberwang: Sure if we don't need `setup-dev` on production we can call locally only, so we can remove that part.

I hadn't thought of that, seems like we don't need queue for this instance. We can keep whatever code we have and pause on this.

The queue might come in useful in the future though

vladyslav-polishchuk: I will create and cancel the PR with the queue, so that it can be looked up easily through GitHub UI if we decide to continue work on it at some point in the future

vladyslav-polishchuk:
Will retarget to `dev` branch, since there are already fixed ts build errors

PR #3132, opened 2023-11-11
fix: error during export

vladyslav-polishchuk:
## Description

- an error was thrown during atoms export (too many opened DB connections) because all atoms were imported in parallel. Instead import 10 at a time
- allow to specify `selectionSet` in backend repo classes for more effective queries. In future can benefit from this also by updating `exists` method on all repos to fetch only record `id` instead of getting entire record with all relationships

## Related Issue(s)

Fixes #

### DIFF
--- a/libs/backend/application/atom/src/services/atom.application.service.ts
+++ b/libs/backend/application/atom/src/services/atom.application.service.ts
@@ -25,13 +25,14 @@ export class AtomApplicationService {
       selectionSet: `{ id }`,
     })
 
-    return await Promise.all(
-      atomIds.map(
-        async ({ id }) =>
-          await this.commandBus.execute<ExportAtomCommand, IAtomBoundedContext>(
-            new ExportAtomCommand({ id }),
-          ),
-      ),
-    )
+    const exportedAtoms = []
+
+    for (const { id } of atomIds) {
+      exportedAtoms.push(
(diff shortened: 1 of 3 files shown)


libs/backend/application/atom/src/services/atom.application.service.ts
```diff
@@ -20,19 +19,29 @@ export class AtomApplicationService {
...
-          ),
-      ),
-    )
+    // can't import all atoms in parallel,
+    // because the DB errors of too many opened connections (100+)
+    // import 10 atoms at a time
+    const atomChunks = chunk(atomIds, 10)
+    const atoms = []
+
+    for (const atomsChunk of atomChunks) {
```
webberwang, line 35: instead of chunking would it work if we used `for of` instead of `promise.all`?

webberwang (changes requested):

libs/backend/application/atom/src/services/atom.application.service.ts
vladyslav-polishchuk: And use `await` inside, meaning import just one after another? Yes, this also works

webberwang: yeah let's just do that without chunk, seems simpler

PR #3134, opened 2023-11-12
feat: builder dnd

### LINKED ISSUE #3122: Drag & drop functionality
## Description

Our previous dnd was deleted here https://github.com/codelab-app/platform/pull/3055/files#diff-0b00bad046ac925a5be0f510cf8c5815121bf00e2a795933a8de19256cbbc9f9 perhaps unintentionally

A new PR was WIP here https://github.com/codelab-app/platform/pull/3095 and it uses a different strategy. I'm not too sure which way is best, maybe you could take a look

mohasarc:
#### Component Droppables

Components are directly made droppable without an additional container wrapper. The reason for this is to:

- cut the total number of elements to roughly half,
- and eliminate the potential styling issues that would happen because of the intermediate (probably `div`) element.

Note that this design assumes that all components appropriately forward refs.

An exception of this though is the Fragment component. Due to its lack of corresponding HTML element, it's wrapped in a `div`.

#### Drag Knob Addition
A knob has been added to drag builder elements, eliminating the need for handling special cases (like input elements) separately.

#### Drop Indicator Injection
The drop indicator is integrated directly into the HTML of the droppable component. This approach simplifies handling scenarios where the droppable container’s element order is non-standard (e.g., vertical, reversed).

#### Hierarchical Intersection Detection Algorithm

This algorithm leverages parent-child relationships between droppable components as follows:

- When multiple droppables intersect the pointer, the algorithm prioritizes the child farthest from the root for dropping.
- It detects if the draggable container is a parent of the droppable container.
- Allows backtracking to make the draggable's parent container droppable in case the user tries to drop the the container itself or any of its children.

#### Margin Around Droppable Components

Margins prevent child components from obscuring their parent when sizes are similar and ensure space for dropping elements between tightly packed children in a droppable container. These scenarios are illustrated in the image below

image

#### Positioning: Before, After, Within

- Three primary positions are identified: before, after, and within.
- Visually, this translates to five potential drop locations: up, down, left, right, and within.
- A diagonal line from bottom left to top right is used to determine the dragged element's placement, as illustrated in the following image.

image

#### move an element

https://github.com/codelab-app/platform/assets/51242349/9dab9076-3548-4839-93b7-8d45c38ed582

#### create an element

https://github.com/codelab-app/platform/assets/51242349/7a2243f3-2e4e-4f3a-bfe3-1be8886176d0

### Notes

- Handling of void elements is not necessary as the builder already manages errors. This issue is not unique to drag and drop; any changes can be applied at the same time to both drag and drop and tree structures.

https://github.com/codelab-app/platform/assets/51242349/bc49e7dd-23b1-4770-a4c5-59e6ff665a26

#### Drop Indicator Accuracy
Enhance the drop indicator to more precisely represent the final layout post-drop. the issue is shown in the following example:

because the drop indicator is a line, it doesn't show how the element would wrap when the element being dragged is dropped. we could instead make the indicator of the same size as the draggable element. I've tried that quickly, but faced some issues so droped it for now.

https://github.com/codelab-app/platform/assets/51242349/f15b7e5c-0329-45f7-b185-e30e8b90eb4f

#### Drag Handler Accessibility
Improve access to the drag handler by integrating it into the element's hover overlay, enhancing usability.

---------

#### TODO

- [x] implementation of the core functionality of dnd
- [x] integrate dnd to element move
- [x] integrate dnd to element create (drag & drop a component)
- [x] use a different drag overlay for element create other than the component card
- [x] Wrap elements of FragmentComponent type with a div to make them droppable
- [x] disable dnd in production
- [x] ~prevent void elements from being droppable~ (unnecessary)
- [x] enhance hover and click overlays styles
- [x] unit test collision detection algorithm

## Related Issue(s)

Fixes #3122


nettohadi:
Hi @mohasarc, regarding the drop Indicator accuracy, we can try a different approach. Rather than showing a drop indicator, we could insert a copy of dragged element (let's call it "ghost element") directly into the html. This way, we can have an accurate glimpse of final layout post-drop. Just remember to remove the ghost element `onDragEnd`. I believe this will solve the wrap issue.


https://github.com/codelab-app/platform/assets/32300655/cd0240d1-fe89-4ffd-a412-5476150fb536

https://codepen.io/WebDevSimplified/pen/JjdveeV

mohasarc:
Thanks for the suggestion. Actually, I have considered this before; however, there are a couple of issues with it:

1. We don't know the size of the subtree under the dragged element, which could potentially cause performance issues during insertion.

2. Any event listeners attached to the dragged element or its children might interfere with the drag-and-drop process.

mohasarc:
We don't really need to use `onDropEnd`, please check the current implementation https://github.com/codelab-app/platform/pull/3134/files#diff-78828ab6f4ce29ffe17a9057349d8b5455eb7184a7cc467299dd5160369da07cR1-R52. 

By events I mean any event listener attached to the element by the user (not dnd events). On the other hand, if we remove the children, we wouldn't get an accurate representation because most probably the dragging element's size is determined by it's children (if it doesn't have a set size).

Instead we could make the current drop indicator of the same size as the dragging element.

webberwang:
> We don't really need to use `onDropEnd`, please check the current implementation https://github.com/codelab-app/platform/pull/3134/files#diff-78828ab6f4ce29ffe17a9057349d8b5455eb7184a7cc467299dd5160369da07cR1-R52.
> 
> By events I mean any event listener attached to the element by the user (not dnd events). On the other hand, if we remove the children, we wouldn't get an accurate representation because most probably the dragging element's size is determined by it's children (if it doesn't have a set size).
> 
> Instead we could make the current drop indicator of the same size as the dragging element.

Having same size box is out of scope, we don’t really need it for now. Just keep the code simple

melhamin:
@mohasarc looks quite good 💯.

libs/frontend/application/renderer/src/renderer.application.service.ts
```diff
@@ -53,6 +53,10 @@ export class RendererApplicationService
     return renderer
   }
 
+  get activeElementTree() {
+    return this.activeRenderer?.current.containerNode.current
+  }
+
```
mohasarc, line 59: @webberwang I've added this method to get teh active element tree and used it everywhere instead of the one in `BuilderService` because:

1. we don't have more than one active renderer at any time (we used to do a long time ago)
2. the method in `BuilderService` depend on selected nodes; however, we don't always have a selected node.

I kept the implementation inside the `BuilderService` for now because it is internally used; however, I feel like the builder doesn't need to know about the active element tree or selected nodes as they aren't domain data.

nettohadi:
Hi @mohasarc. Regarding "Margin Around Droppable Components", do we still show the margins in production?

mohasarc:
Hi @nettohadi, the margins are used within the detection algorith only; no margins are added to HTML Elements. in addition, the dnd is not enabled in production.

nettohadi:
> Hi @nettohadi, the margins are used within the collision detection algorith only; no margins are added to HTML Elements. in addition, the dnd is not enabled in production.

Ah I see. Thanks for explaining.

PR #3174, opened 2023-12-20
fix: app does not open when element of type component is used

### LINKED ISSUE #3173: App fails to open if element render type is a component.
## Description of Bug

creating an element with a render type of a component makes the app not open anymore.

https://github.com/codelab-app/platform/assets/51242349/8692f844-9f83-4d2a-a2de-0820feac7108

mohasarc:
## Related Issue(s)

Fixes #3173


vladyslav-polishchuk:
@mohasarc , could you double check, please, why integration tests job started to fail on this PR, maybe there is some problem with the changes

mohasarc:
It seems like the test is flaky, I reran the workflow and it passed this time.

PR #3185, opened 2024-01-08
fix: for making e-commerce app

### LINKED ISSUE #3157: New element is not rendered
## Description of Bug
New elements are not rendered until the page is refreshed.

## How to Reproduce

1.
2.

https://github.com/codelab-app/platform/assets/48331678/97b277d5-60e6-4bd0-9fb7-398479590660

### LINKED ISSUE #3184: Try to make e-commerce app
## Description of Bug
We were already able to create an e-commerce app before the huge refactor. Sample demos can be seen in [here](https://github.com/codelab-app/platform/pull/2886).

Since the huge refactor, there might be some breaking changes or bugs that make it unable to create the e-commerce app again.

This issue is just a placeholder for collecting the fixes that will be needed to make the [mock e-commerce app](https://demo.vercel.store/search) again.

jethrocabaluna:
## Description

This PR will be a collection of fixes for trying to copy the [mock e-commerce app](https://demo.vercel.store/search) again since the huge refactoring we had.

Fixes:
- added needed antd components
  - AntDesignLayoutHeader
  - AntDesignLayoutContent
  - AntDesignLayoutFooter
  - AntDesignInputSearch
- fix error on non-nullable `Field.api` error when there is an element with a field that is an interface type
- fix missing type ref error on element dependent types
- fix not able to create an element when not as a first child
- fix element not showing right away on the builder upon creation and needed to reload the page so that it shows
  - it was because the `element.current.children` is not in synced with the `sortedRuntimeChildren`, the latter is not being updated when there is a new child
  - there could be a better way to fix this
- fix empty elements options when selecting a page content container
- fix actions not showing up on the pre/post render actions dropdown
- fix pre/post render hook actions not working
- fix config data for api actions are not being saved properly, causing nested `config.data.data.data.data...data` when saving
- fix using success and error actions on API actions with code actions
- fix using state props on code actions
- fix access of state from provider store on regular pages
- fix pre/post render hooks triggering when resizing that can cause infinite renders if a code action save a data on state
- fix child mapper props not being used on mapped components
- fix actions not running when used on props like `onSearch`, `onClick`

## Video or Image

**WIP**

https://github.com/codelab-app/platform/assets/27695022/3ad0321f-ed3d-42d1-81c7-052d88ac92df

## Related Issue(s)

Fixes #3157 
Fixes #3184


libs/frontend/application/renderer/src/runtime-element.model.ts
```diff
@@ -250,6 +251,13 @@ export class RuntimeElement
    */
   @computed
   get renderChildren(): ArrayOrSingle<ReactNode> {
+    if (
```
vladyslav-polishchuk, line 254: Could you check if child-mapper feature would not cause infinite re-render loop, please?
I am afraid that if there is child mapper defined - this condition will always be true

jethrocabaluna: Ah I see. Yeah it does seem like it would cause an infinite rendering, although I did not encounter that in my app that has a child mapper feature. I think its because I filtered the `sortedRuntimeChildren` with `isRuntimeElement` and the child that is added by the child mapper pipe is a [component](https://github.com/codelab-app/platform/blob/b6a9c8b0f5cba5cd537299674bf3f2cc3a9eb4bd/libs/frontend/application/renderer/src/renderPipes/child-mapper-render-pipe.ts/#L38) which would be of type container node based on this [logic](https://github.com/codelab-app/platform/blob/82532606f3152d05490d634d2edaf8739669fb6b/libs/frontend/application/renderer/src/runtime-element.model.ts#L239).

Although, I think there might be a better solution than the fix I did here currently. I'll update this fix if Im able to find a better way.

vladyslav-polishchuk: Oh, thanks for the explanation.
And the fix does look OK to me, if it does not break anything else. Since I also faced this problem and was not able to find any better solution

vladyslav-polishchuk: I am working on getting e2e be passing on PRs, and this fix seems to be quite important one, a lot of tests fail because of the issue. I applied your fix and it seems to fix many of the cases, but there was a new problem introduced with it as well. There is infine rendering in some cases, and app hangs. One of such cases is simply adding two elements on a page (see video below). I created slightly updated version of your fix, that eliminates this problem (I do not fully re-render the current element) - https://github.com/codelab-app/platform/pull/3200. I made this as separate PR to get it merged sooner and unblock my further work.

https://github.com/codelab-app/platform/assets/74900868/29c4018c-dfc4-43aa-bf24-462218432958

Issue #3171, opened 2023-12-13
Fix CircleCI builds (e2e job)


vladyslav-polishchuk:
## What Feature Can Be Enhanced?
 
Currently the e2e job is the last part of the build that is constantly failing.
In order to get "green" builds, e2e tests as well as issues in platform should be fixed.
This is prerequisite for updating CircleCI to trigger production deployments on successful builds.

PR #3196, opened 2024-01-18
fix: impossible to delete apps

vladyslav-polishchuk:
## Video or Image

Before:

https://github.com/codelab-app/platform/assets/74900868/470eb86c-9c81-4150-8e1e-2c30758bba0b

After:

https://github.com/codelab-app/platform/assets/74900868/ab4f27b9-8678-4276-b0fa-c8cff0f36868

## Related Issue(s)

Part of #3171

### DIFF
--- a/libs/frontend/application/app/src/services/app.service.ts
+++ b/libs/frontend/application/app/src/services/app.service.ts
@@ -87,33 +87,35 @@ export class AppService
   @modelFlow
   @transaction
   delete = _async(function* (this: AppService, apps: Array<IAppModel>) {
-    const deleteApp = _async(function* (this: AppService, app: IAppModel) {
+    const deleteApp = async (app: IAppModel) => {
       /**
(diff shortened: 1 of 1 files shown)


libs/frontend/application/app/src/services/app.service.ts
```diff
@@ -87,10 +87,14 @@ export class AppService
   @modelFlow
   @transaction
   delete = _async(function* (this: AppService, apps: Array<IAppModel>) {
-    const deleteApp = _async(function* (this: AppService, app: IAppModel) {
+    for (const app of apps) {
```
webberwang, line 90: Mouse_Highlight_Overlay

I think keeping the `deleteApp`  function makes it more separate, I think the issue was some odd interaction with the `_async`. Can you try using a normal async function like the screenshot?

webberwang (changes requested):

PR #3200, opened 2024-01-21
fix: newly added children are not rendered until refresh

vladyslav-polishchuk:
## Description

Re-render element if new children was added

## Video or Image

Before:

https://github.com/codelab-app/platform/assets/74900868/2d60aec3-1b20-4c2f-bc36-7aed3ebb87cf

After:

https://github.com/codelab-app/platform/assets/74900868/ce0232f5-80cd-4bf2-8d56-7470770aeec8

## Related Issue(s)

Part of #3171

### DIFF
--- a/libs/frontend/application/renderer/src/runtime-element.model.ts
+++ b/libs/frontend/application/renderer/src/runtime-element.model.ts
@@ -8,6 +8,7 @@ import type {
 } from '@codelab/frontend/abstract/application'
 import {
   getRendererService,
+  isRuntimeElement,
   isRuntimeElementRef,
   RendererType,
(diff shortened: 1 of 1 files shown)


libs/frontend/application/renderer/src/runtime-element.model.ts
```diff
@@ -250,6 +251,14 @@ export class RuntimeElement
    */
   @computed
   get renderChildren(): ArrayOrSingle<ReactNode> {
+    if (
```
webberwang, line 254: this should have a comment, something like the PR title

webberwang (changes requested):

PR #3202, opened 2024-01-24
fix: routing issue and multiple children in component issue

### LINKED ISSUE #3199: Cannot open any app if a component has more than one element
## How to Reproduce

1. Create a component
2. Add to elements

https://github.com/codelab-app/platform/assets/48331678/ff0a105f-48f3-4f25-856a-5749c0579223

vladyslav-polishchuk:
## Description

1. Fix issue when sidebar buttons were not disabled when necessary (when on `/app` page, buttons for `components`, `pageList` and `builder` should be disabled). This happened because `routerServer` assigned values like this: 
```
    this.setPath({
      appSlug: `${appSlug}`,
      componentSlug: `${componentSlug}`,
      pageSlug: `${pageSlug}`,
      userSlug: `${userSlug}`,
    })
```
and in case if slug was undefined - value "undefined" as string was saved.

Before             |  After
Screenshot 2024-01-24 at 09 28 47|Screenshot 2024-01-24 at 09 27 56

2. Fix the issue when apps fail to open when a component with 2 children exists. This happened because of this validation:
```
assertContainsAtMostOne([nextSibling, parentElement])
```
But this is not valid, since the first child will have both references to parentElement, and to the nextSibling.

## Related Issue(s)

Fixes #3199

### DIFF
--- a/libs/frontend/application/shared/store/src/router/router.service.ts
+++ b/libs/frontend/application/shared/store/src/router/router.service.ts
@@ -5,17 +5,7 @@ import type {
 } from '@codelab/frontend/abstract/application'
 import { throwIfUndefined } from '@codelab/shared/utils'
 import { computed } from 'mobx'
-import {
-  _async,
-  _await,
-  Model,
-  model,
-  modelAction,
-  modelFlow,
-  prop,
-  transaction,
-} from 'mobx-keystone'
-import type { NextRouter } from 'next/router'
+import { Model, model, modelAction, prop } from 'mobx-keystone'
 import { ParsedUrlQuery } from 'querystring'
 
 const init = (routerQuery: ParsedUrlQuery) => {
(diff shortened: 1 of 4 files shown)


libs/frontend/application/shared/store/src/router/router.service.ts
```diff
@@ -83,10 +73,10 @@ export class RouterService
       routerQuery
 
     this.setPath({
-      appSlug: `${appSlug}`,
-      componentSlug: `${componentSlug}`,
-      pageSlug: `${pageSlug}`,
-      userSlug: `${userSlug}`,
+      appSlug: `${appSlug ?? ''}`,
```
webberwang, line 76: we couldn't use `({ appSlug })` directly because of `Array<string>` casting.

Would make more sense to keep the value undefined

```
appSlug: appSlug ? `${appSlug}` : undefined
```

empty string doesn't really make sense here

webberwang (changes requested):

PR #3209, opened 2024-02-02
fix: exporting an app

### LINKED ISSUE #3208: Exporting/Importing an app is not working
## Description of Bug
We used to be able to export and import an app before but right now it is broken if you try to export or import an app via UI

## How to Reproduce

1. Go to the Apps page and click the context menu on one of the app and select export
2. You will be redirected to an export route but with a 404 response

jethrocabaluna:
## Description

Fixed exporting and importing an app.
- the exported data will include the used resources and components as well
- the import logic will also import the resources and components that are used by the app
- fixed typings used on export/import logic
- improve logic around exporting and importing nested components used in the app
- added logic for exporting and importing resources used in the app

### Exporting an app

https://github.com/codelab-app/platform/assets/27695022/0a15309d-192a-4b1a-a7d7-d0060f1a73a4

### Importing an app

https://github.com/codelab-app/platform/assets/27695022/92ef60d6-1afc-4198-a86b-364257ac1faa

## Related Issue(s)

Fixes #3208


webberwang:
@jethrocabaluna 🚀 any feedback on the import/export code overall?

webberwang:
> > > @jethrocabaluna 🚀 any feedback on the import/export code overall?
> > 
> > 
> > Yeah I like the approach. Took me some time to fully understand the implementation and idea that was already in there but just needed some fixing and additions.
> 
> I guess another thought is in the use of the `*Repository.find`s, maybe harder to make them more flexible when we start using them in some other cases because the same `selectionSet` is used. Maybe we can make the `find` accept `selectionSet` instead.

Oh good thought, it has a method override

Mouse_Highlight_Overlay

PR #3214, opened 2024-02-08
fix: build error and digital ocean deployment

vladyslav-polishchuk:
## Description

- fix typescript script error when looping through types and trying to read `fields` prop, but `fields` exist only on interfaces
- fix deployment issue in DO, when was unable to determine port and server could not start

## Related Issue(s)

Fixes #

### DIFF
--- a/libs/backend/infra/adapter/codelab/src/graphql/endpoint.config.ts
+++ b/libs/backend/infra/adapter/codelab/src/graphql/endpoint.config.ts
@@ -8,9 +8,16 @@ export const ENDPOINT_CONFIG_KEY = 'graphql'
  */
 export const endpointConfig = registerAs(ENDPOINT_CONFIG_KEY, () => {
   return {
-    graphqlApiPort: env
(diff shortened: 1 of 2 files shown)


libs/backend/infra/adapter/codelab/src/graphql/endpoint.config.ts
```diff
@@ -8,9 +8,16 @@ export const ENDPOINT_CONFIG_KEY = 'graphql'
  */
 export const endpointConfig = registerAs(ENDPOINT_CONFIG_KEY, () => {
   return {
-    graphqlApiPort: env
-      .get('NEXT_PUBLIC_PLATFORM_API_HOST')
-      .required()
-      .asUrlObject().port,
+    get graphqlApiPort() {
```
webberwang, line 11: any benefit to have 2 env var, so we won't have conditional logic?

host = postname + port

would it be better to do `HOSTNAME` and `PORT`?

webberwang (changes requested):

PR #3215, opened 2024-02-08
fix: components cannot be drag/dropped

### LINKED ISSUE #3175: cannot drag and drop components in builder
## Description of Bug

cannot select and drag/drop an element that's an instance of a component. When drag/dropped a new `HTMLDiv` is created and the component is removed from the builder.

## How to Reproduce

1. create an element as an instance of a component
2. select and drag/drop it into the builder

https://github.com/codelab-app/platform/assets/51242349/f6262320-97c7-45a6-8c1f-751f87ea5e3e

mohasarc:
## Description

This PR fixes the issue with components cannot be drag/dropped, and along the way does the following:

- [x] Refactor overlays
- [x] Show overlays for non-concrete elements (elements that don't have corresponding html like components and fragments)
- [ ] hover and select the component instead of its children.
- [ ] handle circular referance cases when drag/dropping components
- [ ] fix the issue with component children not having unique `data-element-id`s which cause overlays to overlap.

## Related Issue(s)

Fixes #3175


mohasarc:
closed in favor of #3400

PR #3216, opened 2024-02-09
fix: state expressions not evaluated

### LINKED ISSUE #3213: State expressions are not evaluated
## How to Reproduce

1.
2.

https://github.com/codelab-app/platform/assets/48331678/de532624-d5fc-4ecd-baf9-33a96d76b37a

melhamin:
## Description
There was a small truthy/falsy bug that caused variables with a value of 0 not to be included in the state.

## Related Issue(s)

Fixes #3213


webberwang:
@melhamin good catch thanks

PR #3229, opened 2024-02-15
wip: styles for different states

### LINKED ISSUE #3221: Styles for different states of an element
## Description
Allow editing styles of different states of an element. Users should be able to define CSS styles for `hover`, `focused`... states of the element.

## Proposed implementation
Categorize CSS styles of an element by states. So the user selects the state and then the corresponding styles can be edited.

Screenshot 2024-02-10 at 3 00 19 PM

melhamin:
## Description
Allows setting CSS styles for different states (hover, focus...) of elements.

- [x] Styles for different states of an element
- [x] fix Style specificity issues

## Video or Image

https://github.com/codelab-app/platform/assets/48331678/52fd2f94-2bbd-4f67-bdea-ec773a6e4d0b

## Related Issue(s)

Fixes #3221

### DIFF
--- a/libs/frontend/abstract/domain/src/element/element-style.model.interface.ts
+++ b/libs/frontend/abstract/domain/src/element/element-style.model.interface.ts
@@ -2,9 +2,15 @@ import type { Nullable } from '@codelab/shared/abstract/types'
 import type { BuilderWidthBreakPoint } from '../builder'
 import type { CssMap } from './element.model.interface'
 
+export enum ElementStyleSelector {
+  None = 'none',
+  Hover = 'hover',
+  Focus = 'focus',
+}
+
 export interface IBreakpointStyle {
   cssString?: string
-  guiString?: string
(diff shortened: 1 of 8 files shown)


libs/frontend/abstract/domain/src/element/element-style.model.interface.ts
```diff
@@ -2,9 +2,15 @@ import type { Nullable } from '@codelab/shared/abstract/types'
 import type { BuilderWidthBreakPoint } from '../builder'
 import type { CssMap } from './element.model.interface'
 
+export enum ElementStyleSelector {
```
webberwang, line 5: https://developer.mozilla.org/en-US/docs/Web/CSS/Pseudo-classes

Naming should relate this, something like `ElementStylePseudoClass`

webberwang (changes requested):

Issue #3063, opened 2023-10-09
Fix websites projects


assimhabeek:
## What Feature Can Be Enhanced?

Get websites project working following refactor done by #3023

webberwang:
@vladyslav-polishchuk Add a cypress spec for a simple production site app.

Just build 2 pages and visit each page and assert some text content on the page

vladyslav-polishchuk:
> @vladyslav-polishchuk Add a cypress spec for a simple production site app.
> 
> Just build 2 pages and visit each page and assert some text content on the page
> 
> This will be basically like `in-app-routing.cy.ts`

@webberwang, do you mean manually creating an application in production (admin.codelab.app), creating a couple of pages, assigning a domain and then in spec only try to access the resulting website (over custom user domain)?

vladyslav-polishchuk:
@webberwang, do you think it makes sense to assign some demo domains to every app that are created by users?
For example, whenever user creates an app - it gets `app-id.websites.codelab.app` domain, and the app becomes immediately available by that URL. And if user wants to add custom one - this could be a part of some paid plan.
Just an idea. Seems like could be easily implemented

webberwang:
> @webberwang, do you think it makes sense to assign some demo domains to every app that is created by users? For example, whenever user creates an app - it gets `app-id.websites.codelab.app` domain, and the app becomes immediately available by that URL. And if user wants to add custom one - this could be a part of some paid plan. Just an idea. Seems like could be easily implemented

Yeah definitely could you create a task for the backlog?

PR #3280, opened 2024-03-10
feat: production websites fixes and enhancements

### LINKED ISSUE #3287: Refactor build app
## Description

BuildAppModal_tsx_—_platform

Move shared logic under `appService.regeneratePages()` or something similar

vladyslav-polishchuk:
## Description

- Added Cypress test that checks if user app is available in production
- Fix script error in production websites when URLs were concatenated incorrectly
- Refactored app/pages build functionality to not duplicate code
- Fixed graphql query that used duplicated elements
- Improved error/success handling when app/pages is build for production

## Related Issue(s)

Fixes #3287 
Also part of #3063

### DIFF
--- a/libs/shared/config/src/env/services/endpoint.ts
+++ b/libs/shared/config/src/env/services/endpoint.ts
@@ -33,7 +33,7 @@ export class EndpointEnvVars implements IEndpointEnvVars {
    * URL is protocol + origin
    */
   get canActivateUrl() {
-    return `${this.platformApiHost}/api/can-activate`
+    return new URL('api/can-activate', this.platformApiHost).toString()
   }
 
   /**
@@ -42,7 +42,7 @@ export class EndpointEnvVars implements IEndpointEnvVars {
    * This uses the Next.js proxy middleware
    */
   get graphqlApiProxyUrl() {
-    return `${this.platformHost}/api/graphql`
+    return new URL('api/graphql', this.platformHost).toString()
   }
 
   get isLocal() {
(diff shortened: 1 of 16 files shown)


libs/shared/config/src/env/services/endpoint.ts
```diff
@@ -33,7 +33,7 @@ export class EndpointEnvVars implements IEndpointEnvVars {
    * URL is protocol + origin
    */
   get canActivateUrl() {
-    return `${this.platformApiHost}/api/can-activate`
+    return new URL('api/can-activate', this.platformApiHost).toString()
```
vladyslav-polishchuk, line 36: Use `URL` constructor to create url strings. This allows to avoid invalid concatenation and as a result - invalid urls like `admin.codelab.app//api/can-activate`

terraform/modules/digitalocean-codelab/platform-api.tf
```diff
@@ -9,6 +9,7 @@ resource "digitalocean_app" "platform-api" {
     }
 
     service {
+      instance_size_slug = "basic-xxs"
```
vladyslav-polishchuk, line 12: Adding this just to avoid diff during `terraform plan/terraform apply`. This is the smallest instance and this is the same value that is used by default

webberwang: I see just for testing in the meantime

libs/shared/config/src/env/services/endpoint.ts
webberwang: Sure seems better

PR #3283, opened 2024-03-11
chore: in-app routing specs

### LINKED ISSUE #3011: Simplify route change hooks in the preview mode
We have this logic in the [preview mode](https://github.com/codelab-app/platform/blob/36f2515683549d5b611106ec5701fb6c47c46f63/apps/platform/pages/apps/%5BuserName%5D/%5BappSlug%5D/pages/%5BpageSlug%5D/index.tsx) where we use three effect hooks. We could simplify this, and should add notes on how or why it is implemented this way.

Also make sure that there are no dupe listeners being added into the `routeChangeStart` event. You can check this by adding console log on the event handler, it should only trigger once per route change

We should also remove the routing hooks logic in the builder file, since the NextLink behaviour should be disabled in the builder, and only works on the preview mode, as mentioned in here - https://github.com/codelab-app/platform/pull/2988#issuecomment-1715393437

image

jethrocabaluna:
## Description

This is for improving the in-app routing specs
- when in preview mode, should be able to navigate to static and dynamic page
- when in builder mode, navigation is prevented in the app
- refactored spec to seed setup data via api

Also implemented an override props for props like `href` so it is changed to `#` when in builder mode so it doesn't navigate when clicked.

## Related Issue(s)

Fixes #3011

### DIFF
--- a/apps/platform-e2e/src/e2e/builder/in-app-routing.cy.ts
+++ b/apps/platform-e2e/src/e2e/builder/in-app-routing.cy.ts
@@ -1,292 +1,106 @@
-import { FIELD_TYPE } from '@codelab/frontend/test/cypress/antd'
-import type { IAppDto } from '@codelab/shared/abstract/core'
-import { IAtomType, IPageKindName } from '@codelab/shared/abstract/core'
-import { ROOT_ELEMENT_NAME } from '@codelab/shared/config'
-import { slugify } from '@codelab/shared/utils'
+import type { App } from '@codelab/shared/abstract/codegen'
+import type { IAppDto, IPage, IPageDto } from '@codelab/shared/abstract/core'
+import { IPageKind, IPageKindName } from '@codelab/shared/abstract/core'
+import { findOrFail, slugify } from '@codelab/shared/utils'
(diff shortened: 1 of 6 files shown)


apps/platform-e2e/src/e2e/builder/in-app-routing.cy.ts
```diff
@@ -1,292 +1,129 @@
...
+      )
+    })
+    it('should disable navigation', () => {
+      cy.get('#render-root')
+        .contains(providerPageLinkElementCreateData.propsData?.customText)
+        .click()
+
+      // No navigation should occur. Give it some time to navigate and check if the text is still there.
+      // eslint-disable-next-line cypress/no-unnecessary-waiting
+      cy.wait(2000)
```
webberwang, line 120: You likely saw this https://stackoverflow.com/questions/76253544/cypress-how-to-assert-that-no-navigation-happend

But would this approach work instead? https://stackoverflow.com/questions/73849805/cypress-test-nothing-happen

webberwang (changes requested):

Issue #3249, opened 2024-02-28
Fix e2e tests


jethrocabaluna:
## What Feature Can Be Enhanced?
This is for fixing the cypress tests that are still failing that may be due to the huge recent refactors, mostly in the builder logic.
- [x] child-mapper.cy.ts @jethrocabaluna 
- [x] pre-render-api-action-and-set-state.cy.ts @jethrocabaluna 
- [x] child-mapper-api-actions-with-props.cy.ts @jethrocabaluna 
- [x] actions-inside-code-actions.cy.ts  @jethrocabaluna 
- [x] state-sharing.cy.ts (@webberwang)
- [x] in-app-routing.cy.ts @jethrocabaluna 
- [x] nested-api-actions.cy.ts @jethrocabaluna 
- [x] form.cy.ts (@webberwang )
- [x] convert-element-to-component.cy.ts @jethrocabaluna 
- [x] component.cy.ts
  - **NOTE**: this is flaky
- [x] provider-page.cy.ts @jethrocabaluna

PR #3286, opened 2024-03-13
fix: child-mapper spec with refactors

jethrocabaluna:
## Description

This PR fixes these e2e specs related to child mapper logic and clean up some test data used for the specs
- child-mapper.cy.ts
- child-mapper-api-actions-with-props.cy.ts

## Related Issue(s)

Relates #3249


webberwang (changes requested):
Some changes

PR #3289, opened 2024-03-14
fix: convert element to component spec

jethrocabaluna:
## Description

This PR fixes the spec `convert-element-to-component.cy` and convert the seeding logic to use the APIs

## Related Issue(s)

Relates #3249


libs/frontend/application/renderer/src/runtime-element.model.ts
```diff
@@ -292,9 +292,9 @@ export class RuntimeElementModel
           ? {
               ...child.runtimeRootElement.treeViewNode,
               ...(!isNil(child.childMapperIndex) ? { children: [] } : {}),
-              isChildMapperComponentInstance: isComponent(
```
webberwang, line 295: Glad the spec caught some errors

PR #3290, opened 2024-03-14
chore: fix pre/post render actions with set state e2e

jethrocabaluna:
## Description

Fixes the `pre-render-api-action-and-set-state.cy.ts` spec
- also adds test for the post render action
- refactor seeding to use APIs

## Video or Image

Screenshot 2024-03-15 at 2 36 10 PM

## Related Issue(s)

Relates #3249


apps/platform-e2e/src/e2e/store/pre-post-render-api-action-and-set-state.cy.ts
```diff
@@ -0,0 +1,168 @@
...
+  })
+
+  it('should add the state variables where the api responses will be stored to', () => {
+    cy.getCuiSidebarViewHeader('State').should('exist').click()
+    cy.waitForSpinners()
+
+    for (const stateKey of [stateKeyPreRender, stateKeyPostRender]) {
+      cy.getCuiSidebarViewHeader('State').getCuiToolbarItem('Add Field').click()
+
+      cy.setFormFieldValue({
```
webberwang, line 107: This part should also be seeded via api, since we specifically test the UI for creating state inside store.cy

apps/platform-e2e/src/e2e/store/pre-post-render-api-action-and-set-state.cy.ts
```diff
@@ -0,0 +1,168 @@
...
+  preRenderCodeActionCreateData,
+  providerPageElement,
+  stateKeyPostRender,
+  stateKeyPreRender,
+} from './pre-post-render-api-action-and-set-state.data'
+
+let app: IAppDto
+let page: IPageDto
+
+const setupTest = () => {
```
webberwang, line 29: I think spec filename could be `pre-post-api-action`

can also move setup to `pre-post-api-action.setup.ts`

the render & set state is more about the method in which we we test the main subject

webberwang (changes requested):

PR #3291, opened 2024-03-14
fix: interfaceForm unit test error

### LINKED ISSUE #3285: ErrorField causes form to not render in specs
## Description

Form_tsx_—_platform

`InterfaceForm.test.tsx` not working if we add the error fields

melhamin:
## Related Issue(s)

Fixes #3285


webberwang:
Oh nice didn't know could work like this

PR #3295, opened 2024-03-18
chore: refactor nested-api-actions spec

jethrocabaluna:
## Description

- refactors the nested-api-actions spec to seed via API
  - creation of page state is still done via UI
  - added test case for error action
- separate setupTest functions on the other spec files

## Related Issue(s)

Relates #3249


webberwang:
great work!

PR #3309, opened 2024-03-27
chore: reusable layout component

### LINKED ISSUE #3061: Create re-useable layout components
## Description

Layout_-_Ant_Design
Layout_-_Ant_Design

Create a single `Layout Component`, that takes in a prop of

```
{
  "style": "A" | "B" | "C" | "D"
}
```

jethrocabaluna:
## Description

Still need to figure out / fix
~- [ ] rendering the children into the nested component's children container (layout content) e.g.~
```
<LayoutComponent>
  <LayoutA>
    {children}
  </LayoutA>
<LayoutComponent>
```
  - changed approach to make content a prop as well, as there is still some issues when trying to use children
~- [ ] conditionally rendering children in one of the nested components based on selected layout style~
  - changed approach to make content a prop as well, as there is still some issues when trying to use children
- [ ] sometimes encountering infinite renders and hangs the browser tab

## Video or Image

WIP: layout custom component, still encounter infinite renders when changing layout props e.g. `layoutStyle`, but the layout works after reloading

https://github.com/codelab-app/platform/assets/27695022/c9bb8a91-9bfc-436b-8433-b1344c9f31cb

## Related Issue(s)

Fixes #3061

### DIFF
--- a/examples/app-with-layout/export.json
+++ b/examples/app-with-layout/export.json
@@ -1,0 +1,1914 @@
+{
+  "app": {
+    "__typename": "App",
+    "id": "2ace741c-716c-45be-b1a0-b3471f7d109f",
+    "name": "My App",
+    "slug": "my-app",
+    "domains": [],
+    "pages": [
+      {
+        "id": "d63aec78-fc2c-42ec-8458-c877d190da3b"
+      },
+      {
+        "id": "e23393d7-28cb-4064-8cf6-fe799dcacc01"
+      },
+      {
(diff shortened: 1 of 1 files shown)


webberwang:
@jethrocabaluna I would build `LayoutA` first with children prop binding, then `LayoutComponent` could pass via with `layoutAChildren` prop, which is bound to children on `LayoutA`.

Which means we would need prop key mapping feature>

---

Do you mean say if we have 3 children, how do we conditionally render them?

PR #3322, opened 2024-04-10
test: writing specs for element tree view

### LINKED ISSUE #3321: Write unit specs for antdTree
Element tree view has many cases that needs to be tested:
- Show element which belongs to the main tree.
- When element is a component instance don't show component as a child. (This goes against the real shape of tree)
- Show instance element children as direct children (even thought they will be wrapped by component on render. or not rendered at all)
- Show child mapper children     
- Set correct rootKey

Many issues are related to this is one :

#3319
#3308
#3272

assimhabeek:
## Related Issue(s)

Fixes #3321


webberwang:
specs look great

assimhabeek:
Thanks

PR #3363, opened 2024-05-19
feat: combine 2 renderType fields into one

### LINKED ISSUE #3206: Micro UX improvement: unify render type lists
## Description

Currently user has to use the render type kind first then pick the render type itseld; however, the concept of render type kind seems to be irrelevent for the user. hence we can have a single list for all render types with a toggle to show/hide based on render type kind (for advanced users).

image

vladyslav-polishchuk:
## Related Issue(s)

Fixes #3206


vladyslav-polishchuk:
Marked as draft since requires quick re-testing once migration to new app router is completed.

PR #3364, opened 2024-05-19
fix: state-sharing.cy e2e spec

vladyslav-polishchuk:
## Related Issue(s)

Fixes #


webberwang:
Oh great, this was causing many issues

vladyslav-polishchuk:
> Oh great, this was causing many issues

Does not work yet, though. Just had some thoughts and was experimenting with fixes

webberwang:
Oh I will revert manually

PR #3390, opened 2024-06-01
Render if fix

### LINKED ISSUE #3389: "Render if" field looses focus after each key press
## How to Reproduce

1. Select any element in builder
2. Navigate to "Render if" field and try to type some expression there
3. The focus is lost on each key press. Probably because field gets re-rendered each time new character is added

https://github.com/codelab-app/platform/assets/74900868/9f2f7994-9213-4e14-93e1-db8247f5b92b

vladyslav-polishchuk:
## Video or Image

https://github.com/codelab-app/platform/assets/74900868/507cf543-a89b-4416-b012-e51a80257c96

## Related Issue(s)

Fixes #3389

### DIFF
--- a/libs/frontend/application/element/src/use-cases/element/update-element/UpdateElementForm.tsx
+++ b/libs/frontend/application/element/src/use-cases/element/update-element/UpdateElementForm.tsx
@@ -9,14 +9,14 @@ import { useStore } from '@codelab/frontend/application/shared/store'
 import { SelectActionField } from '@codelab/frontend/application/type'
 import {
   CodeMirrorField,
-  createAutoCompleteOptions,
   Form,
+  useAutocompleteOptions,
 } from '@codelab/frontend/presentation/view'
 import { createFormErrorNotificationHandler } from '@codelab/frontend/shared/utils'
 import { CodeMirrorLanguage } from '@codelab/shared/abstract/codegen'
 import { Collapse } from 'antd'
 import { observer } from 'mobx-react-lite'
-import React from 'react'
+import React, { useMemo } from 'react'
 import { AutoField, AutoFields } from 'uniforms-antd'
 import { AutoComputedElementNameField } from '../../../components/auto-computed-element-name'
 import ChildMapperCompositeField from '../../../components/ChildMapperCompositeField'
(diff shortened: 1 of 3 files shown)


libs/frontend/application/element/src/use-cases/element/update-element/UpdateElementForm.tsx
```diff
@@ -71,12 +73,9 @@ export const UpdateElementForm = observer<UpdateElementFormProps>(
       {
         children: (
           <AutoField
-            component={CodeMirrorField({
-              customOptions: createAutoCompleteOptions(
-                runtimeProps.runtimeContext,
-              ),
-              language: CodeMirrorLanguage.Javascript,
-            })}
+            component={codeMirrorField}
```
vladyslav-polishchuk, line 76: Fix the re-rendering issue. The issue was because the Form has autosave attribute and each new character caused state update, as a result autocompletion props got recalculated and cause the field to re-render. For now the CodeMirrorComponent is not re-created on each render, as well as autocompletion array persists the same reference, whenever possible.

libs/frontend/presentation/view/src/components/codeMirror/completion-options.ts
```diff
@@ -7,15 +7,19 @@ import type { Completion } from '@codemirror/autocomplete'
...
+import { useEffect, useState } from 'react'
 
 const getOptions = (
   ctx: IPropData = {},
   parentKey: string,
   sectionName: string,
 ): Array<Completion> =>
   Object.entries(ctx).flatMap(([key, value]) => {
-    const fullKey = `${parentKey}.${key}`
+    const fullKey = key.includes('-')
```
vladyslav-polishchuk, line 20: Fix autocompletion options that have '-' in the name. Instead of invalid `props.data-element-id` now is used `props['data-element-id']`.
Screenshot 2024-06-01 at 16 26 48

libs/frontend/presentation/view/src/components/codeMirror/styles.ts
```diff
@@ -9,10 +9,10 @@ export const editorStyles = css`
...
 export const ExpandButton = styled(Button)`
   position: absolute !important;
-  bottom: 2px;
-  right: 2px;
-  width: 20px;
-  height: 20px;
+  bottom: 3px;
+  right: 3px;
+  width: 25px !important;
+  height: calc(100% - 6px);
```
vladyslav-polishchuk, line 15: Fix styles of expand button on CodeMirrorFields.

Before             |  After
Screenshot 2024-06-01 at 16 30 03|Screenshot 2024-06-01 at 16 29 26

Issue #3238, opened 2024-02-21
GraphQL endpoints are not guarded by auth0


vladyslav-polishchuk:
## How to Reproduce

Currently it is possible to access graphql endpoint in incognito mode in browser (to make sure you are not authenticated) and perform mutations on the database.

1. Go to https://admin.codelab.app/api/graphql
2. Perform mutation, for example change existing app name:
<img width="1728" alt="Screenshot 2024-02-21 at 09 17 10" src="https://github.com/codelab-app/platform/assets/74900868/e10ac07f-330f-464c-8478-ccde265d8002">

Here are some thoughts about the current implementation and possible ways to solve the issue:
There’s 2 layers here, one at nestjs, another at the neo4j schema level

<img width="959" alt="Screenshot 2024-02-21 at 09 42 40" src="https://github.com/codelab-app/platform/assets/74900868/5270933a-8824-4585-8a3e-516c96dedf7f">

We can use a nestjs level guard for the entire app, as opposed to neo4j level
It would be good to have both, but nestjs level should be less work to get the same effect

PR #3398, opened 2024-06-22
fix: protect neo4j database with authentication

vladyslav-polishchuk:
## Description

Add authentication guard on neo4j level for all mutation queries.
GET access is still granted for everyone.

## Video or Image

https://github.com/codelab-app/platform/assets/74900868/d0755288-75b9-4d0a-860f-357426fcac64

## Related Issue(s)

Part of #3238


libs/backend/infra/adapter/neo4j/src/schema/model/action.schema.ts
```diff
@@ -17,16 +17,16 @@ export const actionSchema = gql`
     id: ID!
     name: String!
     type: ActionKind! @settable(onUpdate: false)
-    store: Store! @relationship(type: "STORE_ACTION", direction: IN)
-    element: Element @relationship(type: "ELEMENT_ACTION", direction: OUT)
+    store: Store! @declareRelationship
```
vladyslav-polishchuk, line 20: This changes required because of migration @neo4j packages to next major version (v4 -> v5).
Interfaces no longer allowed to have implementation

PR #3407, opened 2024-06-27
refactor: create app use case

assimhabeek:
## Related Issue(s)

Fixes #


apps/web/components/Providers.tsx
```diff
@@ -0,0 +1,25 @@
+'use client'
```
vladyslav-polishchuk, line 1: I suggest to move this to some other folder, since `components` is going to be a route folder with `layout.ts` and `page.ts` files

vladyslav-polishchuk: Oh, you are right, my bad

assimhabeek: no, worries

PR #3417, opened 2024-07-14
feat: headless ui

mohasarc:
## Related Issue(s)

Fixes #


apps/web/styles/global.css
```diff
@@ -9,6 +9,10 @@
 /* Used by https://github.com/react-grid-layout/react-grid-layout */
 @import 'react-resizable/css/styles.css';
 
+@tailwind base;
```
webberwang, line 12: don't we already import tailwind.css above?

apps/web/styles/global.css
```diff
@@ -32,3 +36,81 @@
   font-weight: unset;
   font-size: unset;
 }
+
+@layer base {
```
webberwang, line 40: best to put this in a css file like tailwind.css above, but what are these for? 

is it required setup, or our config override

webberwang (changes requested):

PR #3434, opened 2024-08-04
fix: component and admin pages issues

vladyslav-polishchuk:


libs/frontend/application/admin/src/use-cases/import-data/import-admin-data.use-case.ts
```diff
@@ -2,9 +2,12 @@
 
 import { fetchWithAuth } from '@codelab/frontend-infra-fetch'
 
-export const importAdminDataUseCase = () =>
-  fetchWithAuth('admin/import', {
+export const importAdminDataUseCase = async () => {
+  const result = await fetchWithAuth('admin/import', {
```
vladyslav-polishchuk, line 6: Need to be careful, we can't return from server actions anything except primitives or plain objects. Othervise next.js throws `Error: Only plain objects, and a few built-ins, can be passed to Client Components from Server Components. Classes or null prototypes are not supported.`

libs/frontend/application/component/src/use-cases/create-component/create-component.use-case.ts
```diff
@@ -12,6 +12,6 @@ export const createComponentUseCase = async (
     name: componentData.name,
   })
 
-  await createComponentRepository(componentModel)
+  await createComponentRepository(componentModel.toCreateInput())
```
vladyslav-polishchuk, line 15: We can't pass mobx models from client to server actions, only plain objects

PR #3699, opened 2025-05-07
fix: script error on attempt to change renderType to component

### LINKED ISSUE #3696: Changing renderType to component causing builder failure
## How to Reproduce

1. Add element of type ReactFragment to page
2. Change renderType to component (GoogleFonts for example)

https://github.com/user-attachments/assets/bb07fc8b-1557-4052-b35d-81a64aa2ec4f

vladyslav-polishchuk:
## Related Issue(s)

Fixes #3696

### DIFF
--- a/libs/frontend/application/element/src/services/element.service.ts
+++ b/libs/frontend/application/element/src/services/element.service.ts
@@ -76,17 +76,21 @@ export const useElementService = (): IElementService => {
   const { componentDomainService, elementDomainService } = useDomainStore()
   const selectedNode = builderService.selectedNode
 
+  const loadElementRenderTypeApi = async (element: IUpdateElementData) => {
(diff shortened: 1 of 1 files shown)


libs/frontend/application/element/src/services/element.service.ts
```diff
@@ -156,7 +154,15 @@ export const useElementService = (): IElementService => {
     if (newRenderTypeId !== oldRenderTypeId) {
       await propService.reset(currentElement.props.toJson)
 
-      await atomService.loadApi(newRenderTypeId)
+      if (newElement.renderType.__typename === 'Atom') {
```
webberwang, line 157: this same logic is used in create as well, can create a private method on element to handle both

webberwang (changes requested):

PR #3701, opened 2025-05-13
wip: do not merge

vladyslav-polishchuk:
## Related Issue(s)

Fixes #3700


libs/frontend/shared/utils/src/uniformUtils.hook.ts
```diff
@@ -98,7 +99,19 @@ export const createValidator = (schema: Schema) => {
...
+      componentProps: {},
+      props: {},
+      refs: {},
+      rootActions: {},
+      rootRefs: {},
+      rootState: {},
+      state: {},
+      url: {},
+    })
+
```
assimhabeek, line 113: I believe we should evaluate at `Form` level where we have a real context, and keep this as abstract as possible,
https://github.com/codelab-app/platform/blob/e78fb496cfe32c95f828f1f8d544dc94ef8b1ac0/libs/frontend/application/element/src/use-cases/update-element-props/UpdateElementPropsForm.tsx#L109

PR #3824, opened 2025-10-03
2Upgrade uniforms s

vladyslav-polishchuk:
## Related Issue(s)

Fixes #


apps/web-e2e/src/modules/builder/builder.fixture.ts
```diff
@@ -150,7 +150,7 @@ export class BuilderPage extends BasePage {
       await this.page.getByText(`Delete \`${element.name}\``).click()
       await this.clickModalConfirmButton()
 
-      await expect(this.getGlobalProgressBar()).toBeHidden()
+      await this.waitForProgressBar()
```
vladyslav-polishchuk: This first waits for progressbar to appear and then dissappear. This is more reliable then just checking it does not exist. Since this check may pass even before the form autosubmits

